PR #46, opened 2022-11-01
Adds a new setting for saving document delay

hbcarlos:
I need to make the delay configurable because of JupyterLab UI Tests. When making a snapshot after opening/editing a notebook the icon for the dirty flag may or may not be there. We make sure the dirty icon is cleaned by setting the delay to 0 in the test.

### DIFF
--- a/jupyter_server_ydoc/ydoc.py
+++ b/jupyter_server_ydoc/ydoc.py
@@ -260,7 +260,7 @@ class YDocWebSocketHandler(WebSocketHandler, JupyterHandler):
         seconds = self.settings["collaborative_document_save_delay"]
         if seconds is None:
             return
-        # save after X second of inactivity to prevent too frequent saving
+        # save after X seconds of inactivity
         await asyncio.sleep(seconds)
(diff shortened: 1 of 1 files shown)


jupyter_server_ydoc/ydoc.py
```diff
@@ -257,8 +257,11 @@ def on_document_change(self, event):
         self.saving_document = asyncio.create_task(self.maybe_save_document())
 
     async def maybe_save_document(self):
-        # save after 1 second of inactivity to prevent too frequent saving
-        await asyncio.sleep(1)
+        seconds = self.settings["collaborative_document_save_delay"]
+        if seconds is None:
+            return
+        # save after X second of inactivity to prevent too frequent saving
```
davidbrochart, line 263:
```suggestion
        # save after X seconds of inactivity
```

davidbrochart (approved):
Thanks Carlos, just a minor change.

davidbrochart:
I'm going to release `0.2.0` so that it doesn't break JupyterLab.

PR #58, opened 2022-11-16
Pass logger to YRoom, require ypy-websocket >=0.5.0

davidbrochart:


jupyter_server_ydoc/handlers.py
```diff
@@ -37,8 +38,8 @@ class DocumentRoom(YRoom):
 
     is_transient = False
 
-    def __init__(self, type: str, ystore: BaseYStore):
-        super().__init__(ready=False, ystore=ystore)
+    def __init__(self, type: str, ystore: BaseYStore, log: Logger):
```
fcollonval, line 41:
```suggestion
    def __init__(self, type: str, ystore: BaseYStore, log: Optional[Logger]=None):
```

Should this be optional too as it is in the base class?

jupyter_server_ydoc/handlers.py
```diff
@@ -50,6 +51,9 @@ class TransientRoom(YRoom):
 
     is_transient = True
 
+    def __init__(self, log):
```
fcollonval, line 54: Same question

```suggestion
    def __init__(self, log: Optional[Logger]=None):
```

jupyter_server_ydoc/handlers.py
```diff
@@ -233,7 +244,10 @@ async def maybe_load_document(self):
 
     async def send(self, message):
         # needed to be compatible with WebsocketServer (websocket.send)
-        self.write_message(message, binary=True)
+        try:
+            self.write_message(message, binary=True)
+        except BaseException:
+            pass
```
fcollonval, line 250: It would be interesting to log for debugging, isn't it?

```suggestion
        except Exception as e:
            self.log.debug("Failed to write message", exc_info=e)
```

jupyter_server_ydoc/handlers.py
```diff
@@ -329,8 +346,10 @@ async def maybe_save_document(self):
             model["format"] = file_format
             model["content"] = self.room.document.source
             self.log.debug("Saving Y document to disk: %s", file_path)
-            model = await ensure_async(self.contents_manager.save(model, file_path))
-            self.last_modified = model["last_modified"]
+            async with self.lock:
+                model = await ensure_async(self.contents_manager.save(model, file_path))
+                self.last_modified = model["last_modified"]
+            self.log.debug(f"{self.last_modified=}")
```
fcollonval, line 352: The `=` syntax was introduced in Python 3.8, so you must not use it.
https://docs.python.org/3/reference/lexical_analysis.html#f-strings

```suggestion
            self.log.debug(f"last_modified={self.last_modified}")
```

fcollonval (changes requested):

fcollonval:
Thanks David

jupyter_server_ydoc/handlers.py
davidbrochart: Left-over, I forced-pushed before your comment :+1:

davidbrochart:
Thanks for the review @fcollonval.
Sorry to not commit your changes directly, as I forced-pushed before seeing your review, and I don't know if it would have been OK.

fcollonval:
> Sorry to not commit your changes directly, as I forced-pushed before seeing your review, and I don't know if it would have been OK.

No problem

PR #61, opened 2022-11-22
Add base_url to routes

davidbrochart:
Fixes https://github.com/jupyterlab/jupyterlab/issues/13463.

### DIFF
--- a/jupyter_server_ydoc/app.py
+++ b/jupyter_server_ydoc/app.py
@@ -6,6 +6,7 @@ try:
 except ModuleNotFoundError:
     raise ModuleNotFoundError("Jupyter Server must be installed to use this extension.")
 
+from jupyter_server.utils import url_path_join
 from traitlets import Float, Int, Type
 from ypy_websocket.ystore import BaseYStore  # type: ignore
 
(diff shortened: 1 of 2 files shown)


jtpio (approved):
Thanks!

hbcarlos (approved):
Thanks, @davidbrochart!

jupyter_server_ydoc/app.py
```diff
@@ -62,7 +63,9 @@ def initialize_settings(self):
     def initialize_handlers(self):
         self.handlers.extend(
             [
-                (r"/api/yjs/roomid/(.*)", YDocRoomIdHandler),
-                (r"/api/yjs/(.*)", YDocWebSocketHandler),
+                url_path_join(
+                    self.settings["base_url"], (r"/api/yjs/roomid/(.*)", YDocRoomIdHandler)
```
fcollonval, line 67: I don't think this is correct as for example this is not what is done by Jupyter Server Terminals:

https://github.com/jupyter-server/jupyter_server_terminals/blob/574e1018b7d924f09c1ca45389182f1fd314caee/jupyter_server_terminals/app.py#L72

davidbrochart: Yes, and it seems to be confirmed by the Binder.

davidbrochart: I just yanked jupyter-server-ydoc v0.5.0 on PyPI.

PR #64, opened 2022-11-25
Remove metadata

davidbrochart:
The `SQLiteYStore` now has a built-in [timestamp](https://github.com/y-crdt/ypy-websocket/blob/3d55469a27f47168713c55f93e23bd0164a03bfa/ypy_websocket/ystore.py#L202), and I think we should have it built-in for `FileYstore` too.
The current metadata was also storing a timestamp, there is no need to store it twice.


davidbrochart:
In https://github.com/y-crdt/ypy-websocket/pull/57, clearing the document history has been disabled by default. We are setting it back here.

PR #66, opened 2022-12-01
Allow to set SQLiteYStore's database path and document time-to-live

davidbrochart:
This PR allows to set the database path and the document time-to-live parameters of an `SQLiteYStore`. It doesn't apply to other types of `YStore`. The previous way of doing it was to pass a `YStore` class, so this PR is just a way to make it easier to parameterize an `SQLiteYstore`.

But more importantly, the default document time-to-live of an `SQLiteYStore` is now `None` (instead of 24 hours), which means that the documents' history will never be cleared. I think it is safer to do so, considering [the issues](https://github.com/y-crdt/ypy-websocket/pull/58#issuecomment-1332527064) that we are still trying to figure out.

cc @ellisonbg @dlqqq


dlqqq:
Small nit, is it possible to hide the CodeCov warnings in the review UI? It's really noisy and makes review more difficult.

jupyter_server_ydoc/app.py
```diff
@@ -41,14 +41,47 @@ class YDocExtension(ExtensionApp):
...
+
+    sqlite_ystore_document_ttl = Int(
+        None,
+        allow_none=True,
+        config=True,
+        help="""The document time-to-live in seconds. Defaults to None (document history is never
+        cleared). Only applicable if the YStore is an SQLiteYStore.""",
+    )
+
+    @observe("sqlite_ystore_document_ttl")
```
dlqqq, line 76: Could you elaborate on the difference between using the `validate` and `observe` traitlets decorators?

jupyter_server_ydoc/app.py
```diff
@@ -41,14 +41,47 @@ class YDocExtension(ExtensionApp):
...
+    @observe("sqlite_ystore_db_path")
+    def _observe_sqlite_ystore_db_path(self, change):
+        if issubclass(self.ystore_class, SQLiteYStore):
+            self.ystore_class.db_path = change["new"]
+        else:
+            raise RuntimeError(
+                f"ystore_class must be an SQLiteYStore to be able to set sqlite_ystore_db_path, not {self.ystore_class}"
+            )
+
+    sqlite_ystore_document_ttl = Int(
```
dlqqq, line 68: Wait, so why not just:

1. Have `JupyterSQLiteYStore` be a subclass of both `Configurable` and `SQLiteYStore`, defined at file scope rather than in a factory
2. Define these traits on that class itself
3. Revert back to `default_value=JupyterSQLiteYStore` on the `ystore_class` trait

I don't think adding configurable traits on the extension app class is the right approach here, given that they're specific to whether the user is using a subclass of `SQLiteYStore`.

jupyter_server_ydoc/handlers.py
```diff
@@ -28,16 +29,22 @@ class JupyterTempFileYStore(TempFileYStore):
     prefix_dir = "jupyter_ystore_"
 
 
-class JupyterSQLiteYStore(SQLiteYStore):
-    db_path = ".jupyter_ystore.db"
-    document_ttl = 24 * 60 * 60
+def sqlite_ystore_factory(
```
dlqqq, line 32: If we do insist on keeping traits on the extension app class and not defining a new subclass of `SQLiteYStore` at file-scope, I would like this function to be renamed to `sqlite_ystore_class_factory`. Factories typically involve creation of objects, not classes (which are technically objects in Python too, but it's better to be as narrow as possible with our names).

jupyter_server_ydoc/handlers.py
```diff
@@ -309,6 +321,7 @@ def on_document_change(self, event):
         self.saving_document = asyncio.create_task(self.maybe_save_document())
 
     async def maybe_save_document(self):
+        assert isinstance(self.room, DocumentRoom)
```
dlqqq, line 324: This doesn't need to be done in this PR, but doing validation imperatively in these methods hinders readability, as we are adding code unrelated to the core purpose of each method. Can this not be done in the constructor? Or if `self.room` is mutable, can we abstract this away in a decorator like `validate_room`?

dlqqq (changes requested):

jupyter_server_ydoc/app.py
davidbrochart: Yes I was also hesitating, I think there is some overlap, but here `sqlite_ystore_document_ttl` is only used to set the corresponding attribute in the YStore class, so it seems there's nothing to validate since nobody else consumes it.

jupyter_server_ydoc/handlers.py
davidbrochart: Yes, I think the name is not well chosen. Actually I don't even think we need a factory. I'll change that :+1:

jupyter_server_ydoc/handlers.py
davidbrochart: It's actually the type checker that was complaining, and for a good reason. The `assert` guides the type checker, but we could also use `cast`. I think it also guides us, since we know that at this point it cannot be a `TransientRoom`.

davidbrochart:
> Small nit, is it possible to hide the CodeCov warnings in the review UI? It's really noisy and makes review more difficult.

Indeed, it's very annoying. Maybe @blink1073 knows?

blink1073:
I think this is the setting you'd put in codecov.yaml: https://docs.codecov.com/docs/codecovyml-reference#github_checks-github-users-only

davidbrochart:
Thanks, actually it's easy to [hide annotations in the file view](https://docs.codecov.com/docs/github-checks#hiding-annotations-in-the-files-view).

dlqqq (approved):
Looks good, you can resolve my other conversations.

PR #82, opened 2023-02-08
Fix releaser job

fcollonval:
This fixes the jupyter releaser job. Actually the culprit was the installation package step.

Configuration to include the `packages` files in sdist without the transient build files is a bit tricky. The final solution is to call `clean:lib` after `build:prod` in the build python step.


hbcarlos (approved):
Thanks, @fcollonval!!

fcollonval:
cc @blink1073 FYI the failure of the jupyter-releaser was due to the step installing the package in the job before executing the check-release action.

blink1073:
Ah, understood, thanks for following up.

PR #88, opened 2023-02-13
Renames python package and lab extension

hbcarlos:


fcollonval (approved):

davidbrochart:
I was wondering if the Python package name should be `jupyter-collaboration` instead of `jupyterlab-collaboration`, since it is independent of JupyterLab.

PR #91, opened 2023-02-17
Rename to `jupyter_collaboration` and `@jupyter/collaboration`

### LINKED ISSUE #89: Package name
I was wondering if the Python package name should be `jupyter-collaboration` instead of `jupyterlab-collaboration`, since it is independent of JupyterLab.

_Originally posted by @davidbrochart in https://github.com/jupyterlab/jupyterlab_collaboration/issues/88#issuecomment-1427904149_

More information to discuss the proper naming:

[@fcollonval] The name `jupyterlab_rtc` is taken on PyPI.org (used to release previous attempt based on automerge). And @bollywl points out using the abbreviation `rtc` is not great to catch the purpose of the package.

fcollonval:
Fixes #89

I use the name `jupyter_collaboration` for Python and the scope `@jupyter` for the NPM package.

As the NPM package are renamed, I reset the version to 1.0.0-alpha.0.

I went for the shorter name and to align NPM and Python package.

> GitHub repo and readthedocs project must be udpated as a follow-up.


hbcarlos (approved):
Thanks, @fcollonval!

The check_links test fails because we need to rename the repository. We can do it once merged.

Issue #92, opened 2023-02-17
not enough arguments for format string


cghanke:
In the debug-statement 
https://github.com/jupyterlab/jupyterlab_collaboration/blob/8d7cd9007bf687f48c0c1f671dcbf00dc6137890/jupyter_collaboration/handlers.py#L161
the parameter "file_id" is apparently missing.

hbcarlos:
Hey @cghanke! Thanks for opening an issue.
You are right. The `file_id` is missing. Do you want to open a PR to fix it?

PR #93, opened 2023-02-17
Add missing argument for format string (#92)

cghanke:
I hope the format is ok.
Otherwise just change it please.


hbcarlos (approved):
Thanks, @cghanke!

CI failures are unrelated.

PR #106, opened 2023-02-21
Keep reference to asyncio tasks

### LINKED ISSUE #105: Keep a reference to the result of create_task()
## Description

A reference to the result of `asyncio.create_task()` should be kept, to avoid a task disappearing mid-execution. See [the documentation](https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task):
image

We do it [here](https://github.com/jupyterlab/jupyter_collaboration/blob/e893233828ea906a71add2b390581dfd764813e4/jupyter_collaboration/handlers.py#L82) and [there](https://github.com/jupyterlab/jupyter_collaboration/blob/e893233828ea906a71add2b390581dfd764813e4/jupyter_collaboration/handlers.py#L198).

davidbrochart:
Fixes #105.


jupyter_collaboration/handlers.py
```diff
@@ -195,7 +199,9 @@ async def open(self, path):
         self.room = self.websocket_server.get_room(path)
         self.set_file_info(path)
         self.saving_document = None
-        asyncio.create_task(self.websocket_server.serve(self))
+        task = asyncio.create_task(self.websocket_server.serve(self))
+        self.background_tasks.add(task)
```
hbcarlos, line 203: Should we initialize the `background_tasks` attribute in the constructor, or are we using the attribute from the `JupyterWebsocketServer`?

hbcarlos (commented):
Thanks, @davidbrochart!

I left a small comment.

jupyter_collaboration/handlers.py
davidbrochart: Good point, it should be in the `JupyterWebsocketServer`.

hbcarlos (approved):

PR #108, opened 2023-02-23
Creates document session

hbcarlos:
Fixes:
* https://github.com/jupyterlab/jupyterlab/issues/13550
* https://github.com/jupyterlab/jupyterlab/issues/14031

### DIFF
--- a/jupyter_collaboration/app.py
+++ b/jupyter_collaboration/app.py
@@ -5,7 +5,12 @@ from jupyter_server.extension.application import ExtensionApp
 from traitlets import Float, Int, Type
 from ypy_websocket.ystore import BaseYStore
 
-from .handlers import SQLiteYStore, YDocRoomIdHandler, YDocWebSocketHandler
+from .handlers import (
+    DocSessionHandler,
+    SQLiteYStore,
+    YDocRoomIdHandler,
+    YDocWebSocketHandler,
+)
 
 
 class YDocExtension(ExtensionApp):
@@ -57,7 +62,11 @@ class YDocExtension(ExtensionApp):
     def initialize_handlers(self):
         self.handlers.extend(
             [
+                # Deprecated - to remove for 1.0.0
                 (r"/api/yjs/roomid/(.*)", YDocRoomIdHandler),
+                # Deprecated - to remove for 1.0.0
                 (r"/api/yjs/(.*)", YDocWebSocketHandler),
+                (r"/api/collaboration/room/(.*)", YDocWebSocketHandler),
+                (r"/api/collaboration/session/(.*)", DocSessionHandler),
             ]
         )
--- a/jupyter_collaboration/handlers.py
+++ b/jupyter_collaboration/handlers.py
@@ -3,6 +3,7 @@
 
 import asyncio
 import json
+import uuid
 from logging import Logger
 from pathlib import Path
 from typing import Any, Dict, Optional, Set, Tuple
@@ -23,6 +24,8 @@ from ypy_websocket.ystore import YDocNotFound
 from ypy_websocket.yutils import YMessageType
 
 YFILE = YDOCS["file"]
+
+DOCUMENT_SESSION = str(uuid.uuid4())
 
 
 class TempFileYStore(_TempFileYStore):
@@ -201,6 +204,11 @@ class YDocWebSocketHandler(WebSocketHandler, JupyterHandler):
         self.websocket_server.background_tasks.add(task)
         task.add_done_callback(self.websocket_server.background_tasks.discard)
 
+        # Close the connection if the document session expired
+        session = self.get_query_argument("session", "")
+        if isinstance(self.room, DocumentRoom) and not DOCUMENT_SESSION == session:
+            self.close(1, "Document session expired")
+
         # cancel the deletion of the room if it was scheduled
         if isinstance(self.room, DocumentRoom) and self.room.cleaner is not None:
             self.room.cleaner.cancel()
@@ -227,6 +235,7 @@ class YDocWebSocketHandler(WebSocketHandler, JupyterHandler):
                     # if YStore updates and source file are out-of-sync, resync updates with source
                     if self.room.document.source != model["content"]:
                         read_from_source = True
+
                 if read_from_source:
                     self.room.document.source = model["content"]
                     if self.room.ystore:
@@ -385,18 +394,15 @@ class YDocRoomIdHandler(APIHandler):
     @web.authenticated
     @authorized
     async def put(self, path):
-        body = json.loads(self.request.body)
-        ws_url = f"{body['format']}:{body['type']}:"
-
         file_id_manager = self.settings["file_id_manager"]
 
         idx = file_id_manager.get_id(path)
         if idx is not None:
             # index already exists
             self.set_status(200)
-            ws_url += str(idx)
-            self.log.info("Request for Y document '%s' with room ID: %s", path, ws_url)
-            return self.finish(ws_url)
+            self.log.info("Request for Y document '%s' with room ID: %s", path, idx)
+            data = json.dumps({"file_id": idx, "session": DOCUMENT_SESSION})
+            return self.finish(data)
 
         # try indexing
         idx = file_id_manager.index(path)
@@ -406,6 +412,6 @@ class YDocRoomIdHandler(APIHandler):
 
         # index successfully created
         self.set_status(201)
-        ws_url += str(idx)
-        self.log.info("Request for Y document '%s' with room ID: %s", path, ws_url)
-        return self.finish(ws_url)
+        self.log.info("Request for Y document '%s' with room ID: %s", path, idx)
+        data = json.dumps({"file_id": idx, "session": DOCUMENT_SESSION})
+        return self.finish(data)
--- a/jupyter_collaboration/handlers.py
+++ b/jupyter_collaboration/handlers.py
@@ -206,7 +206,7 @@ class YDocWebSocketHandler(WebSocketHandler, JupyterHandler):
 
         # Close the connection if the document session expired
         session = self.get_query_argument("session", "")
-        if isinstance(self.room, DocumentRoom) and not DOCUMENT_SESSION == session:
+        if isinstance(self.room, DocumentRoom) and DOCUMENT_SESSION != session:
             self.close(1, "Document session expired")
 
         # cancel the deletion of the room if it was scheduled
--- a/jupyter_collaboration/handlers.py
+++ b/jupyter_collaboration/handlers.py
@@ -25,7 +25,7 @@ from ypy_websocket.yutils import YMessageType
 
 YFILE = YDOCS["file"]
 
-DOCUMENT_SESSION = str(uuid.uuid4())
+SERVER_SESSION = str(uuid.uuid4())
 
 
 class TempFileYStore(_TempFileYStore):
@@ -205,9 +205,9 @@ class YDocWebSocketHandler(WebSocketHandler, JupyterHandler):
         task.add_done_callback(self.websocket_server.background_tasks.discard)
 
         # Close the connection if the document session expired
-        session = self.get_query_argument("session", "")
-        if isinstance(self.room, DocumentRoom) and DOCUMENT_SESSION != session:
-            self.close(1, "Document session expired")
+        session_id = self.get_query_argument("sessionId", "")
+        if isinstance(self.room, DocumentRoom) and SERVER_SESSION != session_id:
+            self.close(1003, f"Document session {session_id} expired")
 
         # cancel the deletion of the room if it was scheduled
         if isinstance(self.room, DocumentRoom) and self.room.cleaner is not None:
@@ -394,15 +394,18 @@ class YDocRoomIdHandler(APIHandler):
     @web.authenticated
     @authorized
     async def put(self, path):
+        body = json.loads(self.request.body)
+        ws_url = f"{body['format']}:{body['type']}:"
+
         file_id_manager = self.settings["file_id_manager"]
 
         idx = file_id_manager.get_id(path)
         if idx is not None:
             # index already exists
             self.set_status(200)
-            self.log.info("Request for Y document '%s' with room ID: %s", path, idx)
-            data = json.dumps({"file_id": idx, "session": DOCUMENT_SESSION})
-            return self.finish(data)
+            ws_url += str(idx)
+            self.log.info("Request for Y document '%s' with room ID: %s", path, ws_url)
+            return self.finish(ws_url)
 
         # try indexing
         idx = file_id_manager.index(path)
@@ -412,6 +415,42 @@ class YDocRoomIdHandler(APIHandler):
 
         # index successfully created
         self.set_status(201)
+        ws_url += str(idx)
+        self.log.info("Request for Y document '%s' with room ID: %s", path, ws_url)
+        return self.finish(ws_url)
+
+
+class DocSessionHandler(APIHandler):
+    auth_resource = "contents"
+
+    @web.authenticated
+    @authorized
+    async def put(self, path):
+        body = json.loads(self.request.body)
+        format = body["format"]
+        content_type = body["type"]
+        file_id_manager = self.settings["file_id_manager"]
+
+        idx = file_id_manager.get_id(path)
+        if idx is not None:
+            # index already exists
+            self.log.info("Request for Y document '%s' with room ID: %s", path, idx)
+            data = json.dumps(
+                {"format": format, "type": content_type, "fileId": idx, "sessionId": SERVER_SESSION}
+            )
+            self.set_status(200)
+            return self.finish(data)
+
+        # try indexing
+        idx = file_id_manager.index(path)
+        if idx is None:
+            # file does not exists
+            raise web.HTTPError(404, f"File {path!r} does not exist")
+
+        # index successfully created
         self.log.info("Request for Y document '%s' with room ID: %s", path, idx)
-        data = json.dumps({"file_id": idx, "session": DOCUMENT_SESSION})
+        data = json.dumps(
+            {"format": format, "type": content_type, "fileId": idx, "sessionId": SERVER_SESSION}
+        )
+        self.set_status(201)
         return self.finish(data)
--- a/packages/docprovider/src/yprovider.ts
+++ b/packages/docprovider/src/yprovider.ts
@@ -4,14 +4,17 @@
 |----------------------------------------------------------------------------*/
 
 import { URLExt } from '@jupyterlab/coreutils';
+import { showErrorMessage, Dialog } from '@jupyterlab/apputils';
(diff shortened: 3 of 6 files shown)


jupyter_collaboration/handlers.py
```diff
@@ -201,6 +204,11 @@ async def open(self, path):
         self.websocket_server.background_tasks.add(task)
         task.add_done_callback(self.websocket_server.background_tasks.discard)
 
+        # Close the connection if the document session expired
+        session = self.get_query_argument("session", "")
+        if isinstance(self.room, DocumentRoom) and not DOCUMENT_SESSION == session:
```
davidbrochart, line 209:
```suggestion
        if isinstance(self.room, DocumentRoom) and DOCUMENT_SESSION != session:
```

jupyter_collaboration/handlers.py
```diff
@@ -385,18 +394,15 @@ class YDocRoomIdHandler(APIHandler):
...
 
         idx = file_id_manager.get_id(path)
         if idx is not None:
             # index already exists
             self.set_status(200)
-            ws_url += str(idx)
-            self.log.info("Request for Y document '%s' with room ID: %s", path, ws_url)
-            return self.finish(ws_url)
+            self.log.info("Request for Y document '%s' with room ID: %s", path, idx)
+            data = json.dumps({"file_id": idx, "session": DOCUMENT_SESSION})
```
davidbrochart, line 404: Why did you remove `format:type:` from `file_id`?

hbcarlos: Because we are not using it.

The WebSocketProvider requests the file id and sends the format and type in that request, then the backend retrieves the file id, composes the room id with `format:type:file_id`, and returns the room id. We have that info in the front. We only need the file id. We cannot even retrieve the file id when we don't have the format and type. So I don't think sending that info just to compose the room id in the backend makes sense. I believe it is better to retrieve the file id and compose the room id on the client side

hbcarlos: Now I'm thinking that maybe is better to use a get request since we only need to retrieve the file id.

davidbrochart: The `PUT /api/yjs/roomid/{path}` endpoint is for getting a room ID for a document path, format and type. What is returned is an implementation detail, it could also be a pure UUID. If it returns only the file ID, we cannot distinguish between a document e.g. as a plain file or as a notebook type.

davidbrochart: > Now I'm thinking that maybe is better to use a get request since we only need to retrieve the file id.

This has been discussed and we decided to use ` PUT`.

hbcarlos: Ooh, I got it. That makes sense.

davidbrochart: Let's not mix everything, `PUT /api/yjs/roomid/{path}` is for getting a room ID, as its name says.
Shared model version may be implemented using WebSocket subprotocols. Store version is independent.
BTW right now your implementation looks more like a "server session" than a "document session", since it's global. But it should be document-based, because updates in a YStore can be squashed on a document-based basis, right?

packages/docprovider/src/yprovider.ts
```diff
@@ -121,14 +128,31 @@ export class WebSocketProvider implements IDocumentProvider {
     this._awareness.setLocalStateField('user', user.identity);
   }
 
+  private _onConnectionClosed(event: any): void {
+    if (event.code === 1) {
+      console.error('Document provider closed:', event.reason);
+
+      showErrorMessage(
+        'Session expired',
+        'The document session expired. We need to reload this browser tab.',
```
fcollonval, line 137:
```suggestion
        this.trans.__('The document session expired. You need to reload this browser tab.'),
```

packages/docprovider/src/yprovider.ts
```diff
@@ -121,14 +128,31 @@ export class WebSocketProvider implements IDocumentProvider {
     this._awareness.setLocalStateField('user', user.identity);
   }
 
+  private _onConnectionClosed(event: any): void {
+    if (event.code === 1) {
+      console.error('Document provider closed:', event.reason);
+
+      showErrorMessage(
+        'Session expired',
```
fcollonval, line 136:
```suggestion
        this.trans.__('Session expired'),
```

packages/docprovider/src/yprovider.ts
```diff
@@ -121,14 +128,31 @@ export class WebSocketProvider implements IDocumentProvider {
...
+        'The document session expired. We need to reload this browser tab.',
+        [Dialog.okButton()]
```
fcollonval, line 138:
```suggestion
        [Dialog.okButton({ label: this.trans.__('Reload) })]
```

fcollonval (changes requested):
Thanks @hbcarlos 

In addition to the point raised by David, could you pass the `ITranslator` object to the provider to the error message can be translated?

I guess you need to pass it through the `IDrive` at

https://github.com/jupyterlab/jupyter_collaboration/blob/d8481ce01934a671e5da049f31062bbbb6bccd56/packages/collaboration-extension/src/filebrowser.ts#L44

jupyter_collaboration/handlers.py
hbcarlos: > BTW right now your implementation looks more like a "server session" than a "document session", since it's global. But it should be document-based, because updates in a YStore can be squashed on a document-based basis, right?

This implementation is a server session to patch, as soon as possible, the issue where the content gets duplicated.

This implementation should be revised for JupyterLab v4. We should design a robust architecture with room sessions that contemplate squashing or removing the YStore, and checks the shared model's version.

> Let's not mix everything, `PUT /api/yjs/roomid/{path}` is for getting a room ID, as its name says.

In my opinion, everything is related. Maybe we should change this endpoint and create a more generic one to fetch room info.

When a client wants to join a collaborative session (a DocumentRoom), this client should meet some requirements.
* a compatible shared model
* non-outdated updates (after a squash or deletion of the store)

We could have different clients connecting to the same room (JupyterLab, Notebook, JupyterLite, etc.), and we should check those requirements before establishing a connection.

We could create a room session so every client that supports the requirement of that room can join. Using a more generic endpoint, we can fetch those requirements simultaneously (all of them are related and necessary to connect to a room). In addition, we need to check the requirement every time a client tries to join a collaborative session. As we see with the content duplication issue, the user may have old tabs open, and they will try to reconnect immediately. Anything could happen between a reconnection, even a new store version.

davidbrochart: All that is true, but I'm not sure we should change `PUT /api/yjs/roomid/{path}`. We could, however, accept the Y-WebSocket connection if all the requirements are met.

hbcarlos: And how do we request the requirement? Should we create a new endpoint?

jupyter_collaboration/handlers.py
davidbrochart, line 209:
```suggestion
        if isinstance(self.room, DocumentRoom) and DOCUMENT_SESSION != session:
```

jupyter_collaboration/handlers.py
davidbrochart: Maybe using WebSocket subprotocols for the Y-WebSocket? The client sends everything it supports, and the server checks them and accepts or closes the connection.

fcollonval: > The client will unlikely send `everything` it supports because it will need to come with a specific version of `jupyter_ydoc` (whatever the language) and any doc provider. And I don't see us going into the trouble of supporting multiple versions within the same package. So my guess is that this will be simplified to _the client sends the "version" it supports_.

davidbrochart: I'm not sure supporting multiple versions is such an issue. The problem I see with the client sending only one supported version is that it could quickly lead to incompatible clients. But I'm not sure either this will be the case.

hbcarlos: > Maybe using WebSocket subprotocols for the Y-WebSocket? The client sends everything it supports, and the server checks them and accepts or closes the connection.

I don't think using subprotocols is the solution. In my opinion, we should validate the requirements before opening a connection.

After investigating a little bit, I believe the only way we can clean and reset a client, it is by disposing the shared model (destroying the YDoc and YWebSocketProvider) and instantiating a new one. I don't think it is possible to remove the updates from a YDoc and reinitialize it again (It doesn't make sense either).

The document session requirements are the information we need before establishing the communication. Once clients start communicating, those requirements can not change. Even in the case of squashing updates in the Store, we can only do it when the room is empty. Otherwise, we will need to:
1 - communicate to every client that there was a squash in the document, then on each client
2 - we will need to close the connection
3 - dispose the resources
4 - instantiate a new shared document
5 - open a new connection
That's inconvenient from a user perspective. Imagine you are working on a document, and every X minutes, the document closes (or asks you to close it) and reopens.

I also thought using subprotocols was the best solution, but after debugging this issue and investigating a little, I don't think it is the right solution.

davidbrochart: > I don't think using subprotocols is the solution. In my opinion, we should validate the requirements before opening a connection.

With WebSocket subprotocols, the requirements are validated before opening the connection.

davidbrochart:
@hbcarlos do you need help to finish this PR? I think it's almost ready, as a quick fix to the duplication issue, and then we can iterate and improve on it. You just need to revert the changes to the `PUT /api/yjs/roomid/{path}` handler (keeping the returned session of course).

jupyter_collaboration/app.py
```diff
@@ -59,5 +64,7 @@ def initialize_handlers(self):
             [
                 (r"/api/yjs/roomid/(.*)", YDocRoomIdHandler),
                 (r"/api/yjs/(.*)", YDocWebSocketHandler),
+                (r"/api/collaboration/room/(.*)", YDocWebSocketHandler),
```
davidbrochart, line 67: Why is there two endpoints with the same handler?

jupyter_collaboration/app.py
```diff
@@ -59,5 +64,7 @@ def initialize_handlers(self):
             [
                 (r"/api/yjs/roomid/(.*)", YDocRoomIdHandler),
```
fcollonval, line 65:
```suggestion
                # Deprecated - to remove for 1.0.0
                (r"/api/yjs/roomid/(.*)", YDocRoomIdHandler),
                # Deprecated - to remove for 1.0.0
```

fcollonval (changes requested):
Thanks @hbcarlos 

The new endpoints are better as they are based on meaning rather than technology and avoid the need to deal with regex order. Moreover as we need to backport this, this will allow to not break the client installation as for JLab 3.6.x, the user may end up calling the previous roomid not expecting a JSON answer.

There is naming convention that requires some changes but otherwise it is good to go.

jupyter_collaboration/app.py
```diff
@@ -59,5 +64,7 @@ def initialize_handlers(self):
...
                 (r"/api/yjs/(.*)", YDocWebSocketHandler),
+                (r"/api/collaboration/room/(.*)", YDocWebSocketHandler),
+                (r"/api/collaboration/session/(.*)", DocSessionHandler),
```
davidbrochart, line 68: I don't think we should grow the API surface.
I think we should use the already existing `PUT /api/yjs/roomid/{path}` endpoint. It could return a value that is a combination of the file ID, the document type and format, and the session ID. That way, the frontend doesn't have to ask for a session ID, all it asks for is a room ID. When it tries to connect to the WebSocket with a room ID that doesn't exit (because the server session has changed), the room cannot be found and the frontend asks for the room ID again, discarding the shared document that it may have.

davidbrochart (changes requested):

jupyter_collaboration/app.py
hbcarlos: The yjs endpoint is deprecated in favor of a more explicit name. Yjs is a framework, not an action or a resource.

jupyter_collaboration/app.py
davidbrochart: By combination, I don't mean necessarily concatenating multiple attributes, it can be a UUID and the server keeps a mapping of this UUID to the attributes.

fcollonval (approved):

jupyter_collaboration/app.py
davidbrochart: It is easy for the frontend to know that the session has changed: it fails to connect with its room ID. It can now ask for the room ID again, if that fails it's because the document doesn't exist, otherwise it's because the server has restarted, and in that case it can tell the user to not use the tab.

jupyter_collaboration/handlers.py
```diff
@@ -56,6 +59,7 @@ class DocumentRoom(YRoom):
     def __init__(self, type: str, ystore: BaseYStore, log: Optional[Logger]):
         super().__init__(ready=False, ystore=ystore, log=log)
         self.type = type
+        self.session_id = str(uuid.uuid4())
```
davidbrochart, line 62: What is this for? It doesn't seem to be used.

hbcarlos: Ouch, good catch. That's a left over form some experiments I did.

davidbrochart:
So to summarize, this PR (correct me if I'm wrong) introduces a new session ID on top of the room ID, and is a breaking change.
While what I'm suggesting it to only deal with a room ID, and not break the API.
I'm not sure a new session ID is needed, so we should make sure it is before merging and breaking the API.

hbcarlos:
> So to summarize, this PR (correct me if I'm wrong) introduces a new session ID on top of the room ID, and is a breaking change.

It is not a breaking change. We did not change any API. We are adding a new endpoint to retrieve session information next to the room id.

> I'm not sure a new session ID is needed

Why is not needed anymore?

davidbrochart:
> It is not a breaking change. We did not change any API. We are adding a new endpoint to retrieve session information next to the room id.

But you are [deprecating the old API](https://github.com/jupyterlab/jupyter_collaboration/pull/108/files#diff-e751c32650f4de25bc914afd58e459e1944a11c79e823bd494a8bec1d360744cR65-R67), isn't that a breaking change?

> Why is not needed anymore?

Because we can do everything with the already existing room ID (see [my comment](https://github.com/jupyterlab/jupyter_collaboration/pull/108#discussion_r1118405263)).

jupyter_collaboration/app.py
fcollonval: Implicit deduction is never a good path for code evolution, maintenance and understanding.

fcollonval:
Ok let's move forward keeping a single opaque identifier. This will likely induce the need for storing a mapping between the identifier and the room + session information to make the bridge between the request of the single identifier and the request of the document connection.

davidbrochart:
@hbcarlos mentioned the issue of transient rooms which don't require a room ID, so we cannot rely on room IDs to support sessions for them. I cannot think of a solution for them, that wouldn't involve a session ID, so let's got with explicit session ID and merge this PR.
Thanks Carlos for exploring all possible solutions.

hbcarlos:
Thanks, @fcollonval and @davidbrochart, for looking into this PR.

I reverted my latest commits to come back to the session ID implementation.

davidbrochart:
I just tested that PR on JupyterLab `master` with the following steps:
- Launcher JupyterLab, create a notebook
- Open a new browser window and open the same notebook
- Stop and restart the server, but don't close the two browser windows
- A window detects that the session has changed and asks to reload the page, accept
- A blank notebook appears (no cell)

@hbcarlos have you tried that scenario?

davidbrochart:
It seems to be a `jupyter_ydoc` version issue because I can see these errors in the backend:
```
[E 2023-03-01 14:34:27.379 ServerApp] Uncaught exception GET /api/collaboration/room/json:notebook:01aac1bc-a0c5-42c7-8ac9-6a147601e341?sessionId=6a88536d-04df-4935-9bf2-549b6254c915 (127.0.0.1)
    HTTPServerRequest(protocol='http', host='127.0.0.1:8888', method='GET', uri='/api/collaboration/room/json:notebook:01aac1bc-a0c5-42c7-8ac9-6a147601e341?sessionId=6a88536d-04df-4935-9bf2-549b6254c915', version='HTTP/1.1', remote_ip='127.0.0.1')
      File "/home/david/micromamba/envs/jupyterlab/lib/python3.11/site-packages/tornado/websocket.py", line 944, in _accept_connection
        await open_result
      File "/home/david/github/davidbrochart/jupyter_collaboration/jupyter_collaboration/handlers.py", line 236, in open
        if self.room.document.source != model["content"]:
           ^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/home/david/micromamba/envs/jupyterlab/lib/python3.11/site-packages/jupyter_ydoc/ybasedoc.py", line 60, in source
        return self.get()
               ^^^^^^^^^^
      File "/home/david/micromamba/envs/jupyterlab/lib/python3.11/site-packages/jupyter_ydoc/ynotebook.py", line 213, in get
        metadata=meta["metadata"],
                 ~~~~^^^^^^^^^^^^
    KeyError: 'metadata'
```

davidbrochart:
It seemed to be a `.jupyter_ystore.db` issue, it works fine after removing it :+1:

jupyter_collaboration/handlers.py
```diff
@@ -409,3 +418,39 @@ async def put(self, path):
...
+        idx = file_id_manager.index(path)
+        if idx is None:
+            # file does not exists
+            raise web.HTTPError(404, f"File {path!r} does not exist")
+
+        # index successfully created
+        self.log.info("Request for Y document '%s' with room ID: %s", path, idx)
+        data = json.dumps(
+            {"format": format, "type": content_type, "fileId": idx, "sessionId": SERVER_SESSION}
+        )
```
davidbrochart, line 454: Maybe have this code only once, since it's duplicated on lines 437-438?

davidbrochart (approved):

davidbrochart:
Thanks @hbcarlos !

fperez:
Quick question - if we want to use this to mitigate the document corruption problems, do we need to wait for a 3.6.2 release or can we update a separate subpackage? I'm not sure whether any of this code goes into the core lab package or not, sorry!

This is on the critical path for us to [test RTC in teaching settings at Berkeley](https://github.com/berkeley-dsep-infra/datahub/issues/4223), so I'm trying to figure out what the order of operations is for us. Thanks for the great work!

fperez:
Ah, I just realized that #14128 may be the answer to my question, sorry for the noise. In that case I presume a 3.6.2 is coming soon with these fixes! 🎉

hbcarlos:
> Ah, I just realized that https://github.com/jupyterlab/jupyterlab/pull/14128 may be the answer to my question, sorry for the noise. In that case I presume a 3.6.2 is coming soon with these fixes! 🎉

Yep, that is fixed and backported to lab 3.6 we just need to wait for the 3.6.2 release.

PR #113, opened 2023-03-01
Backport PR #108 on branch 0.8.x (Creates document session)

hbcarlos:


jupyter_server_ydoc/app.py
```diff
@@ -62,7 +67,11 @@ def initialize_settings(self):
     def initialize_handlers(self):
         self.handlers.extend(
             [
+                # Deprecated - to remove for 1.0.0
                 (r"/api/yjs/roomid/(.*)", YDocRoomIdHandler),
+                # Deprecated - to remove for 1.0.0
+                (r"/api/yjs/session/(.*)", DocSessionHandler),
+                # Deprecated - to remove for 1.0.0
                 (r"/api/yjs/(.*)", YDocWebSocketHandler),
```
fcollonval, line 75: We should introduce the new endpoint as deprecation is about ensuring backward compatibility but yet allow to migrate to the newer compatible alternative. So I would also not introduce `/api/yjs/session/(.*)` but only `/api/collaboration/session/(.*)`.

fcollonval (changes requested):
Thanks @hbcarlos

jupyter_server_ydoc/app.py
hbcarlos: There are concerns regarding the update from JupyterLab v3.6 (jupyter-server-ydoc) to JupyterLab v4 (jupyter_collaboration).

Both are jupyter-server extensions registering the same endpoints. If someone just updates JupyterLab without cleaning the environment (removing jupyter-server-ydoc) we can not ensure the endpoint used will be the one from jupyter_collaboration.

That's why I used the old name even for the new endpoint.

hbcarlos: See [this comment](https://github.com/jupyterlab/jupyterlab/issues/14031#issuecomment-1445164761)

fcollonval: Good point - then let's go that road

jupyter_server_ydoc/__init__.py
```diff
@@ -2,7 +2,7 @@
 
 from .app import YDocExtension
 
-__version__ = "0.6.1"
+__version__ = "0.6.2"
```
fcollonval, line 5:
```suggestion
__version__ = "0.8.0"
```

fcollonval (approved):
Thanks @hbcarlos 

> I bump already to 0.8.0 to avoid any trouble :wink:

jupyter_server_ydoc/__init__.py
```diff
@@ -2,7 +2,7 @@
 
 from .app import YDocExtension
 
-__version__ = "0.6.1"
+__version__ = "0.8.0"
```
davidbrochart, line 5: Isn't it the role of the releaser to bump the version?

davidbrochart (approved):

jupyter_server_ydoc/__init__.py
davidbrochart: You can give the releaser the branch and specify the version (instead of `next`), so I would not change the version here.

PR #117, opened 2023-03-06
Fixes experimental document-wide undo/redo flag

hbcarlos:
Fixes experimental document-wide undo/redo flag. https://github.com/jupyterlab/jupyterlab/pull/14063
Fixes https://github.com/jupyterlab/jupyterlab/pull/14128#discussion_r1126065742.

Commented by @fcollonval in https://github.com/jupyterlab/jupyterlab/pull/14128#discussion_r1126065742
> I oversight this one in jupyter_collaboration. But we should be consistent and not reload if the user did not want that (for example that will happen if it clicks outside the dialog)


packages/collaboration-extension/src/filebrowser.ts
```diff
@@ -60,6 +67,28 @@ export const defaultFileBrowser: JupyterFrontEndPlugin<IDefaultFileBrowser> = {
...
+    if (settingRegistry) {
+      settingRegistry
+        .load('@jupyterlab/notebook-extension:tracker')
+        .then(settings => {
+          const updateSettings = (settings: ISettingRegistry.ISettings) => {
+            const enableDocWideUndo = settings?.get(
+              'experimentalEnableDocumentWideUndoRedo'
+            ).composite as boolean;
+            const disableDocumentWideUndo = !enableDocWideUndo ?? true;
+            drive.sharedModelFactory.disableDocumentWideUndo =
```
fcollonval, line 81: What do you think of introducing a generic pattern:

```ts
drive.sharedModelFactory.documentOptions: Map<string, Record<string, any>>
```

The key of the map could be the document type directly instead of string I guess. And at the instantiation of the shared models, you will pass the options.

Something like:

```ts
      case 'notebook':
        sharedModel = new YNotebook(this.documentOptions.get('notebook'));
```

fcollonval (changes requested):
Thanks a lot @hbcarlos 

I have suggestion for the options storage.

packages/collaboration-extension/src/filebrowser.ts
hbcarlos: Sound good! I'm on it.

fcollonval (approved):
Thanks @hbcarlos

PR #118, opened 2023-03-08
Remove a console.debug for plugin activate

jtpio:
Not sure we need to keep this `console.debug` for shipping the final version of this extension.

Unless it can be considered useful for debugging.


fcollonval (approved):
Thanks @jtpio

PR #121, opened 2023-03-14
Creates a new FileLoader class to separate the logic of watching files

hbcarlos:
Creates a new FileLoader class to separate the logic of watching files from the WebSocketHandler.

# Code changes:
* Split classes in different files
* Creates a new FileLoader class to centralize loading, saving, and watching files.
   Multiple rooms can get an instance of this class and use it to access the content safely since we can create a single lock by file to ensure two rooms are not accessing the same file simultaneously.
* Updates the DocumentRoom to use the FileLoader.

# TODO:
- [x] Create an abstract class for FileLoader to extend from.
    * Some extensions (like jupytercad) may want their own loader.
    * Allow registering new file loaders.
- [ ] Add documentation

### DIFF
--- a/jupyter_collaboration/handlers.py
+++ b/jupyter_collaboration/handlers.py
@@ -55,7 +55,7 @@ class JupyterWebsocketServer(WebsocketServer):
             self.ypatch_nb = 0
 
     def room_exists(self, path: str) -> bool:
-        return path in self.rooms
+        return path in self.rooms.keys()
 
     def add_room(self, path: str, room: YRoom) -> None:
         self.rooms[path] = room
@@ -94,7 +94,7 @@ class YDocWebSocketHandler(WebSocketHandler, JupyterHandler):
 
     def __init__(self, app: ServerWebApplication, request: HTTPServerRequest, **kwargs):
         super().__init__(app, request, **kwargs)
-        print("[__init__] args:", app, request, kwargs)
+        print("\n[YDocWebSocketHandler.__init__] args:", app, request, kwargs)
 
         # CONFIG
         self._file_id_manager = self.settings["file_id_manager"]
@@ -106,7 +106,7 @@ class YDocWebSocketHandler(WebSocketHandler, JupyterHandler):
         # Instantiate the JupyterWebsocketServer as a Class property
         # if it doesn't exist yet
         if self.websocket_server is None:
-            print("[DocumentWebSocketHandler] First client:")
+            print("\t[DocumentWebSocketHandler] First client:")
             for k, v in self.config.get(self._ystore_class.__name__, {}).items():
                 setattr(self._ystore_class, k, v)
             
@@ -117,27 +117,32 @@ class YDocWebSocketHandler(WebSocketHandler, JupyterHandler):
                 log=self.log,
             )
         
+        self._message_queue = asyncio.Queue()
+        
         # Get room
         self._room_id: str = request.path.split("/")[-1]
-        print("ROOM ID:", self._room_id)
+        print("\tROOM ID:", self._room_id)
 
         if self.websocket_server.room_exists(self._room_id):
             self.room: YRoom = self.websocket_server.get_room(self._room_id)
+            print("\tROOM exists:", self.room.room_id)
         
         else :
             if self._room_id.count(":") >= 2:
                 # DocumentRoom
-                _, file_type, file_id = decode_file_path(self._room_id)
+                file_format, file_type, file_id = decode_file_path(self._room_id)
                 self._path = self._file_id_manager.get_path(file_id)
-                print("File path:", self._path)
+                print("\tFile: path", self._path)
+                print("\tFile: format,type,id", file_format, file_type, file_id)
                 
                 # Instantiate the FileLoader if it doesn't exist yet
                 file = YDocWebSocketHandler.files.get(self._path)
                 if file is None:
                     file = FileLoader(
                         self._path,
+                        file_format,
+                        file_type,
                         self.contents_manager,
-                        self.settings["collaborative_document_save_delay"],
                         self.settings["collaborative_file_poll_interval"]
                     )
                     self.files[self._path] = file
@@ -145,24 +150,27 @@ class YDocWebSocketHandler(WebSocketHandler, JupyterHandler):
                 path = Path(self._path)
                 updates_file_path = str(path.parent / f".{file_type}:{path.name}.y")
                 ystore = self._ystore_class(path=updates_file_path, log=self.log)
-                self.room = DocumentRoom(self._room_id, file, ystore, self.log)
+                self.room = DocumentRoom(
+                    self._room_id,
+                    file_format,
+                    file_type,
+                    file,
+                    ystore,
+                    self.log,
+                    self.settings["collaborative_document_save_delay"]
+                )
             
             else :
                 # TransientRoom
                 # it is a transient document (e.g. awareness)
-                self.room = TransientRoom(self.log)
+                self.room = TransientRoom(self._room_id, self.log)
             
             self.websocket_server.add_room(self._room_id, self.room)
-            print("room:", self._room_id, self.room.ready)
-        
-        self._message_queue = asyncio.Queue()
-
-        task = asyncio.create_task(self.websocket_server.serve(self))
-        self.websocket_server.background_tasks.add(task)
-        task.add_done_callback(self.websocket_server.background_tasks.discard)
+            print("\troom:", self._room_id, self.room.ready)
 
     @property
     def path(self):
+        # needed to be compatible with WebsocketServer (websocket.path)
         return self._room_id
     
     # Override max_message size to 1GB
@@ -188,7 +196,12 @@ class YDocWebSocketHandler(WebSocketHandler, JupyterHandler):
         return await super().get(*args, **kwargs)
 
     async def open(self, path):
-        #self.set_file_info(path)
+        print("\n[YDocWebSocketHandler.open] path:", path)
+        print("\tROOM_ID:", self._room_id)
+
+        task = asyncio.create_task(self.websocket_server.serve(self))
+        self.websocket_server.background_tasks.add(task)
+        task.add_done_callback(self.websocket_server.background_tasks.discard)
         
         # Close the connection if the document session expired
         session_id = self.get_query_argument("sessionId", "")
@@ -240,6 +253,7 @@ class YDocWebSocketHandler(WebSocketHandler, JupyterHandler):
                     YMessageType(message_type).name,
                 )
                 return skip
+        
         self._message_queue.put_nowait(message)
         self.websocket_server.ypatch_nb += 1
 
@@ -264,7 +278,7 @@ class YDocWebSocketHandler(WebSocketHandler, JupyterHandler):
 
         # Clean the file loader
         file = self.files[self._path]
-        if file.number_of_rooms() == 0 :
+        if file.number_of_subscriptions() == 0 :
             del self.files[self._path]
 
         # Remove the room from the websocket server
--- a/jupyter_collaboration/handlers.py
+++ b/jupyter_collaboration/handlers.py
@@ -54,7 +54,7 @@ class JupyterWebsocketServer(WebsocketServer):
             self.ypatch_nb = 0
 
     def room_exists(self, path: str) -> bool:
-        return path in self.rooms.keys()
+        return path in self.rooms
 
     def add_room(self, path: str, room: YRoom) -> None:
         self.rooms[path] = room
--- a/jupyter_collaboration/handlers.py
+++ b/jupyter_collaboration/handlers.py
@@ -43,28 +43,61 @@ class JupyterWebsocketServer(WebsocketServer):
         self.ypatch_nb = 0
         self.connected_users = {}
         self.background_tasks = set()
-        self.monitor_task = asyncio.create_task(self.monitor())
+        self.monitor_task = asyncio.create_task(self._monitor())
 
-    async def monitor(self):
+    def room_exists(self, path: str) -> bool:
+        """
+        Returns true is the room exist or false otherwise.
+
+            Parameters:
+                path (str): Room ID.
+
+            Returns:
+                exists (bool): Whether the room exists or not.
+        """
+        return path in self.rooms
+
+    def add_room(self, path: str, room: YRoom) -> None:
+        """
+        Adds a new room.
+
+            Parameters:
+                path (str): Room ID.
+                room (YRoom): A room.
+        """
+        self.rooms[path] = room
+
+    def get_room(self, path: str) -> YRoom:
+        """
+        Returns the room for the specified room ID or raises a RoomNotFound
+        error if the room doesn't exist.
+
+            Parameters:
+                path (str): Room ID.
+
+            Returns:
+                room (YRoom): The room.
+        """
+        if path not in self.rooms:
+            # Document rooms need a file
+            raise RoomNotFound
+
+        return self.rooms[path]
+
+    async def _monitor(self):
+        """
+        An infinite loop with a 60 seconds delay for counting the number
+        of patches processed in a minute and how many clients are connected.
+
+        #### Note:
+            This method runs in a coroutine for debugging purposes.
+        """
         while True:
             await asyncio.sleep(60)
             clients_nb = sum(len(room.clients) for room in self.rooms.values())
             self.log.info("Processed %s Y patches in one minute", self.ypatch_nb)
             self.log.info("Connected Y users: %s", clients_nb)
             self.ypatch_nb = 0
-
-    def room_exists(self, path: str) -> bool:
-        return path in self.rooms
-
-    def add_room(self, path: str, room: YRoom) -> None:
-        self.rooms[path] = room
-
-    def get_room(self, path: str) -> YRoom:
-        if path not in self.rooms:
-            # Document rooms need a file
-            raise RoomNotFound
-
-        return self.rooms[path]
 
 
 class YDocWebSocketHandler(WebSocketHandler, JupyterHandler):
--- a/jupyter_collaboration/handlers.py
+++ b/jupyter_collaboration/handlers.py
@@ -7,25 +7,23 @@ import uuid
 from pathlib import Path
 from typing import Any, Dict, Optional, Set
 
+from jupyter_server.auth import authorized
+from jupyter_server.base.handlers import APIHandler, JupyterHandler
+from jupyter_server.serverapp import ServerWebApplication
+from jupyter_ydoc import ydocs as YDOCS
 from tornado import web
 from tornado.httputil import HTTPServerRequest
 from tornado.websocket import WebSocketHandler
-
-from jupyter_server.auth import authorized
-from jupyter_server.serverapp import ServerWebApplication
-from jupyter_server.base.handlers import APIHandler, JupyterHandler
-
-from jupyter_ydoc import ydocs as YDOCS
-
 from ypy_websocket.websocket_server import WebsocketServer, YRoom
 from ypy_websocket.yutils import YMessageType
 
-from .rooms import DocumentRoom, TransientRoom, FileLoader
+from .rooms import DocumentRoom, FileLoader, TransientRoom
 from .utils import decode_file_path
 
 YFILE = YDOCS["file"]
 
 SERVER_SESSION = str(uuid.uuid4())
+
 
 class RoomNotFound(Exception):
     pass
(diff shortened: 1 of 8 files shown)


hbcarlos:
It would be great to have a way to swap loaders so extensions can register new document types and the logic to load those documents. For example, JupyterCAD registers a new document type that uses the FreeCAD format. This format is a compressed folder with multiple XML files. Jupyter Server needs to learn how to open those documents, and the contents service needs a way to plug new loaders.

I have looked into the contents service of Jupyter Server, trying to find the best solution for this problem. Currently, the ContentsManager only allows registering pre/post save hooks. We could use these hooks to load the content of a specific file type, but we can not ensure that the hook we register will be executed the last one. In addition, there are only hooks for saving the file. There are no hooks for loading the file.

I'm not trying to solve the problem in this PR, but I want to redesign the architecture of jupyter_collaboration mindful of this problem to solve it in a follow-up PR.

At the moment, I see two options:
1 - Create a new file loader manager that allows registering loaders depending on the file types, and create the default FileLoader using Jupyter Server's content manager as we currently do.
2 - Create a new content manager based on the default content manager that allows registering loaders for each file type.

There is a caveat with the second option. If jupyter_collaboration is installed with another extension that also swaps the default content manager, it may not work.

hbcarlos:
Hi @Zsailer. I'm pining for you since your thoughts here would be appreciated, and you might be interested in the comment I posted above.

I will try to join the jupyter server team meeting this week to discuss this topic. Jupyter Server's collaborators should be aware of and participate in the decisions taken here in the jupyter_collaboration extension since this involves some parts of the jupyter server.

Thanks in advance for your comments and your time!

Zsailer:
Thanks for pinging me here, @hbcarlos. I'll give this a more thorough review over the next couple of days. 

I don't know if I'll make it to tomorrow's Jupyter Server meeting—I have a family commitment that conflicts with part of the meeting. If I don't see you then, let's connect over gitter and maybe set up a time to meet separately?

hbcarlos:
Thanks, @Zsailer! 

> If I don't see you then, let's connect over gitter and maybe set up a time to meet separately?

It's fine, we can continue the discussion here and discuss also it at next week's meeting.

jupyter_collaboration/handlers.py
```diff
@@ -129,10 +86,86 @@ class YDocWebSocketHandler(WebSocketHandler, JupyterHandler):
...
+                rooms_ready=False,
+                auto_clean_rooms=False,
+                ystore_class=self._ystore_class,
+                log=self.log,
+            )
+
+        self._message_queue = asyncio.Queue()
+
+        # Get room
+        self._room_id: str = request.path.split("/")[-1]
```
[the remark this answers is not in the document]
hbcarlos: Nothing in particular. It was just to separate the initialization of the client.

> with potential ambiguities if there are / in the room id?

There should not be any ambiguities. We are using IDs for rooms instead of file paths.

jupyter_collaboration/handlers.py
```diff
@@ -129,10 +86,86 @@ class YDocWebSocketHandler(WebSocketHandler, JupyterHandler):
...
+    files: Dict[str, FileLoader] = {}
     websocket_server: Optional[JupyterWebsocketServer] = None
     _message_queue: "asyncio.Queue[Any]"
 
+    def __init__(self, app: ServerWebApplication, request: HTTPServerRequest, **kwargs):
+        super().__init__(app, request, **kwargs)
+
+        # CONFIG
+        self._file_id_manager = self.settings["file_id_manager"]
+        self._ystore_class = self.settings["collaborative_ystore_class"]
```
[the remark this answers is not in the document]
davidbrochart:
```suggestion
        _ystore_class = self.settings["collaborative_ystore_class"]
```

jupyter_collaboration/handlers.py
```diff
@@ -149,130 +186,33 @@ async def __anext__(self):
...
-            self.log.debug("Reverting file that had out-of-band changes: %s", file_path)
-            model = await ensure_async(
-                self.contents_manager.get(file_path, type=file_type, format=file_format)
-            )
-            self.room.document.source = model["content"]
-            self.last_modified = model["last_modified"]
+            # Initialize the room
+            async with self.room.lock:
+                if not self.room.ready:
+                    await self.room.initialize()
```
davidbrochart, line 215: I think the room lock should not be used outside of the room, but rather in its `initialize()` method.
Maybe also the check for room ready.

jupyter_collaboration/handlers.py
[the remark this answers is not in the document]
hbcarlos: Yep, that's right.

hbcarlos:
With the current implementation, the content is not synced between different rooms. It doesn't sync because there is only one property, `_last_modified`, for each file. If room A saves to disk, we update the property `_last_modified`. Next time we load the content, we will not notify every other room because, for the file, the content is up to date. See:
https://github.com/jupyterlab/jupyter_collaboration/blob/b1aedb92da354cf6c9df8979755145e8a3e952d7/jupyter_collaboration/loaders.py#L89

We could change this logic and keep a timestamp by room, allowing to sync between rooms. However, I prefer not to sync between rooms for now. If we sync the content between rooms, each time the content changes in one room, on every other room, we will create a Y update replacing the entire content. This could increase memory usage, so I would like to experiment before allowing rooms to sync.

davidbrochart:
> With the current implementation, the content is not synced between different rooms.

You mean with this PR? Before this PR, the content is synced between rooms.

davidbrochart:
> If we sync the content between rooms, each time the content changes in one room, on every other room, we will create a Y update replacing the entire content.

See https://github.com/jupyter-server/jupyter_ydoc/issues/15.

hbcarlos:
> You mean with this PR?

Yes, I meant in this PR.

davidbrochart:
So you mean that this PR breaks document synchronization between rooms?

jupyter_collaboration/loaders.py
```diff
@@ -0,0 +1,94 @@
...
+        return len(self._subscriptions)
+
+    def observe(self, id: str, callback: Callable[[str], Coroutine[Any, Any, None]]) -> None:
+        self._subscriptions[id] = callback
+
+    def unobserve(self, id: str) -> None:
+        del self._subscriptions[id]
+
+    async def load_content(self, format: str, file_type: str, content: bool) -> Dict[str, Any]:
+        return await ensure_async(
```
davidbrochart, line 53: Don't we want to use the lock here?

davidbrochart: I think it's problematic to load a file while it is being saved. You are probably missing a lock [here](https://github.com/jupyterlab/jupyter_collaboration/pull/121/files#diff-54b0d5450e8be93702b7ef60bc9596a423a203f5591cb9b61c6ff29dbcadf39bR113).

jupyter_collaboration/loaders.py
```diff
@@ -0,0 +1,94 @@
...
+            self._contents_manager.get(self._path, format=format, type=file_type, content=content)
+        )
+
+    async def save_content(self, model: Dict[str, Any]) -> None:
+        async with self._lock:
+            m = await self.load_content(model["format"], model["type"], False)
+
+            if self._last_modified is None or self._last_modified >= m["last_modified"]:
```
davidbrochart, line 61: I don't like this condition: `self._last_modified >= m["last_modified"]`. It is confusing because `self._last_modified` cannot be greater than `m["last_modified"]`, and the way it is written now lets people think that it can.
You can either rewrite it like this:
```py
if self._last_modified is None or self._last_modified == m["last_modified"]:
```
Or like that (switching the blocks):
```py
if self._last_modified is not None and self._last_modified < m["last_modified"]:
```

jupyter_collaboration/loaders.py
```diff
@@ -0,0 +1,94 @@
...
+    async def _maybe_load_document(self) -> None:
+        async with self._lock:
+            model = await self.load_content(self._file_format, self._file_type, False)
+
+            # do nothing if the file was saved by us
+            if self._last_modified is not None and self._last_modified < model["last_modified"]:
+                self._log.info("Notifying rooms. The file on disk changed: %s", self._path)
+                self._last_modified = model["last_modified"]
+                # Notify that the content changed on disk
+                for _, callback in self._subscriptions.items():
```
davidbrochart, line 93:
```suggestion
                for callback in self._subscriptions.values():
```

jupyter_collaboration/loaders.py
```diff
@@ -0,0 +1,94 @@
...
+                self._last_modified = model["last_modified"]
+
+            else:
+                # file changed on disk, let's revert
+                self._log.info(
+                    "Notifying rooms. Out-of-band changes while trying to save: %s", self._path
+                )
+                self._last_modified = model["last_modified"]
+                # Notify that the content changed on disk
+                for _, callback in self._subscriptions.items():
```
davidbrochart, line 73:
```suggestion
                for callback in self._subscriptions.values():
```

jupyter_collaboration/loaders.py
```diff
@@ -0,0 +1,94 @@
...
+
+    @property
+    def path(self):
+        return self._path
+
+    def clean(self) -> None:
+        self._watcher.cancel()
+
+    def rename_file(self, path: str) -> None:
+        self._path = path
```
davidbrochart, line 41: Where is this used?

jupyter_collaboration/rooms.py
```diff
@@ -0,0 +1,154 @@
...
+    def cleaner(self) -> Optional[asyncio.Task]:
+        return self._cleaner
+
+    @cleaner.setter
+    def cleaner(self, value: asyncio.Task) -> None:
+        self._cleaner = value
+
+    def _clean(self) -> None:
+        super()._clean()
+        # TODO: Should we cancel or wait ?
```
davidbrochart, line 59: I would say that we should wait, but that needs changing some methods to async in `ypy-websocket`.

jupyter_collaboration/loaders.py
hbcarlos: We are using the contents manager. I'm assuming it is thread-safe. Otherwise, we would have the same issue with the contents API.

jupyter_collaboration/rooms.py
hbcarlos: Then I think we should leave it like this for now.

This is a rare case. Every time the user stops typing we save the content with 1-second delay.

jupyter_collaboration/loaders.py
hbcarlos: Is not used

davidbrochart: How do you handle file renames then? Shouldn't it be connected to the file ID service, which knows about files that are renamed?

hbcarlos: Good catch! I was not handling the renames.
I just pushed a commit to use file ids instead of paths and do the transformation when saving. This fixes the rename.

jupyter_collaboration/handlers.py
```diff
@@ -94,18 +53,17 @@ async def monitor(self):
             self.log.info("Connected Y users: %s", clients_nb)
             self.ypatch_nb = 0
 
+    def room_exists(self, path: str) -> bool:
+        return path in self.rooms.keys()
```
fcollonval, line 57: Unneeded keys():

```suggestion
        return path in self.rooms
```

jupyter_collaboration/handlers.py
```diff
@@ -129,10 +87,90 @@ class YDocWebSocketHandler(WebSocketHandler, JupyterHandler):
...
+        self, app: ServerWebApplication, request: HTTPServerRequest, **kwargs: Dict[str, Any]
+    ):
+        super().__init__(app, request, **kwargs)
+
+        # CONFIG
+        file_id_manager = self.settings["file_id_manager"]
+        ystore_class = self.settings["collaborative_ystore_class"]
+        self._cleanup_delay = self.settings["collaborative_document_cleanup_delay"]
+        # self.settings["collaborative_file_poll_interval"]
+        # self.settings["collaborative_document_save_delay"]
```
fcollonval, line 104: Dead code

```suggestion
```

jupyter_collaboration/handlers.py
```diff
@@ -321,72 +261,51 @@ def on_close(self) -> None:
...
-                self.last_modified = model["last_modified"]
-        self.room.document.dirty = False
+        # Clean room
+        del self.room
+        self.log.info("Room %s deleted", self._room_id)
+
+        # Clean the file loader if there are not rooms using it
+        _, _, file_id = decode_file_path(self._room_id)
+        file = self.files[file_id]
+        if file.number_of_subscriptions() == 0:
```
fcollonval, line 287: As you use a name, it makes more sense to define that as a `property` rather than a `method`.

jupyter_collaboration/handlers.py
```diff
@@ -321,72 +261,51 @@ def on_close(self) -> None:
...
+            self.log.info("Deleting file %s", file.path)
+            file.clean()
+            del self.files[file_id]
 
     def check_origin(self, origin):
         return True
 
+    @classmethod
+    def clean_up(cls):
```
fcollonval, line 296: This is another proof that the architecture is not yet great - but let put it aside for this PR.

jupyter_collaboration/loaders.py
```diff
@@ -0,0 +1,108 @@
...
+class FileLoader:
+    def __init__(
+        self,
+        file_id: str,
+        file_format: str,
+        file_type: str,
+        file_id_manager: Any,
+        contents_manager: Any,
+        log: Logger | None,
+        poll_interval: float | None = None,
```
fcollonval, line 19: Seems that this was introduced only in 3.10, could you keep using `Optional` for now:

```suggestion
        log: Optional[Logger],
        poll_interval: Optional[float]= None,
```

jupyter_collaboration/loaders.py
```diff
@@ -0,0 +1,108 @@
...
+        self._lock = asyncio.Lock()
+        self._poll_interval = poll_interval
+        self._file_id_manager = file_id_manager
+        self._contents_manager = contents_manager
+
+        self._log = log or getLogger(__name__)
+        self._subscriptions: dict[str, Callable[[str], Coroutine[Any, Any, None]]] = {}
+
+        if self._poll_interval:
+            self._watcher = asyncio.create_task(self.watch_file())
```
fcollonval, line 35: Let's be sure the attribute exits

```suggestion
        self._watcher = asyncio.create_task(self.watch_file()) if self._poll_interval else None
```

jupyter_collaboration/loaders.py
```diff
@@ -0,0 +1,108 @@
...
+                # file changed on disk, let's revert
+                self._log.info(
+                    "Notifying rooms. Out-of-band changes while trying to save: %s", path
+                )
+                self._last_modified = model["last_modified"]
+                # Notify that the content changed on disk
+                for callback in self._subscriptions.values():
+                    await callback("changed")
+
+    async def watch_file(self) -> None:
```
fcollonval, line 85: This should be private and therefore we can drop the assertion.

jupyter_collaboration/rooms.py
```diff
@@ -0,0 +1,156 @@
...
+
+    def __init__(
+        self,
+        room_id: str,
+        file_format: str,
+        file_type: str,
+        file: FileLoader,
+        ystore: BaseYStore | None,
+        log: Logger | None,
+        save_delay: int | None = None,
```
fcollonval, line 27: Same comment

```suggestion
        ystore: Optional[BaseYStore],
        log: Optional[Logger],
        save_delay: Optional[int] = None,
```

jupyter_collaboration/rooms.py
```diff
@@ -0,0 +1,156 @@
...
+        await self._file.save_content(
+            {"format": self._file_format, "type": self._file_type, "content": self._document.source}
+        )
+        self._document.dirty = False
+
+
+class TransientRoom(YRoom):
+    """A Y room for sharing state (e.g. awareness)."""
+
+    def __init__(self, room_id: str, log: Logger | None):
```
fcollonval, line 149:
```suggestion
    def __init__(self, room_id: str, log: Optional[Logger]):
```

jupyter_collaboration/stores.py
```diff
@@ -0,0 +1,32 @@
...
+from traitlets.config import LoggingConfigurable
+from ypy_websocket.ystore import SQLiteYStore as _SQLiteYStore
+from ypy_websocket.ystore import TempFileYStore as _TempFileYStore
+
+
+class TempFileYStore(_TempFileYStore):
+    prefix_dir = "jupyter_ystore_"
+
+
+class SQLiteYStoreMetaclass(type(LoggingConfigurable), type(_SQLiteYStore)):  # type: ignore
```
fcollonval, line 14: Could you add a comment why we need this complex code based on metaclass rather than the classical direct inheritance?

fcollonval (changes requested):
Thanks @hbcarlos 

Here are some suggestions. It will be good to add doc string to document the code and unit tests to at least test for the correctness of the instantiation and clean actions.

jupyter_collaboration/loaders.py
davidbrochart: You can use it with older Python versions if you add `from __future__ import annotations`.

jupyter_collaboration/handlers.py
hbcarlos: No, I want to keep it there as comments. It helps to rapidly see what settings are available.

davidbrochart:
> So you mean that this PR breaks document synchronization between rooms?

I've not been following all the commits, has this been addressed?

hbcarlos:
> I've not been following all the commits, has this been addressed?

I'm not restoring it. This PR removes the synchronization between rooms in favor of simplicity and stability. I'll add it back once the extension is stable and we have a pleasant RTC experience.

davidbrochart:
I think it is a big regression.
So what is the user experience when a notebook is opened as a notebook document and as a JSON document, for instance?
What happens to one document when the user makes a change to the other?

jupyter_collaboration/handlers.py
hbcarlos: I know I'm doing my best, but there are already a lot of changes in this PR.

hbcarlos: Feel free to open an issue, and I'll work on it in another PR

jupyter_collaboration/loaders.py
hbcarlos: Good catch! Thanks!

jupyter_collaboration/stores.py
hbcarlos: I just moved it to a new file. I'm not sure why we need it. I have not looked into the stores yet.

@davidbrochart Could you open a PR adding documentation?

hbcarlos:
Users are notified that they might lose changes if they open the same document with multiple views.

If you edit in one room, it won't sync with the other one. The last edited room is the one saved to disk.

davidbrochart:
I think it needs to be discussed, as it is a significant regression and it could lead to a lot of user frustration.
What about changes that are made on the backend, like checking out a git branch that would change the open file? Is the change reflected on the frontend, or is that also not supported anymore?

hbcarlos:
> What about changes that are made on the backend, like checking out a git branch that would change the open file? Is the change reflected on the frontend, or is that also not supported anymore?

That is supported

davidbrochart:
So there is something I don't understand. If the frontend reacts to file changes made in the backend, and you have one user with a notebook opened as a notebook document, and another user with the same notebook opened as a JSON document, what is the behavior?
Users won't be warned because they individually only have one view of the notebook, right?

hbcarlos:
Yes, because I prioritize what is in the disk instead of automatically overwriting it whenever a change occurs.

I'm not in favor of the synchronization between rooms or the auto-save. It is unpredictable, and I'm sure it is related to some of the issues we are having.

davidbrochart:
It is unclear to me what you do in this PR regarding room synchronization. Commit https://github.com/jupyterlab/jupyter_collaboration/pull/121/commits/1ee384833a5e7587d08f3fdd2dfc4af0f558bd0d says "Fixes sync between rooms", but it doesn't seem to work. Can you clarify what is fixed?

fcollonval (approved):
Thanks @hbcarlos 

Let's move forward with this one.

davidbrochart:
@fcollonval I was not done testing this PR, and there was still pending [questions](https://github.com/jupyterlab/jupyter_collaboration/pull/121#issuecomment-1502894043).

davidbrochart:
> The synchronization has been fixed.

It's hard to tell, I think we should give more time for reviewers to get their questions answered before merging.

PR #122, opened 2023-03-16
Display remote user cursors

fcollonval:
This add an CodeMirror extension to display remote user cursors.

image

This uses a view plugin to communicate the user cursors through the awareness. Then the UI elements are displayed in 3 companion extensions:
- A remote cursors layer
- A remote selection layer
- Hover on tooltip to display user info


hbcarlos (approved):
Thanks @fcollonval!

PR #133, opened 2023-03-28
Allow other extensions to register shared models

hbcarlos:
Fixes JupyterCAD and Glue-lab

cc @trungleduc @martinRenou

### DIFF
--- a/packages/collaboration-extension/src/filebrowser.ts
+++ b/packages/collaboration-extension/src/filebrowser.ts
@@ -14,7 +14,16 @@ import { ISettingRegistry } from '@jupyterlab/settingregistry';
 
 import { CommandRegistry } from '@lumino/commands';
 
-import { YDrive } from '@jupyter/docprovider';
+import {
+  YFile,
+  YNotebook
+} from '@jupyter/ydoc';
+
+import { 
+  ICollaborativeDrive,
+  SharedDocumentFactory,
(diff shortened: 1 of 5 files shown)


trungleduc:
Why do we accept only the `string` file format here? https://github.com/jupyterlab/jupyter_collaboration/blob/2432fe032cce3414836ef035fe975acb1fcd8655/packages/docprovider/src/ydrive.ts#L201-L203

hbcarlos:
Because the [file format can be null](https://github.com/jupyterlab/jupyterlab/blob/ed0e0eb9251546cce0acef580c7b6349ee9bbad8/packages/services/src/contents/index.ts#L134) and we need it to create the room id in the document provider. See https://github.com/jupyterlab/jupyter_collaboration/blob/2432fe032cce3414836ef035fe975acb1fcd8655/packages/docprovider/src/ydrive.ts#L124

trungleduc:
`JupyterCad` defines the `FCStd` file format as `base64`, it will work here?

hbcarlos:
> JupyterCad defines the FCStd file format as base64, it will work here?

Yes. We are checking the attribute type (`typeof options.format`), not the attribute itself.

trungleduc (approved):

davidbrochart (approved):

PR #136, opened 2023-03-30
Follow up #133

hbcarlos:
> * Could you raise an error in registerDocumentFactory if there is already an factory for the content type?
> * Could you correct the doc string for the plugins? It will be also good to define the new description attribute of the application plugin

_Review by @fcollonval_


fcollonval (approved):
Thanks @hbcarlos

PR #138, opened 2023-04-11
Fix metadata

hbcarlos:
Updates jupyter_ydoc to fix an issue with the YNotebook. See https://github.com/jupyter-server/jupyter_ydoc/pull/158

Follow up #121 :
* Use a lock to ignore self-updates
* Fixes a formatting error

### DIFF
--- a/jupyter_collaboration/loaders.py
+++ b/jupyter_collaboration/loaders.py
@@ -149,7 +149,7 @@ class FileLoader:
                 await self._maybe_load_document()
 
             except Exception as e:
-                self._log.error("Error watching file: %s\n", self.path, e)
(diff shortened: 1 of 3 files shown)


davidbrochart:
@hbcarlos I would appreciate if you give me a bit of time to be able to review, instead of merging your own PR.

fcollonval:
@davidbrochart @hbcarlos what do you think of adding a rule requiring an approved review to allow merge (I guess we should also do the same on jupyter_ydoc)?

davidbrochart:
Yes I think it would be useful, here and in jupyter-ydoc.

PR #139, opened 2023-04-21
Emit events for collaborative sessions

hbcarlos:
Emits events during a collaborative session.

Adds a new lab plugin to log the events in the dev console of the client. This might be useful to debug issues from users that don't have access to the terminal where the server is running, for example, users of JupyterHub-based deployments.

Adds a new dialog to alert users that multiple rooms are accessing the same file.


jupyter_collaboration/handlers.py
```diff
@@ -173,13 +173,20 @@ def __init__(
...
                         self.contents_manager,
                         self.log,
                         self.settings["collaborative_file_poll_interval"],
                     )
                     self.files[file_id] = file
 
+                else:
+                    self._emit(
+                        "initialize",
+                        "There is another collaborative session accessing the same file.\nThe synchronization between rooms is not supported and you might lose some of your progress.",
```
davidbrochart, line 186: From [here](https://github.com/jupyterlab/jupyter_collaboration/pull/121#issuecomment-1503510439), I thought synchronization between rooms was supported. Could you clarify?

jupyter_collaboration/events/session.yaml
```diff
@@ -0,0 +1,50 @@
...
+      - initialize
+      - load
+      - save
+      - overwrite
+      - clean
+    description: |
+      Action performed in a room during a collaborative session.
+      Possible values:
+      1. initialize
+         Initialize a room by loading the content from disk or a store.
```
fcollonval, line 32:
```suggestion
         Initialize a room by loading the content from the contents manager or a store.
```

jupyter_collaboration/events/session.yaml
```diff
@@ -0,0 +1,50 @@
...
+      2. load
+         Load the content from disk.
```
fcollonval, line 34:
```suggestion
         Load the content from the contents manager.
```

jupyter_collaboration/events/session.yaml
```diff
@@ -0,0 +1,50 @@
...
+      3. save
+         Save the content to disk.
```
fcollonval, line 36:
```suggestion
         Save the content with the contents manager.
```

jupyter_collaboration/events/session.yaml
```diff
@@ -0,0 +1,50 @@
...
+      4. overwrite
+         Overwrite the content in a room with content from disk.
```
fcollonval, line 38:
```suggestion
         Overwrite the content in a room with content from the contents manager.
```

jupyter_collaboration/events/session.yaml
```diff
@@ -0,0 +1,50 @@
...
+         This could happen when multiple rooms access the same file or
```
fcollonval, line 39:
```suggestion
         This can happen when multiple rooms access the same file or
```

jupyter_collaboration/events/session.yaml
```diff
@@ -0,0 +1,50 @@
...
+         when a user modifies the file with a different app.
```
fcollonval, line 40:
```suggestion
         when an user modifies the file outside Jupyter Server (e.g. using a different app).
```

jupyter_collaboration/events/session.yaml
```diff
@@ -0,0 +1,50 @@
...
+      5. clean
+         Clean the room once is empty.
```
fcollonval, line 42:
```suggestion
         Clean the room once is empty (aka there is no more users connected to it).
```

jupyter_collaboration/events/session.yaml
```diff
@@ -0,0 +1,50 @@
...
+title: Events during collaborative sessions
+personal-data: true
+description: |
+  Record the event that happen in the backend during a collaborative session.
+type: object
+required:
+  - room
+  - path
+  - action
+properties:
```
fcollonval, line 12: Should we add an optional property to provide the store used by the file?

jupyter_collaboration/events/session.yaml
```diff
@@ -0,0 +1,50 @@
...
+  room:
+    type: string
+    description: |
+      Room Id. Usually composed by the file type, format and Id.
+  path:
+    type: string
+    description: |
+      File path.
```
fcollonval, line 20: Should we provide the file ID?

jupyter_collaboration/handlers.py
fcollonval: I guess we should rather say that it is _experimental_.

fcollonval, line 186:
```suggestion
                        "There is another collaborative session accessing the same file.\nThe synchronization between rooms is not supported and you might lose some of your changes.",
```

jupyter_collaboration/handlers.py
```diff
@@ -321,6 +331,14 @@ def on_close(self) -> None:
             self.log.info("Cleaning room: %s", self._room_id)
             self.room.cleaner = asyncio.create_task(self._clean_room())
 
+    def _emit(self, action: str, msg: str, warn: bool = False) -> None:
```
fcollonval, line 334: Here we should switch to an enum:

```suggestion
    def _emit(self, action: str, msg: str, level: Enum = INFO) -> None:
```

packages/collaboration-extension/src/filebrowser.ts
```diff
@@ -160,6 +161,42 @@ export const defaultFileBrowser: JupyterFrontEndPlugin<IDefaultFileBrowser> = {
...
 
+/**
+ * The default collaborative drive provider.
+ */
+export const logger: JupyterFrontEndPlugin<void> = {
+  id: '@jupyter/collaboration-extension:logger',
+  description: 'A logging plugin for debugging purposes.',
+  autoStart: true,
+  requires: [ITranslator],
+  optional: [],
```
fcollonval, line 172: You should try to request as few as possible token.

```suggestion
  optional: [ITranslator],
```

packages/collaboration-extension/src/filebrowser.ts
```diff
@@ -160,6 +161,42 @@ export const defaultFileBrowser: JupyterFrontEndPlugin<IDefaultFileBrowser> = {
...
+  activate: (app: JupyterFrontEnd, translator: ITranslator): void => {
+    const trans = translator.load('jupyter_collaboration');
```
fcollonval, line 174:
```suggestion
  activate: (app: JupyterFrontEnd, translator: ITranslator | null): void => {
    const trans = (translator ?? nullTranslator).load('jupyter_collaboration');
```

packages/collaboration-extension/src/filebrowser.ts
```diff
@@ -160,6 +161,42 @@ export const defaultFileBrowser: JupyterFrontEndPlugin<IDefaultFileBrowser> = {
...
+    app.serviceManager.events.stream.connect((_, emission) => {
+      if (
+        emission.schema_id ===
+        'https://events.jupyter.org/jupyter_server/jupyter_collaboration/v1'
+      ) {
+        console.debug(
+          `[${emission.room}(${emission.path})] ${emission.action}: ${emission.msg}`
+        );
+
+        if (emission.warn === true) {
```
fcollonval, line 185: This will need update when switching to an enumeration.

packages/collaboration-extension/src/filebrowser.ts
```diff
@@ -160,6 +161,42 @@ export const defaultFileBrowser: JupyterFrontEndPlugin<IDefaultFileBrowser> = {
...
+  optional: [],
+  activate: (app: JupyterFrontEnd, translator: ITranslator): void => {
+    const trans = translator.load('jupyter_collaboration');
+
+    app.serviceManager.events.stream.connect((_, emission) => {
+      if (
+        emission.schema_id ===
+        'https://events.jupyter.org/jupyter_server/jupyter_collaboration/v1'
+      ) {
+        console.debug(
```
fcollonval, line 181: Would you mind switching to the logger that is a perfect fit for this case: see https://github.com/jupyterlab/extension-examples/tree/4.0/log-messages

fcollonval (changes requested):
Thanks a lot @hbcarlos this is a great addition.

I have minor wording comments and two suggestions:
- Use an enum for the level of message
- Use the JupyterLab logger instead of the browser console

jupyter_collaboration/handlers.py
hbcarlos: The UX of synchronization between rooms is pretty bad. With a good connection, a noticeable delay makes it unpredictable.

When two users are editing the same file using different models (for example, a user editing a notebook using the notebook view and another one using the plain text editor), propagating the changes from user A in room A to user B in room B takes a few seconds (there is a 1s delay to save the document + 1s delay for polling the file assuming that the communication and load delay is 0). During those 2s, user B can write a sentence that will disappear without any notification or asking the user if they want to overwrite the content.

In addition, we use the default content manager to load and save documents. When loading or saving a notebook, the content manager throws an error if the notebook is invalid (so it doesn't load/save the content). If user A in room A is editing the notebook using the plain text view, room B (with user B using the notebook editor) will be updated only when room A saves a valid notebook. Suppose user A forgets a curly bracket (which often happens because there is no visual indicator). In that case, everything user B is typing will be deleted once user A fixes the notebook without allowing user B to overwrite what is on the disk. Furthermore, if user B doesn't realize that their changes are not being saved, they can be typing for hours, and everything will be deleted.

The combination of auto-saving and room synchronization can be catastrophic; in my opinion, we can not call that "supported".

jupyter_collaboration/events/session.yaml
hbcarlos: I don't think it is necessary, the room ID includes the file ID. In addition, if we show messages to the user, they don't know what file it is.

jupyter_collaboration/events/session.yaml
```diff
@@ -0,0 +1,59 @@
+"$id": https://events.jupyter.org/jupyter_server/jupyter_collaboration/v1
```
afshin, line 1:
```suggestion
"$id": https://schema.jupyter.org/jupyter_collaboration/session/v1
```

jupyter_collaboration/events/session.yaml
```diff
@@ -0,0 +1,59 @@
...
+version: 1
+title: Collaborative session events
+personal-data: true
+description: |
+  Record the events that happen in the backend during a collaborative session.
```
afshin, line 6:
```suggestion
  Events emitted server-side during a collaborative session.
```

jupyter_collaboration/events/session.yaml
```diff
@@ -0,0 +1,59 @@
...
+      Possible values:
+      1. initialize
+         Initialize a room by loading the content from the contents manager or a store.
+      2. load
+         Load the content from the contents manager.
+      3. save
+         Save the content with the contents manager.
+      4. overwrite
+         Overwrite the content in a room with content from the contents manager.
+         This can happen when multiple rooms access the same file or when an user
```
afshin, line 52:
```suggestion
         This can happen when multiple rooms access the same file or when a user
```

package.json
```diff
@@ -50,6 +50,124 @@
...
+    "@lumino/messaging": "^2.0.0",
+    "@lumino/polling": "^2.0.0",
+    "@lumino/properties": "^2.0.0",
+    "@lumino/signaling": "^2.0.0",
+    "@lumino/virtualdom": "^2.0.0",
+    "@lumino/widgets": "^2.0.1",
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0",
+    "yjs": "^13.5.40"
+  },
```
fcollonval, line 170: Let's remove the resolutiosns as discussed

```suggestion
```

packages/collaboration-extension/src/filebrowser.ts
```diff
@@ -160,6 +163,92 @@ export const defaultFileBrowser: JupyterFrontEndPlugin<IDefaultFileBrowser> = {
...
+              emission.msg ?? ''
+            }`
+          });
+
+          if (emission.level === 'WARNING') {
+            showDialog({
+              title: trans.__('Warning'),
+              body: trans.__(
+                `Two collaborative sessions are accessing the file ${emission.path} simultaneously.
+                \nOpening the same file using different views simultaneously is not supported. Please, close one view; otherwise, you might lose some of your progress.`
```
fcollonval, line 241: Internationalized parametrized string syntax is a bit different:

```suggestion
                'Two collaborative sessions are accessing the file %1 simultaneously.
                \nOpening the same file using different views simultaneously is not supported. Please, close one view; otherwise, you might lose some of your progress.',
                emission.path
```

packages/collaboration-extension/src/filebrowser.ts
```diff
@@ -160,6 +163,92 @@ export const defaultFileBrowser: JupyterFrontEndPlugin<IDefaultFileBrowser> = {
...
+              ),
+              buttons: [Dialog.okButton()]
```
fcollonval, line 243: What about using the _red_ warning button?

```suggestion
              buttons: [Dialog.warnButton({ label: trans.__('Ok') })]
```

fcollonval (changes requested):
Thanks @hbcarlos 

Some final minor suggestions from me and it should be ready.

fcollonval (approved):
Thanks @hbcarlos

PR #144, opened 2023-05-02
Add JavaScript and lab extension test in CI

fcollonval:
## Code changes

Add test for installing the extension and the frontend proving current code is broken


fcollonval:
Merging as is as it does not change any code but add fail safe for follow-up bug hunting work.

PR #145, opened 2023-05-02
Add logger for file editor

fcollonval:
This requires a new release of JupyterLab with https://github.com/jupyterlab/jupyterlab/pull/14477

I checked locally that linking to that dev version of JupyterLab fixes the error.


hbcarlos (approved):
Thanks!

PR #146, opened 2023-05-06
Improve code architecture

fcollonval:
This improves the code architecture by:
- Instantiating the websocket server independently
- Instantiating the file loader mapping independently
- Initialize clean-up logic (some will need upstream work)

### DIFF
--- a/jupyter_collaboration/app.py
+++ b/jupyter_collaboration/app.py
@@ -104,4 +104,3 @@ class YDocExtension(ExtensionApp):
     async def stop_extension(self):
         # Cancel tasks and clean up
         await asyncio.wait({self.ywebsocket_server.clean(), self.file_loaders.clear()}, timeout=3)
-        # self.log.debug("All pending tasks:\n  %s", "\n  ".join(map(lambda t: repr(t), asyncio.all_tasks())))
--- a/jupyter_collaboration/app.py
+++ b/jupyter_collaboration/app.py
@@ -2,9 +2,10 @@
 # Distributed under the terms of the Modified BSD License.
 from __future__ import annotations
 
+import asyncio
 
 from jupyter_server.extension.application import ExtensionApp
-from traitlets import Float, Int, Type
+from traitlets import Float, Type
 from ypy_websocket.ystore import BaseYStore
 
 from .handlers import DocSessionHandler, YDocWebSocketHandler
@@ -50,10 +51,6 @@ class YDocExtension(ExtensionApp):
         directory.""",
     )
 
-    # @property
-    # def ywebsocket_server(self) -> JupyterWebsocketServer | None:
-    #     return self.settings.get("ywebsocket_server")
-
     def initialize(self):
         super().initialize()
         self.serverapp.event_logger.register_event_schema(EVENTS_SCHEMA_PATH)
@@ -81,9 +78,9 @@ class YDocExtension(ExtensionApp):
         )
 
         # self.settings is local to the ExtensionApp but here we need
-        # the global app settings in which the file id manager will later
-        # register itself.
-        file_loaders = FileLoaderMapping(
+        # the global app settings in which the file id manager will register
+        # itself maybe at a later time.
+        self.file_loaders = FileLoaderMapping(
             self.serverapp.web_app.settings, self.log, self.file_poll_interval
         )
 
@@ -95,7 +92,7 @@ class YDocExtension(ExtensionApp):
                     {
                         "document_cleanup_delay": self.document_cleanup_delay,
                         "document_save_delay": self.document_save_delay,
-                        "file_loaders": file_loaders,
+                        "file_loaders": self.file_loaders,
                         "ystore_class": self.ystore_class,
                         "ywebsocket_server": self.ywebsocket_server,
                     },
@@ -106,5 +103,5 @@ class YDocExtension(ExtensionApp):
 
     async def stop_extension(self):
         # Cancel tasks and clean up
-        # if hasattr(self, "ywebsocket_server"):
-        await self.ywebsocket_server.clean()
+        await asyncio.wait({self.ywebsocket_server.clean(), self.file_loaders.clear()}, timeout=3)
+        # self.log.debug("All pending tasks:\n  %s", "\n  ".join(map(lambda t: repr(t), asyncio.all_tasks())))
--- a/jupyter_collaboration/app.py
+++ b/jupyter_collaboration/app.py
@@ -103,4 +103,10 @@ class YDocExtension(ExtensionApp):
 
     async def stop_extension(self):
         # Cancel tasks and clean up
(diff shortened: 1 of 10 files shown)


jupyter_collaboration/loaders.py
```diff
@@ -144,25 +160,97 @@ async def _watch_file(self) -> None:
...
+    @property
+    def contents_manager(self) -> AsyncContentsManager | ContentsManager:
+        return self._settings["contents_manager"]
+
+    @property
+    def file_id_manager(self) -> BaseFileIdManager:
+        return self._settings["file_id_manager"]
+
+    def __contains__(self, file_id: str) -> bool:
+        """Test if a file as a loader."""
```
hbcarlos, line 215:
```suggestion
        """Test if a file has a loader."""
```

jupyter_collaboration/websocketserver.py
```diff
@@ -0,0 +1,155 @@
...
+        ystore_class: BaseYStore,
+        rooms_ready: bool = True,
+        auto_clean_rooms: bool = True,
+        log=None,
+    ):
+        super().__init__(rooms_ready, auto_clean_rooms, log)
+        self.ystore_class = ystore_class
+        self.ypatch_nb = 0
+        self.connected_users = {}
+        # Async loop is not yet ready at the object instatiation
```
hbcarlos, line 38:
```suggestion
        # Async loop is not yet ready at the object instantiation
```

hbcarlos (approved):
Thank you so much, @fcollonval!

hbcarlos:
@davidbrochart, could you take a look at this PR? I would appreciate your experience here.

jupyter_collaboration/app.py
```diff
@@ -61,12 +66,42 @@ def initialize_settings(self):
...
+                ),
                 (r"/api/collaboration/session/(.*)", DocSessionHandler),
             ]
         )
 
     async def stop_extension(self):
-        YDocWebSocketHandler.clean_up()
+        # Cancel tasks and clean up
+        await asyncio.wait({self.ywebsocket_server.clean(), self.file_loaders.clear()}, timeout=3)
+        # self.log.debug("All pending tasks:\n  %s", "\n  ".join(map(lambda t: repr(t), asyncio.all_tasks())))
```
davidbrochart, line 107:
```suggestion
```

jupyter_collaboration/handlers.py
```diff
@@ -123,72 +48,43 @@ class YDocWebSocketHandler(WebSocketHandler, JupyterHandler):
...
 
-    def __init__(
-        self, app: ServerWebApplication, request: HTTPServerRequest, **kwargs: dict[str, Any]
+    def initialize(
+        self,
+        ywebsocket_server: JupyterWebsocketServer,
+        file_loaders: FileLoaderMapping,
+        ystore_class: Type[BaseYStore],
+        document_cleanup_delay: float | None = 60.0,
+        document_save_delay: float | None = 1.0,
```
davidbrochart, line 59: This looks like a code duplication from the traits, is it possible to use their values?

jupyter_collaboration/loaders.py
```diff
@@ -42,12 +45,20 @@ def __init__(
...
+
     @property
     def path(self) -> str:
         """
         The file path.
         """
-        return self._file_id_manager.get_path(self._file_id)
+        path = self._file_id_manager.get_path(self.file_id)
+        if path is None:
+            raise RuntimeError(f"No path found for file ID '{self.file_id}")
```
davidbrochart, line 60:
```suggestion
            raise RuntimeError(f"No path found for file ID '{self.file_id}'")
```

jupyter_collaboration/websocketserver.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+class RoomNotFound(LookupError):
+    pass
+
+
+class JupyterWebsocketServer(WebsocketServer):
+    """YPy websocket server.
+
+    It communicates the document updates to all clients for each rooms.
```
davidbrochart, line 21:
```suggestion
    """Ypy websocket server.

    It communicates the document updates to all clients for each room.
```

jupyter_collaboration/app.py
```diff
@@ -61,12 +66,42 @@ def initialize_settings(self):
...
+                    },
+                ),
                 (r"/api/collaboration/session/(.*)", DocSessionHandler),
             ]
         )
 
     async def stop_extension(self):
-        YDocWebSocketHandler.clean_up()
+        # Cancel tasks and clean up
+        await asyncio.wait({self.ywebsocket_server.clean(), self.file_loaders.clear()}, timeout=3)
```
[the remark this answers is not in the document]
fcollonval: Cancelling is not the issue here. The problem is that cancelled tasks should still be awaited for completion (this is the job of those cleaning functions).
Unfortunately the collaboration extension is broken regarding proper clean up (I still haven't found the root cause). So I'm trying to be protective by returning here to let others extensions a chance to stop properly even if this one doesn't.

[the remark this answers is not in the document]
davidbrochart: But then I don't understand why you need a timeout? Couldn't you use `asyncio.gather`?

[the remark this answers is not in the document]
hbcarlos: We need the timeout to cancel every uncompleted task after 3s and let the server continue with the shutdown. I believe `asyncio.gather` will wait until every task is completed and prevent the server from shutdown if one of the tasks is running forever.

jupyter_collaboration/handlers.py
hbcarlos: Do you mean the default values for `document_cleanup_delay` and `document_save_delay`?

We are using the default values from the traits [here](https://github.com/jupyterlab/jupyter_collaboration/blob/298bf4466bdb6da42138847abae452351e419636/jupyter_collaboration/app.py#L92-L97).

I think it is OK to also have defaults in the header of the initialize method, just in case someone uses that class somewhere else.

jupyter_collaboration/app.py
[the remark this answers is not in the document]
hbcarlos: Ooh, sorry. You are right. The `asyncio.wait` doesn't cancel the tasks after the timeout.

> But what happens if tasks have not completed after the timeout? It seems that they are left running.

I read too fast. I think his idea was to neither cancel nor wait for the tasks to complete. Just wait for 3s, and if a task did not finish, return to let the server continue the shutdown but let the task run as well.

If we use `gather`, we need to wait for every task to complete. If we use `wait_for`, we need to cancel the tasks that are not completed.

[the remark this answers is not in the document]
hbcarlos: I will merge this PR because Fred won't return for a while. We can discuss this in another PR or issue.

PR #148, opened 2023-05-09
Wait for the document to be synced

fcollonval:
Fixes https://github.com/jupyterlab/jupyterlab/issues/14404

Wait for file content to be synced before triggering websocket connection ready promise


hbcarlos (approved):
Thanks, @fcollonval !

PR #149, opened 2023-05-10
Fix injecting user menu bar in top bar.

fcollonval:
The user menu bar was having mixed configuration giving the impression it is injected in the top bar through the toolbar registry when it was not.

This fixes it.


hbcarlos (approved):
Thanks @fcollonval!

PR #150, opened 2023-05-10
Add share link feature

fcollonval:
The need of installing another extension to generate a shared link fade away now that you need to explicitly install this extension to enable live collaboration.

This add the share link feature through a button next to the user menu bar (top right).

It also simplifies the code by dropping the backend code (compared to jupyterlab-share-link) as there is no reason to provide link to another server than the current one.


jtpio:
> It also simplifies the code by dropping the backend code (compared to jupyterlab-share-link) as there is no reason to provide link to another server than the current one.

To give more context about this, the original motivate in `jupyterlab-link-share` was to provide the correct token when running on Binder.

Would be happy to archive https://github.com/jupyterlab-contrib/jupyterlab-link-share if it is not necessary anymore.

fcollonval:
Thanks @jtpio for highlighting that peculiar point.
It would be great if indeed this could work on Binder - haven't tested yet.

Another point I'm missing is the support for notebook V7. But I guess this could be tackled in a specific issue as there is also a question about what to do with the side panel.

hbcarlos:
> Another point I'm missing is the support for notebook V7. But I guess this could be tackled in a specific issue as there is also a question about what to do with the side panel.

For Notebook 7 support, we could use the [`appName`](https://github.com/jupyterlab/jupyterlab/blob/5e1318bcc2d2f98d562737712203a88e779ad0ac/packages/application/src/lab.ts#L95) or something similar to detect the distribution and reposition the different widgets. It would be great to standardize it (or maybe a different approach) so every extension can rely on that to detect the distribution.

hbcarlos (approved):
Thanks!

PR #152, opened 2023-05-21
Fix binder env

fcollonval:


binder/environment.yml
```diff
@@ -1,14 +1,11 @@
 name: example-environment
 channels:
   - conda-forge
-  - conda-forge/label/jupyterlab_rc
   - nodefaults
 dependencies:
   - jupyterlab >=4.0.0
-  - jupyterlab-link-share >=0.2
   - nodejs >=18,<19
   - python >=3.11,<3.12
```
hbcarlos, line 8: Should we ping yarn v3 instead of removing it? We still use yarn in the `postBuild` script

```suggestion
  - python >=3.11,<3.12,
  - yarn >=3,<4
```

hbcarlos (approved):
Thanks!

PR #155, opened 2023-05-29
Chat panel

brichet:
This PR adds the frontend part of a chat panel.
It has to wait until the backend is ready for use.

image


fcollonval:
This looks great @brichet.

I'm wondering if we should not merge the user panel and the chat using an accordion panel to centralize collaboration feature. For example if a user tell the others to jump to his position, in the current layout, the others will need to open the left sidebar to use the feature to jump to a given user position.

> This could also be a setting to leave the layout at the user discretion.

jtpio:
For reference Jupyter AI also has a chat panel: https://jupyter-ai.readthedocs.io/en/latest/users/index.html#the-chat-interface

image


Maybe there could be a way to share some components between the two extensions at some point.

brichet:
> I'm wondering if we should not merge the user panel and the chat using an accordion panel to centralize collaboration feature.

Yes, it can be interesting, also to not overload the page. But it would be interesting to be able to move it to another panel on the fly. Using several collaboration tools at the same time, it can be annoying to switch the accordion every time.

brichet:
> For reference Jupyter AI also has a chat panel: https://jupyter-ai.readthedocs.io/en/latest/users/index.html#the-chat-interface

Thanks for the link, I didn't know about that.
Indeed we should try to share the chat panel component.

Zsailer:
Closing in favor of #237

PR #159, opened 2023-06-05
Fix package name in README

fcollonval:

PR #166, opened 2023-06-15
Fix install instructions to use correct name

fperez:
The actual package name (on both PyPI and conda-forge) is `jupyter-collaboration` with a `-`, not a `_`.

### DIFF
--- a/docs/source/index.md
+++ b/docs/source/index.md
@@ -16,13 +16,13 @@ Editors are not collaborative by default; to activate it, install the extension
 Installation using mamba/conda:
 
 ```sh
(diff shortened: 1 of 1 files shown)


fperez:
#159 fixed the README but missed the docs. I agree with @martinRenou that it might be worth renaming the repo as well to stay consistent.

davidbrochart (approved):
Thanks @fperez !

davidbrochart:
I also renamed the repo.

fperez:
Great, thx for the quick turnaround!

PR #167, opened 2023-06-19
Chat subprotocol

hbcarlos:
Adds a new message type for broadcasting messages and a new API for the global awareness provider to support chat.

### DIFF
--- a/packages/collaboration/package.json
+++ b/packages/collaboration/package.json
@@ -41,6 +41,7 @@
   "dependencies": {
     "@codemirror/state": "^6.2.0",
     "@codemirror/view": "^6.7.0",
+    "@jupyter/docprovider": "1.0.0",
     "@jupyterlab/apputils": "^4.0.0",
     "@jupyterlab/coreutils": "^6.0.0",
     "@jupyterlab/services": "^7.0.0",
--- a/packages/docprovider/src/awareness.ts
+++ b/packages/docprovider/src/awareness.ts
@@ -5,7 +5,7 @@
 
 import { User } from '@jupyterlab/services';
 
-import { IDisposable } from '@lumino/disposable'
+import { IDisposable } from '@lumino/disposable';
 import { ISignal, Signal } from '@lumino/signaling';
 
 import * as decoding from 'lib0/decoding';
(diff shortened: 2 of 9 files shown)


packages/docprovider/src/tokens.ts
```diff
@@ -45,3 +51,27 @@ export interface ISharedModelFactory extends Contents.ISharedFactory {
...
+export type IAwareness = Awareness;
+
+/**
+ * A provider interface for global awareness features.
+ */
+export interface IAwarenessProvider {
+  /**
+   * A signal to subscribe for incoming messages.
+   */
+  get chatMessage(): ISignal<this, IChatMessage>;
```
fcollonval, line 69: Should we use the new stream instead of the signal?

Ref: https://lumino.readthedocs.io/en/latest/api/interfaces/signaling.IStream.html

fcollonval: Should it be rather than a getter?

```suggestion
  readonly chatMessage: ISignal<this, IChatMessage>;
```

> This is the usual pattern in Lab

fcollonval: I would also go for another naming like `messageStream`; something closer to what it provides. `chatMessage` could be misinterpreted, is it the latest message.

packages/collaboration/package.json
```diff
@@ -41,6 +41,7 @@
   "dependencies": {
     "@codemirror/state": "^6.2.0",
     "@codemirror/view": "^6.7.0",
+    "@jupyter/docprovider": "1.0.0",
```
fcollonval, line 44: missing `^1.0.0`

packages/docprovider/src/awareness.ts
```diff
@@ -0,0 +1,141 @@
...
+    this.messageHandlers[MessageType.CHAT] = (
+      encoder,
+      decoder,
+      provider,
+      emitSynced,
+      messageType
+    ) => {
+      const content = decoding.readVarString(decoder);
+      const data = JSON.parse(content) as IChatMessage;
+      console.debug('Chat:', data);
```
fcollonval, line 65: Should we keep this?

packages/docprovider/src/tokens.ts
```diff
@@ -45,3 +51,27 @@ export interface ISharedModelFactory extends Contents.ISharedFactory {
     factory: SharedDocumentFactory
   ): void;
 }
+
+/**
+ * The awareness interface.
+ *
+ * TODO: Move to @jupyter/YDoc
```
fcollonval, line 58: :+1:

PR #175, opened 2023-06-30
Improves error handling when the file does not exist

hbcarlos:


jupyter_collaboration/handlers.py
```diff
@@ -184,13 +184,27 @@ async def open(self, room_id):
...
                 await self.room.initialize()
+            except web.HTTPError as e:
+                _, _, file_id = decode_file_path(self._room_id)
+                file = self._file_loaders[file_id]
+                # Close websocket and propagate error.
+                self.log.error(f"File {file.path} not found.\n{e!r}", exc_info=e)
+                self.close(1004, f"File {file.path} not found.")
+                # Clean up the room and delete the file loader
+                self._cleanup_delay = 0
+                self._clean_room()
```
fcollonval, line 195: We should only clean the room if this handler is the only room client.

jupyter_collaboration/handlers.py
```diff
@@ -184,13 +184,27 @@ async def open(self, room_id):
...
                 _, _, file_id = decode_file_path(self._room_id)
                 file = self._file_loaders[file_id]
+                # Close websocket and propagate error.
                 self.log.error(f"Error initializing: {file.path}\n{e!r}", exc_info=e)
                 self.close(
                     1003, f"Error initializing: {file.path}. You need to close the document."
                 )
+                # Clean up the room and delete the file loader
+                self._cleanup_delay = 0
+                self._clean_room()
```
fcollonval, line 207: Same comment here.

packages/docprovider/src/ydrive.ts
```diff
@@ -75,9 +75,15 @@ export class YDrive extends Drive implements ICollaborativeDrive {
...
-        return model;
+        // If the document does't exist, `super.get` will reject with an
+        // error and the provider will never be resolved.
+        // Use `Promise.all` to reject as soon as possible. The Context will
+        // show a dialog to the user.
+        const resp = await Promise.all([
+          super.get(localPath, { ...options, content: false }),
+          provider.ready
+        ]);
+        return resp[0];
```
fcollonval, line 86: You could use array destructuration for this:

```suggestion
        const [model] = await Promise.all([
          super.get(localPath, { ...options, content: false }),
          provider.ready
        ]);
        return model;
```

jupyter_collaboration/handlers.py
```diff
@@ -184,13 +184,27 @@ async def open(self, room_id):
             try:
                 # Initialize the room
                 await self.room.initialize()
+            except web.HTTPError as e:
+                _, _, file_id = decode_file_path(self._room_id)
+                file = self._file_loaders[file_id]
+                # Close websocket and propagate error.
+                self.log.error(f"File {file.path} not found.\n{e!r}", exc_info=e)
```
fcollonval, line 191: To avoid code duplication, I would only keep `except Exception as e` and test for `e` type to define the error message. This will allow to get only a single cleaning code avoiding discrepancy.

Are we also sure that any `web.HTTPError` implies the file is not found? This seems a bit to broad.

fcollonval (changes requested):
Thanks @hbcarlos 

Cleaning the room on error is probably to strict without checking if the websocket is the only client or not. Would you mind adding some protection?

jupyter_collaboration/handlers.py
hbcarlos: > Are we also sure that any web.HTTPError implies the file is not found? This seems a bit to broad.

That's the error type the content manager raises for a file not found.

jupyter_collaboration/rooms.py
```diff
@@ -203,13 +203,13 @@ def _emit(self, level: LogLevel, action: str | None = None, msg: str | None = No
 
         self._logger.emit(schema_id=JUPYTER_COLLABORATION_EVENTS_URI, data=data)
 
-    def _clean(self) -> None:
+    def stop(self) -> None:
         """
         Cleans the rooms.
```
fcollonval, line 208:
```suggestion
        Stop the room.
```

fcollonval (approved):
Thanks @hbcarlos 

I update a doc string to align it with the API name.

PR #184, opened 2023-08-11
Clean up

hbcarlos:
Removes print and the `jsonschema` dependency. The dependency was pinned to avoid the CI failures because of a deprecation warning.


fcollonval:
@hbcarlos is it the removal of the pinned version of `jsonschema` that is breaking the CI?

hbcarlos:
I think so. the error comes from `referencing`,

PR #186, opened 2023-08-11
Pin jupyter_events

### LINKED ISSUE #185: Updating jsonschema and jupyter_events dependencies
### Problem

The current pin to `jsonschema <4.18.0` makes this package incompatible with `jupyter_events >=0.7.0`, and the `0.6.x` line is unlikely to get further updates barring severe security issues, nor is the `0.7.0` line likely to backpedal on its upstream deps.

As `jupyter_events` is a hard dependency of `jupyter_server`, this will become a larger issue in the future.

### Proposed Solution

Either:

- raise bottom pins to:
  - `jsonschema >=4.18.0`
  - `jupyter_events >=0.7.0`
- determine how to support the entire range (significantly harder)

And then:

- declare minimum versions, removing top pins
- test against the minimum declared versions (at least in CI)

### Additional context

c/f https://github.com/conda-forge/jupyter-collaboration-feedstock/pull/2#issuecomment-1674877877

hbcarlos:
Fixes #185


bollwyvl:
Looks like the failing workflows need `python-version` set for their `setup-python` action to avoid getting whatever ubuntu's happens to have at the time in `dist-packages`.

PR #212, opened 2023-11-09
Update README.md

1reverseengineer:
Changed the conda install method to reflect the fact that the collab module is now available on the conda-forge channel. Without this change, it will not be possible to use conda to install the module


fcollonval (approved):
Thanks @1reverseengineer

PR #220, opened 2023-12-08
Use pycrdt-websocket instead of ypy-websocket

davidbrochart:


davidbrochart:
I will publish v2.0.0 once this is merged.

PR #221, opened 2023-12-11
test

davidbrochart:


trungleduc:
you can remove the "packageManager" field in the package,json file

davidbrochart:
Thanks @trungleduc, that seems to have solved the issue. I'll investigate the next issues.

davidbrochart:
If I commit the changes to `yarn.lock` reported [here](https://github.com/jupyterlab/jupyter-collaboration/actions/runs/7182406324/job/19559132873?pr=221), it [fails](https://github.com/jupyterlab/jupyter-collaboration/pull/221/commits/baf7ac1c9b2f790fef215ba7d00f5e25d789418c).
Any idea? Otherwise I am going to merge.

krassowski:
This is the typescript patch applied by yarn. If you change `yarn` version (directly or by changing updating some jupyterlab versions, e.g. from 4.0 to 4.1.0a4 which changes `yarn` pin in `jlpm`) a different patch version gets used leading to this problem. See https://github.com/jupyterlab/jupyterlab/issues/15503

krassowski:
So you basically need to ensure that your local environment and CI use the same yarn/jlpm/jupyterlab version.

davidbrochart:
Thanks @krassowski.
I tried to have the same environment with `pixi add nodejs=18 yarn=1.22 pip python=3.11`, but `pip install -e ".[test]"` doesn't update `yarn.lock` locally.

krassowski:
What about exact `jupyterlab` version?

krassowski:
Wait, it fails on `pip install --pre -e ".[test]"` (emphasis on `--pre`) so that must be pulling JupyterLab 4.1.0a4, so this is _exactly_ the issue https://github.com/jupyterlab/jupyterlab/issues/15503

davidbrochart:
I don't know, `hatch` requires `jupyterlab>=4.0.0` but I how can I know the exact version in the CI?

krassowski:
Yes, you get the failure in different jobs. With the old patch version you get failure in `Tests / Test Prereleases` which uses `pip install --pre -e ".[test]"` (with `--pre`) and with the new patch version you get a failure in all other jobs as these use `pip install -e ".[test]"` (without `--pre`).

This is like super silly `yarn`/`ts` situation which keeps going for like four years now? In every other yarn release they have a PR like https://github.com/yarnpkg/berry/pull/5916 which just pull the latest version of https://github.com/microsoft/TypeScript/pull/35206 after re-basing on latest.

davidbrochart:
Thanks @krassowski, that was indeed the issue. I removed `--pre` and now tests pass :+1:

martinRenou:
meeseeksdev please backport to 1.x

davidbrochart:
meeseeksdev please backport to 1.x

PR #241, opened 2024-02-28
Fix cancellation of file save operation

davidbrochart:
When we schedule a file saving and one has already been scheduled but not yet completed (for instance, by typing a character and then another one less than one second later), we need to cancel the previously scheduled one. The proper way to do it is not by calling `.cancel()` on the task and awaiting it, but by handling the `CancelledError` exception inside the task. In our case, we want to do nothing if the task is cancelled, so we just return.

### DIFF
--- a/jupyter_collaboration/rooms.py
+++ b/jupyter_collaboration/rooms.py
@@ -235,6 +235,9 @@ class DocumentRoom(YRoom):
             # the document is being saved, cancel that
             saving_document.cancel()
 
+        # all async code (i.e. await statements) must be part of this try/except block
+        # because this coroutine is run in a cancellable task and cancellation is handled here
+
         try:
             # save after X seconds of inactivity
             await asyncio.sleep(self._save_delay)
--- a/jupyter_collaboration/rooms.py
+++ b/jupyter_collaboration/rooms.py
@@ -234,12 +234,11 @@ class DocumentRoom(YRoom):
         if saving_document is not None and not saving_document.done():
             # the document is being saved, cancel that
             saving_document.cancel()
-            await saving_document
-
-        # save after X seconds of inactivity
-        await asyncio.sleep(self._save_delay)
 
         try:
+            # save after X seconds of inactivity
+            await asyncio.sleep(self._save_delay)
+
             self.log.info("Saving the content from room %s", self._room_id)
             await self._file.maybe_save_content(
                 {
@@ -252,6 +251,9 @@ class DocumentRoom(YRoom):
                 self._document.dirty = False
 
             self._emit(LogLevel.INFO, "save", "Content saved.")
+
+        except asyncio.CancelledError:
+            return
 
         except OutOfBandChanges:
             self.log.info("Out-of-band changes. Overwriting the content in room %s", self._room_id)


Zsailer:
@davidbrochart are the UI tests pretty flaky? At quick glance, these failures appear unrelated. The first few failures are due to the collaborations panel being slow to update.

ellisonbg:
In the current implementation what triggers a save to disk? Wondering how these re-entrant saves happen and if there is a deeper issue for us to solve.

davidbrochart:
Yes I think failures are unrelated, we will need to look into this.

Zsailer:
> I think we want to avoid triggering saves (even if there is a pause) on user actions (editing the document as this will increase load on the server as the number of users and edits increases.

I totally agree here. 

> A simpler approach would be to set a periodic timer (say every 60 seconds) and save to disk if there have been changes in that 60s.

I was going to suggest the same thing. 

I think the _this_ PR still fixes a bug and should be merged. If two saves are queued, we can cancel one. But I agree with @ellisonbg, a "periodic backup" instead of user-action-driven saving mechanism would avoid slamming the server, especially as these documents scale larger.

ellisonbg:
Sounds good, let's review and merge this.

jupyter_collaboration/rooms.py
```diff
@@ -253,6 +252,9 @@ async def _maybe_save_document(self, saving_document: asyncio.Task | None) -> No
 
             self._emit(LogLevel.INFO, "save", "Content saved.")
 
+        except asyncio.CancelledError:
```
Zsailer, line 255: Man, this whole block of logic is hard to follow (often true with asyncio cancellation logic). Maybe we need a comment here explaining the most likely reason we'd hit this exception, i.e. a subsequent, future `_maybe_save_document` call happened that cancelled the current one.

jupyter_collaboration/rooms.py
```diff
@@ -234,12 +234,11 @@ async def _maybe_save_document(self, saving_document: asyncio.Task | None) -> No
         if saving_document is not None and not saving_document.done():
             # the document is being saved, cancel that
             saving_document.cancel()
```
Zsailer, line 236: Is there a risk of corruption when cancelling a previous save? 

If the previous save if in the middle of a I/O action, would cancelling this task corrupt the file on disk?

Zsailer:
@davidbrochart, I think this codebase could really use a lot more commenting, because the async task flow is hard to hold in the brain 😅. I'm mentioning this, because I don't want to annoy you when asking for more comments as-we-go in these code reviews. 😄

jupyter_collaboration/rooms.py
davidbrochart: No because the saving task is [shielded](https://github.com/jupyterlab/jupyter-collaboration/blob/d4ec54a78ee60c364805643f72019852f62ac516/jupyter_collaboration/loaders.py#L145).

Zsailer: Thanks for pointing me to this line. I'm still catching up on this codebase :)

Zsailer:
> That's not how Google Docs works, and as a user it feels strange to me: sometimes I will see the document being saved right after a change, sometimes a long time after. I think Google Docs has set some kind of standard, as far as this should behave.

The server-side serialization of the document to its backend storage is opaque to the user in Google Docs, right? It's running as a cloud service, not a local application like most of our users. So I don't know if we can directly compare here. 

In a "Jupyter cloud case", the UX for this doesn't necessarily need to show when the saving happens in the backend, i.e. it can be opaque too.

dlqqq (changes requested):
I believe this is correct, but @Zsailer is right to call out the need for additional doc comments. As @ellisonbg called out, we will want to change this logic in the near future, so I think leaving this untested is OK for the time being.

@davidbrochart There should be a doc comment saying something like, "Note to developers: make sure all `await` statements are wrapped in the `try` block below."

@Zsailer I can offer some insight into why this PR works. Basically, calling `task.cancel()` does not actually cancel the task. Instead, it raises a `asyncio.CancelledError` exception the next time that the task is re-entered by the event loop. _This occurs immediately after every `await` statement._ This PR moves all  `await` statements into the `try` block and catches the `asyncio.CancelledError` exception silently. This allows the task `self._maybe_save_document(...)` to be cancelable as expected.

davidbrochart:
@Zsailer I think it's important to show that edits are saved or not, for instance JupyterLab will still warn you that changes are not saved if you try to exit when changes are not saved. If you didn't have that, you could really close everything and loose your changes.

> Lastly, as @ellisonbg called out, we will want to change this logic in the near future, so I think leaving this untested is OK for the time being.

@dlqqq I think I made good arguments against Brian's [comments](https://github.com/jupyterlab/jupyter-collaboration/pull/241#issuecomment-1969556612).

jupyter_collaboration/rooms.py
dlqqq: @Zsailer To clarify: there are three tasks that get run (most recent call last):

1. `DocumentRoom._maybe_save_document()`
2. `FileLoader.maybe_save_content()`
3. `FileLoader._save_content()`

(3) is the task writing to disk, and is shielded by (2) after it is created:

```
task = asyncio.create_task(self._save_content(model, done_saving))
try:
    await asyncio.shield(task)
except asyncio.CancelledError:
    pass
```

So in this context, (2) is a parent task of (3). When (2) runs (3) with `await asyncio.shield(task)`, this means that even if (1) or (2) get cancelled, the body of (3) will still proceed as expected.

Zsailer:
> I think it's important to show that edits are saved or not, for instance JupyterLab will still warn you that changes are not saved if you try to exit when changes are not saved. If you didn't have that, you could really close everything and loose your changes.

Fair enough. I just opened Google docs to see what they do here. I didn't realize they actually show a "Saving..." message at the top of the doc every time a new keystroke happens. 

Screenshot 2024-02-28 at 2 50 51 PM

davidbrochart:
Released in [v2.0.3](https://github.com/jupyterlab/jupyter-collaboration/releases/tag/v2.0.3).

PR #246, opened 2024-03-12
adding awareness event when open and close websockets

jzhang20133:


jupyter_collaboration/handlers.py
```diff
@@ -184,6 +185,7 @@ async def open(self, room_id):
             try:
                 # Initialize the room
                 await self.room.initialize()
+                self._emit_awareness_event(LogLevel.INFO, self.current_user.name, "join")
```
Zsailer, line 188: `name` isn't required on the `User` object and it isn't necessarily unique. Let's use `username` here. 

https://github.com/jupyter-server/jupyter_server/blob/df27810d56f4fa501b84dad7284ba0a198e659c8/jupyter_server/auth/identity.py#L40-L44

jupyter_collaboration/events/awareness.yaml
```diff
@@ -0,0 +1,43 @@
...
+  - user
+  - action
+properties:
+  level:
+    enum:
+      - INFO
+      - DEBUG
+      - WARNING
+      - ERROR
+      - CRITICAL
```
Zsailer, line 21: I don't think `level` is needed in this event, since this isn't really logging type event, but rather an action in the server.

jupyter_collaboration/handlers.py
jzhang20133: will address

jupyter_collaboration/events/awareness.yaml
```diff
@@ -0,0 +1,43 @@
...
+    description: |
+      Message type.
+  room:
+    type: string
+    description: |
+      Room ID. Usually composed by the file type, format and ID.
+  user:
```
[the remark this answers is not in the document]
jzhang20133: In some of the user model that extended parent User class, there might be jwt token in those user model, not sure if it is secure to emit the whole model. username identifier is a required field in the User class and we could use that.

jupyter_collaboration/events/awareness.yaml
jzhang20133: make sense, will address.

jupyter_collaboration/events/awareness.yaml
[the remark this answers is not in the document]
Zsailer: That's true. The user model might be harder to screen for more sensitive information.

Let's keep it username then?

jupyter_collaboration/utils.py
```diff
@@ -7,6 +7,10 @@
 
 JUPYTER_COLLABORATION_EVENTS_URI = "https://schema.jupyter.org/jupyter_collaboration/session/v1"
 EVENTS_SCHEMA_PATH = pathlib.Path(__file__).parent / "events" / "session.yaml"
+JUPYTER_COLLABORATION_AWARENESS_EVENTS_URI = (
+    "https://schema.jupyter.org/jupyter_collaboration/awareness/v1"
+)
+AWARENESS_EVENTS_SCHEMA_PATH = pathlib.Path(__file__).parent / "events" / "awareness.yaml"
```
davidbrochart, line 13: I see `pathlib.Path(__file__).parent` is used twice, maybe worth storing it in a variable and not repeat the code?
Also, we can do `from pathlib import Path`.

jupyter_collaboration/events/awareness.yaml
```diff
@@ -0,0 +1,33 @@
...
+personal-data: true
+description: |
+  Awareness events emitted from server-side during a collaborative session.
+type: object
+required:
+  - room
+  - username
+  - action
+properties:
+  room:
```
davidbrochart, line 14: Maybe:
```suggestion
  roomid:
```

jupyter_collaboration/handlers.py
```diff
@@ -284,6 +286,8 @@ def on_close(self) -> None:
             # keep the document for a while in case someone reconnects
             self.log.info("Cleaning room: %s", self._room_id)
             self.room.cleaner = asyncio.create_task(self._clean_room())
+        if isinstance(self.room, DocumentRoom):
```
davidbrochart, line 289: Why not emitting the event for all room types?

jupyter_collaboration/handlers.py
```diff
@@ -297,6 +301,13 @@ def _emit(self, level: LogLevel, action: str | None = None, msg: str | None = No
 
         self.event_logger.emit(schema_id=JUPYTER_COLLABORATION_EVENTS_URI, data=data)
 
+    def _emit_awareness_event(self, user: str, action: str, msg: str | None = None) -> None:
```
davidbrochart, line 304: Rename `user` to `username`?

dlqqq, line 304: (no action needed) I think we should prefer declaring the type of `msg` as `Optional[str]` instead of `str | None`, since the vertical bar syntax for type unions is [only available for Python 3.10+](https://peps.python.org/pep-0604/). I thought the convention in Jupyter packages was to support every currently maintained version of Python, [the lowest being Python 3.8 at present](https://devguide.python.org/versions/).

However, I also noticed that the rest of the source code uses the `str | None`. I've [opened an issue](https://github.com/jupyterlab/jupyter-collaboration/issues/248) so we can discuss this as a team. No action needed on this PR, just a comment.

jupyter_collaboration/handlers.py
dlqqq: +1

dlqqq (commented):
@jzhang20133 Thank you opening this PR! I left a comment below, but David Brochart's review covers the areas of feedback blocking this PR from being merged. 😁 
 
All I would like to add is that it's possible to emit events through a decorator interface. Emitting events imperatively (as done in this PR) is less preferable, since it's really easy for a developer to accidentally delete the line emitting the event. Using a decorator also allows for this feature to be added without any changes to the method definition, which I take as a sign of clean code.

To implement a decorator interface, you would want to define an `@emit_awareness_event` decorator that looks something like:

```python
def emit_awareness_event(action: str, message: Optional[str] = None):
    def decorator(method):
        async def method_stub(self, *args, **kwargs):
            return_value = await method(self, *args, **kwargs)
            data = { "action": action, ... }
            self.event_logger.emit(..., data=data)
            return return_value
        return method_stub
    return decorator
```

After doing so, the method declarations would look like:

```python
    @emit_awareness_event("join")
    async def open(self, room_id):
        ...

    @emit_awareness_event("leave")
    def on_close(self) -> None:
        ...
```

Let me know if this all sounds like a good idea for future work. If so, I can track this in a new issue so as to not block this PR. Thanks! 🤗

jupyter_collaboration/handlers.py
jzhang20133: For YdocWebsocketHandler that handle global awareness websocket, its room type will be TransientRoom. I found its room id is fixed, called "JupyterLab:globalAwareness" and its creating and closing does not indicate any user joining or leaving the collaboration on a document. Hence we only emit when a websocket close where its room type is DocumentRoom.

jupyter_collaboration/handlers.py
jzhang20133: will address

jupyter_collaboration/events/awareness.yaml
jzhang20133: will address

jupyter_collaboration/utils.py
jzhang20133: will address

jzhang20133:
> ```python
> def emit_awareness_event(action: str, message: Optional[str] = None):
>     def decorator(method):
>         async def method_stub(self, *args, **kwargs):
>             return_value = await method(self, *args, **kwargs)
>             data = { "action": action, ... }
>             self.event_logger.emit(..., data=data)
>             return return_value
>         return method_stub
>     return decorator
> ```

I really love this idea! I just tried it out. In this use case, we only emit when websocket serving a DocumentRoom is successfully opened and closed and in open(self, room_id) method, when initialize room fails, it handles certain type exception by closing websocket with error message and swallows other type of exceptions. And for that failure cases, we won't emit this event. We would need to change original logic in open method in order to leverage decorator which defeat the purpose.

jupyter_collaboration/handlers.py
davidbrochart: It's fine to use `str | None`. Jupyter collaboration [supports Python 3.8 and above](https://github.com/jupyterlab/jupyter-collaboration/blob/8fee21f7cd338057c69febb5508f9b6c8bfd2392/pyproject.toml#L12). Union types as `X | Y` are allowed for python<3.10 with:
```py
from __future__ import annotations
```

jupyter_collaboration/handlers.py
davidbrochart: Global awareness being a `TransientRoom` doesn't mean that clients leaving a document in a `TransientRoom` should not emit an event. Maybe the event should not be emitted specifically when the room ID is `JupyterLab:globalAwareness`? This room ID is a bit unfortunate, as it should probably be a standard in the Jupyter ecosystem and not particularly for JupyterLab, so a better would have been `Jupyter:globalAwareness`.
@Zsailer do you think this room ID should be specified somewhere in the Jupyter schemas, since it's probably a convention we want to rely on?

jzhang20133: We would need to rely on RoomId to identity which document that the host has left from and close all the guest websockets associated as clients in that DocumentRoom. For a TransientRoom it seems that it is not tied to a specific document and mainly used by global awareness. We only want to emit an event when open and on_close has been successful as well.

davidbrochart: All I'm saying is that you could do something like that:
```suggestion
        if self.room.room_id != "JupyterLab:globalAwareness":
```
A `TransientRoom` is not just for global awareness, it could be for any other collaborative document that doesn't need to be persisted, and you probably want to emit join/leave events for them too.

jzhang20133: Sure.

davidbrochart (commented):
Thanks @jzhang20133, I think this is good to go, but it would be great if you could add a test.

jzhang20133:
> Thanks @jzhang20133, I think this is good to go, but it would be great if you could add a test.

@davidbrochart Thank you for reviewing the PR. I have added a test to cover emit awareness event.

Zsailer (approved):
This looks good, @jzhang20133! 

I think we need to set the minimum version of Jupyter Events to the latest release, v0.10.0.

davidbrochart (approved):
Thanks @jzhang20133.

davidbrochart:
@jzhang20133 Test `test_room_handler_doc_client_should_emit_awareness_event` [fails](https://github.com/jupyterlab/jupyter-collaboration/actions/runs/8349958209/job/22855547289) in the CI, would you mind taking a look? Thanks!

Zsailer:
@davidbrochart fixed here: https://github.com/jupyterlab/jupyter-collaboration/pull/258

PR #250, opened 2024-03-15
Refactor `DocumentRoom`, throttle saves instead of debouncing saves

dlqqq:
## Description

- Refactors the `DocumentRoom` class to not rely on the `asyncio.Lock` context manager. This greatly improves readability by reducing nesting and reliance on the `asyncio` API.

- Throttles saves instead of debouncing them, ensuring that document changes will be flushed to disk on a minimum interval. Related: #244

- All existing unit tests pass locally.

## Change summary

- Removes `self._initialization_lock`.
    - This is not necessary as long as the `initialize()` method is only called once and awaited, as is the case in this extension. This is a very reasonable usage constraint. Furthermore, in `asyncio`, [locks do not provide thread safety](https://docs.python.org/3/library/asyncio-sync.html#asyncio.Lock), contrary to the original docstring of this method.

- Removes `self._update_lock`.
  - The purpose of this lock is to prevent `self._on_document_change()` from being called while the lock is held. Removing the locks results in a save loop, as without the lock, `self._maybe_save_document()` would trigger `self._on_document_change()`.
  - However, the save loop was caused by a single `self._document.dirty = False` statement, which could have been removed without consequence. See below.

- Throttles saves instead of debouncing saves by cancelling the previous task
  - Throttling seems preferable to debouncing here, as debouncing could result in the document not being saved if the document is being changed too frequently, which may arise in rooms with lots of collaborators.
  - Previously, every time a new `self._maybe_save_document()` task was started, the previous task was cancelled if it was in progress. The method required the previous task as an argument, which resulted in a weird way of calling the method:

    ```python
    self._saving_document = asyncio.create_task(
        self._maybe_save_document(self._saving_document)
    )
    ```

  - The new implementation of `self._maybe_save_document()` does not require an extra argument or `task.cancel()`; instead it relies on knowledge of its own state, stored in a couple of new instance attributes. Here is an overview of the algorithm:
    - If a previous `self._maybe_save_document()` task is waiting on `self._save_delay`, then the current task can return early, as that previous task will save the Ydoc anyways. 
    - If a previous `self._maybe_save_document()` task is currently saving via `FileLoader`, then the current task should set `self._should_resave = True` and then return. Later, when the previous task is done saving, if this attribute is `True`, then it will re-run itself via `asyncio.create_task(self.maybe_save_document())`.

- Removed `self._document.dirty = False` statements
  - This `dirty` attribute was only referenced in a single unit test, and I could not find it mentioned in `pycrdt` or `pycrdt-websocket`. I removed this because setting `self._document.dirty` triggers the `_on_document_change()` observer, causing a save loop in this branch. Removing the statement from `self._maybe_save_document()` allows for `self._update_lock` to be removed.
  - It seems preferable to avoid triggering a document change when handling one, rather than relying on a lock.

### DIFF
--- a/jupyter_collaboration/rooms.py
+++ b/jupyter_collaboration/rooms.py
@@ -120,7 +120,16 @@ class DocumentRoom(YRoom):
                         # YDoc not found in the YStore, create the document from the source file (no change history)
                         pass
 
-                if not read_from_source:
+                if read_from_source:
+                    self._emit(LogLevel.INFO, "load", "Content loaded from disk.")
+                    self.log.info(
+                        "Content in room %s loaded from file %s", self._room_id, self._file.path
+                    )
+                    self._document.source = model["content"]
+
+                    if self.ystore:
+                        await self.ystore.encode_state_as_update(self.ydoc)
+                else:
                     # if YStore updates and source file are out-of-sync, resync updates with source
                     if self._document.source != model["content"]:
                         # TODO: Delete document from the store.
@@ -134,15 +143,6 @@ class DocumentRoom(YRoom):
                         )
                         read_from_source = True
 
-                if read_from_source:
-                    self._emit(LogLevel.INFO, "load", "Content loaded from disk.")
-                    self.log.info(
-                        "Content in room %s loaded from file %s", self._room_id, self._file.path
-                    )
-                    self._document.source = model["content"]
-
-                    if self.ystore:
-                        await self.ystore.encode_state_as_update(self.ydoc)
 
                 self._document.dirty = False
                 self.ready = True
--- a/jupyter_collaboration/rooms.py
+++ b/jupyter_collaboration/rooms.py
@@ -131,6 +131,19 @@ class DocumentRoom(YRoom):
                 # YDoc not found in the YStore, create the document from the source file (no change history)
                 pass
 
+        if not read_from_source and self._document.source != model["content"]:
+            # TODO: Delete document from the store.
+            self._emit(
+                LogLevel.INFO, "initialize", "The file is out-of-sync with the ystore."
+            )
+            self.log.info(
+                "Content in file %s is out-of-sync with the ystore %s",
+                self._file.path,
+                self.ystore.__class__.__name__,
+            )
+            read_from_source = True
+
+        # if YStore updates and source file are out-of-sync, resync updates with source
         if read_from_source:
             self._emit(LogLevel.INFO, "load", "Content loaded from disk.")
             self.log.info(
@@ -140,20 +153,6 @@ class DocumentRoom(YRoom):
 
             if self.ystore:
                 await self.ystore.encode_state_as_update(self.ydoc)
-        else:
-            # if YStore updates and source file are out-of-sync, resync updates with source
-            if self._document.source != model["content"]:
-                # TODO: Delete document from the store.
-                self._emit(
-                    LogLevel.INFO, "initialize", "The file is out-of-sync with the ystore."
-                )
-                self.log.info(
-                    "Content in file %s is out-of-sync with the ystore %s",
-                    self._file.path,
-                    self.ystore.__class__.__name__,
-                )
-                read_from_source = True
-
 
         self.ready = True
         self._emit(LogLevel.INFO, "initialize", "Room initialized")
--- a/jupyter_collaboration/rooms.py
+++ b/jupyter_collaboration/rooms.py
@@ -43,11 +43,24 @@ class DocumentRoom(YRoom):
         self._logger = logger
         self._save_delay = save_delay
 
-        self._update_lock = asyncio.Lock()
-        self._initialization_lock = asyncio.Lock()
         self._cleaner: asyncio.Task | None = None
-        self._saving_document: asyncio.Task | None = None
-        self._messages: dict[str, asyncio.Lock] = {}
+
+        # the task currently saving to disk via FileLoader, if any.
+        self._save_task: asyncio.Task | None = None
+
+        # flag that indicates whether a previous call to
+        # `self._maybe_save_document()` is waiting to save.
+        #
+        # if `self._maybe_save_document()` is called while this flag is `True`,
+        # then the call does nothing, as a previous task will save the Ydoc
+        # within `self._save_delay` seconds.
+        self._waiting_to_save = False
+
+        # flag that indicates whether a previous call to
+        # `self._maybe_save_document()` should call itself again after
+        # `self._save_task` finishes. this is set to `True` when a document
+        # update occurs while `self._save_task` is running.
+        self._should_resave = False
 
         # Listen for document changes
         self._document.observe(self._on_document_change)
@@ -78,7 +91,8 @@ class DocumentRoom(YRoom):
         """
         Initializes the room.
 
-        This method is thread safe so only one client can initialize the room.
+        This method is not coroutine-safe, i.e. consumers must await this method
+        before calling any other methods.
 
         To initialize the room, we check if the content was already in the store
         as a Y updates and if it is up to date with the content on disk. In this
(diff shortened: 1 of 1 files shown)


jupyter_collaboration/rooms.py
```diff
@@ -89,64 +103,59 @@ async def initialize(self) -> None:
...
-                        "Content in room %s loaded from file %s", self._room_id, self._file.path
-                    )
-                    self._document.source = model["content"]
-
-                    if self.ystore:
-                        await self.ystore.encode_state_as_update(self.ydoc)
-
-                self._document.dirty = False
-                self.ready = True
-                self._emit(LogLevel.INFO, "initialize", "Room initialized")
```
dlqqq, line 149: Aside from removing the locks and the `self._document.dirty = False` statements, there are no code changes in this section. Most of my substantive code changes are in the `_maybe_save_document()` method.

davidbrochart (changes requested):
At first sight, this breaks the dirty indicator in JupyterLab's UI.
I created a [test](https://github.com/jupyterlab/jupyter-collaboration/pull/251) that passes on `main` and fails with this PR.

jupyter_collaboration/rooms.py
```diff
@@ -212,49 +219,60 @@ def _on_document_change(self, target: str, event: Any) -> None:
             document. This tasks are debounced (60 seconds by default) so we
             need to cancel previous tasks before creating a new one.
         """
-        if self._update_lock.locked():
-            return
-
-        self._saving_document = asyncio.create_task(
-            self._maybe_save_document(self._saving_document)
-        )
+        asyncio.create_task(self._maybe_save_document())
```
davidbrochart, line 222: Always [save a reference to the task](https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task).

jupyter_collaboration/rooms.py
```diff
@@ -212,49 +219,60 @@ def _on_document_change(self, target: str, event: Any) -> None:
...
         try:
-            # save after X seconds of inactivity
-            await asyncio.sleep(self._save_delay)
-
+            # do not write to `self._document` in this `try` block, as that will
+            # trigger the observer and result in a save loop.
             self.log.info("Saving the content from room %s", self._room_id)
-            await self._file.maybe_save_content(
+            loop = asyncio.get_running_loop()
+            self._save_task = loop.create_task(self._file.maybe_save_content(
```
davidbrochart, line 262:
```suggestion
            self._save_task = asyncio.create_task(self._file.maybe_save_content(
```

jupyter_collaboration/rooms.py
```diff
@@ -212,49 +219,60 @@ def _on_document_change(self, target: str, event: Any) -> None:
...
-            async with self._update_lock:
-                self._document.dirty = False
-
+            ))
+            await self._save_task
             self._emit(LogLevel.INFO, "save", "Content saved.")
 
+            if self._should_resave:
+                self._should_resave = False
+                asyncio.create_task(self._maybe_save_document())
```
davidbrochart, line 274: Always [save a reference to the task](https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task).

davidbrochart:
> Removes `self._initialization_lock`.
> 
> * This is not necessary as long as the `initialize()` method is only called once and awaited, as is the case in this extension. This is a very reasonable usage constraint. Furthermore, in `asyncio`, [locks do not provide thread safety](https://docs.python.org/3/library/asyncio-sync.html#asyncio.Lock), contrary to the original docstring of this method.

How do you handle the case where the room is created at the same time from two different clients?

> Removes `self._update_lock`.
> 
> * The purpose of this lock is to prevent `self._on_document_change()` from being called while the lock is held. Removing the locks results in a save loop, as without the lock, `self._maybe_save_document()` would trigger `self._on_document_change()`.
> * However, the save loop was caused by a single `self._document.dirty = False` statement, which could have been removed without consequence. See below.

Removing `self._document.dirty = False` had a [consequence](https://github.com/jupyterlab/jupyter-collaboration/pull/250#pullrequestreview-1939045015).

> Throttles saves instead of debouncing saves by cancelling the previous task
> 
> * The new implementation of `self._maybe_save_document()` does not require an extra argument or `task.cancel()`; instead it relies on knowledge of its own state, stored in a couple of new instance attributes. Here is an overview of the algorithm:
>   
>   * If a previous `self._maybe_save_document()` task is waiting on `self._save_delay`, then the current task can return early, as that previous task will save the Ydoc anyways.
>   * If a previous `self._maybe_save_document()` task is currently saving via `FileLoader`, then the current task should set `self._should_resave = True` and then return. Later, when the previous task is done saving, if this attribute is `True`, then it will re-run itself via `asyncio.create_task(self.maybe_save_document())`.

I don't think it's any simpler than previously, as there is now a possible recursive call to `self._maybe_save_document()`.
I don't quite see how the class attributes this method is using are not going to collide with nested calls of this method.

> Removed `self._document.dirty = False` statements
> 
> * This `dirty` attribute was only referenced in a single unit test, and I could not find it mentioned in `pycrdt` or `pycrdt-websocket`. I removed this because setting `self._document.dirty` triggers the `_on_document_change()` observer, causing a save loop in this branch. Removing the statement from `self._maybe_save_document()` allows for `self._update_lock` to be removed.

The `dirty` attribute is used by frontends such as JupyterLab to set/clear a dirty indicator, just like Google Docs shows "Saving..."/"Saved to Drive".

dlqqq:
> The dirty attribute is used by frontends such as JupyterLab to set/clear a dirty indicator, just like Google Docs shows "Saving..."/"Saved to Drive".

Where is `dirty` being set to `True`? I cannot find any mention of this in the source.

jupyter_collaboration/rooms.py
dlqqq: Thanks for this callout! I was not aware of this detail.

dlqqq: Fixed in latest revision.

jupyter_collaboration/rooms.py
dlqqq: Thanks for catching this! Fixed in latest revision.

jupyter_collaboration/rooms.py
dlqqq: Fixed in latest revision.

dlqqq:
@davidbrochart Thank you for your feedback! It has allowed me to simplify and correct the implementation in this branch. Let me address the rest of your feedback here:

> How do you handle the case where the room is created at the same time from two different clients?

The current implementation doesn't handle this case either. Having a lock within `DocumentRoom.initialize()` won't stop the main body of the method from being run more than once concurrently, since the lock is local to each `DocumentRoom` instance. Consider the example:

```python
room1 = DocumentRoom(...)
room2 = DocumentRoom(...)
async with asyncio.TaskGroup() as tg:
    task1 = tg.create_task(room1.initialize())
    task2 = tg.create_task(room2.initialize())
```

This will still result in the main body of `initialize()` being run twice concurrently in the event loop. If you really want to prevent this, you would need to pass a lock from an outer scope as an argument to the constructor, then have the rooms wait on that outer lock before running the rest of `initialize()`.

I don't think this is an issue, since Tornado awaits the `WebsocketHandler.open()` method where `DocumentRoom.initialize()` is called:  https://github.com/tornadoweb/tornado/blob/464f6084a5e18a01b46f77c011b1bbffe1f4d96c/tornado/websocket.py#L937-L940. This means that at most one `DocumentRoom.initialize()` method should be running at any time on the event loop, which means that we shouldn't need to implement an outer lock.

However, we should probably open an issue to Tornado to verify that this method isn't ultimately run as a concurrent task upstream somewhere.

> The dirty attribute is used by frontends such as JupyterLab to set/clear a dirty indicator, just like Google Docs shows "Saving..."/"Saved to Drive".

Thanks for clarifying! I've reverted that change and added some logic to prevent a save loop without needing a lock.

> I don't think it's any simpler than previously, as there is now a possible recursive call to self._maybe_save_document().
I don't quite see how the class attributes this method is using are not going to collide with nested calls of this method.

Only one `self._maybe_save_document()` is run as a task at any given time in the current implementation. So using instance attributes is safe here. BTW, I changed the logic to be more clear, so please take another look! I think the implementation proposed by this PR is definitely simpler and more correct with regards to how `DocumentRoom` saves the Ydoc in response to user updates. 👍

dlqqq:
Proof that this PR correctly sets the `dirty` attribute:


https://github.com/jupyterlab/jupyter-collaboration/assets/44106031/3f7aa7a5-991a-4cb3-9164-13e6c81e8404

dlqqq:
This PR seems to set the `dirty` indicator more responsively than the `main` branch, shown below:

https://github.com/jupyterlab/jupyter-collaboration/assets/44106031/8b0f4aad-9b0e-47c1-8e70-e61571e57a29

I think this is mainly due to the performance benefits of simplifying the async code and removing the lock waits.

davidbrochart:
Just opening JupyterLab and creating a new notebook, it seems to be saved twice without doing any modification:
```
[I 2024-03-20 09:36:49.612 ServerApp] Connecting to kernel 727323aa-f8b0-4cac-aff6-dcccabb68e84.
[I 2024-03-20 09:36:49.892 ServerApp] Saving the content from room json:notebook:e952e112-3652-40b2-a4e6-27f90cd35f4c
[I 2024-03-20 09:36:49.895 YDocExtension] Saving file: Untitled.ipynb
[I 2024-03-20 09:36:50.918 ServerApp] Saving the content from room json:notebook:e952e112-3652-40b2-a4e6-27f90cd35f4c
[I 2024-03-20 09:36:50.922 YDocExtension] Saving file: Untitled.ipynb
```
Can you explain why?

davidbrochart:
> The current implementation doesn't handle this case either. Having a lock within `DocumentRoom.initialize()` won't stop the main body of the method from being run more than once concurrently, since the lock is local to each `DocumentRoom` instance.

Good point. I opened #255 which includes a test for concurrent room creation, which fails on this PR.

> BTW, I changed the logic to be more clear, so please take another look!

I still think that the recursive call to `_maybe_save_document()` is a code smell.

dlqqq:
@davidbrochart Thanks for the feedback!

> I still think that the recursive call to _maybe_save_document() is a code smell.

This has been fixed in the latest revision; this branch now uses a `Task.add_done_callback()` to avoid a recursive coroutine call.

> Just opening JupyterLab and creating a new notebook, it seems to be saved twice without doing any modification:
> Can you explain why?

Sure! Basically, the `main` implementation had a `self._update_lock`, and whenever `self._on_document_change()` was called while this lock was held, that update was entirely ignored and not written to disk until another update occurred while the lock was freed. This meant that in the `main` implementation, some document updates would not trigger `self._maybe_save_document()`. 

The current implementation ensures that all updates are written to disk by using the `self._should_resave` flag. Multiple Ydoc updates occur in quick succession when the room is initialized, so this ultimately results in two saves.

I tried to determine what Ydoc updates were occurring when the room is initialized, but for some reason, I wasn't able to access every `MapEvent` and `ArrayEvent` object. I added the following log statements:

```python
    def _on_document_change(self, target: str, event: Any) -> None:
        """
        ...
        """
        self.log.error(target)
        self.log.error(event)
        if isinstance(event, MapEvent):
            self.log.error(event.keys)
```

But for some reason, `isinstance(event, MapEvent)` evaluates to `False` contrary to the logs produced. Here are the logged document updates that are resulting in a double-write:

```
[E 2024-03-20 11:03:32.171 ServerApp] cells
[E 2024-03-20 11:03:32.171 ServerApp] [<pycrdt.array.ArrayEvent object at 0x10cae9900>]
[E 2024-03-20 11:03:32.171 ServerApp] cells
[E 2024-03-20 11:03:32.171 ServerApp] [<pycrdt.array.ArrayEvent object at 0x10cae9900>]
[E 2024-03-20 11:03:32.171 ServerApp] meta
[E 2024-03-20 11:03:32.171 ServerApp] [<pycrdt.map.MapEvent object at 0x10cae9e40>]
[E 2024-03-20 11:03:32.171 ServerApp] meta
[E 2024-03-20 11:03:32.171 ServerApp] [<pycrdt.map.MapEvent object at 0x10cae9e40>]
[E 2024-03-20 11:03:32.172 ServerApp] meta
[E 2024-03-20 11:03:32.172 ServerApp] [<pycrdt.map.MapEvent object at 0x10cae9bc0>]
[E 2024-03-20 11:03:32.172 ServerApp] meta
[E 2024-03-20 11:03:32.172 ServerApp] [<pycrdt.map.MapEvent object at 0x10cae9bc0>]
[E 2024-03-20 11:03:32.172 ServerApp] meta
[E 2024-03-20 11:03:32.173 ServerApp] [<pycrdt.map.MapEvent object at 0x10cae9880>]
[E 2024-03-20 11:03:32.173 ServerApp] meta
[E 2024-03-20 11:03:32.173 ServerApp] [<pycrdt.map.MapEvent object at 0x10cae9880>]
```

I don't think this issue should block this PR however; we should track this in a separate issue and address it later if possible.

Zsailer:
High level comments... 

I _do_ think this PR dramatically improves the readability of the code. Great job there, @dlqqq! 

Reviewing the code, though, it makes me a bit uneasy to change so much under the hood of the DocumentRoom without more unit test coverage (and possible some integration tests). The current codebase has been "cooking" in a released state for a little while, so I have some confidence—albeit limited—that it works relatively well. We've heard edge case issues, but don't have unit tests to measure if we're getting any closer to a better state with these changes.

Before merging impactful rewrites like this, I'd prefer we increase our test coverage first. 

> I'm happy to change this, since I'm personally indifferent. I've changed this in the latest revision; this branch now uses a Task.add_done_callback() to avoid a recursive coroutine call. See https://github.com/jupyterlab/jupyter-collaboration/commit/9f5540063fd8509bc86ab3013c3ecb9e02b8680a.

Personally—if we're aiming for readability—I like this new commit over the recursive `_maybe_save_document`. I agree with @davidbrochart that this recursion triggers some concern for me as well, though it's hard to tell without some better ways to test.

dlqqq:
@Zsailer I can add more unit test coverage in this PR. 👍

davidbrochart:
@dlqqq Thanks for identifying the bug in concurrent room initialization. A fix for it was merged in #255, and this PR now has conflicts, but it shouldn't be a problem since you are essentially rewriting the `DocumentRoom`. Please go ahead and make the changes you want, keeping in mind that the test added in #255 must pass with this PR.

PR #251, opened 2024-03-15
Add test for document dirty attribute

davidbrochart:


pyproject.toml
```diff
@@ -48,7 +48,8 @@ test = [
     "jupyter_server_fileid[test]",
     "pytest>=7.0",
     "pytest-cov",
-    "websockets"
+    "websockets",
+    "importlib_metadata >=3.6; python_version<'3.10'",
```
Zsailer, line 52: Why are you adding the `python_version` pin here?

tests/test_documents.py
```diff
@@ -0,0 +1,32 @@
...
+    file_type = "file"
+    file_path = "dummy.txt"
+    await rtc_create_file(file_path)
+    jupyter_ydoc = jupyter_ydocs[file_type]()
+
+    async with await rtc_connect_doc_client(file_format, file_type, file_path) as ws:
+        async with WebsocketProvider(jupyter_ydoc.ydoc, ws):
+            jupyter_ydoc.dirty = True
+            assert jupyter_ydoc.dirty
+            await sleep(1.5)
```
Zsailer, line 31: Does this work _only_ because the [`file_poll_interval`](https://github.com/jupyterlab/jupyter-collaboration/blob/3a983e650e3e0d61cbd35e60fca4518854163b59/jupyter_collaboration/app.py#L27C5-L27C23) trait is set to 1 by default? 

If so, we might want to define a constant in this repo that is used in the trait's `default_value` argument and here in the test. That way, if we ever change the default in the future, we don't have to update it in multiple places.

Zsailer (approved):
This looks good @davidbrochart. 

I added some comments, but they shouldn't block from merging.

pyproject.toml
davidbrochart: Python 3.10+ has [importlib.metadata](https://docs.python.org/3/library/importlib.metadata.html), otherwise we need [importlib_metadata](https://github.com/python/importlib_metadata).

tests/test_documents.py
davidbrochart: Agreed, I just didn't find how to do that, even with `jp_configurable_serverapp`. Do you have an idea?

pyproject.toml
Zsailer: Ah, I see. I misunderstood what this line was doing. 

This is bringing in an optional dependency _if_ the current python version is < 3.10. 

I thought it was saying the `test` install can _only_ happen in python runtimes < 3.10.

tests/test_documents.py
Zsailer: I was just thinking we'd define a `constants.py` module with e.g.

```python
# constants.py
FILE_POLL_INTERVAL = 1
```
and use this constant as the default value of the `YDocExtension.file_poll_interval`, like
```python
# app.py
from .contants import FILE_POLL_INTERVAL

...

file_poll_interval = Float(default_value=FILE_POLL_INTERVAL
...
```

Then, in this unit test, do something like:
```python
from jupyter_collaboration.constants import FILE_POLL_INTERVAL

...
await sleep(FILE_POLL_INTERVAL + 0.5)
...
```

Zsailer: I guess this isn't great either, since if we changed `FILE_POLL_INTERVAL` to a larger value some point in the future, it would make this test very slow.

Zsailer: Can you pass the configuration through the `jp_server_config` pytest fixture? 

```python
FILE_POLL_INTERVAL = 2

@pytest.fixture
def jp_server_config():
    return {
        "YDocExtension": {"file_poll_interval": FILE_POLL_INTERVAL}
    }
```

I can't remember if this config makes it to the extension or not...

davidbrochart: Thanks, this works and I added an `rtc_document_save_delay` fixture in https://github.com/jupyterlab/jupyter-collaboration/pull/251/commits/c8fa7fdc5606ead3420e05c6e81aea23d09246d8 so that we can parameterize it.

PR #255, opened 2024-03-20
Fix concurrent room initialization

davidbrochart:


Zsailer (approved):
Thanks, @davidbrochart. This looks good to me.

Zsailer:
> I should mention that this PR not only has a test, but also a fix.

Yeah, I saw this. 

No objection. This makes sense to me. I'm not sure how this affects @dlqqq's PR. Perhaps some of this logic will change in #250, but for now this makes sense.

Zsailer:
I'm kicking the CI really fast to see if the flaky windows test passes and triggers the remaining tests. 

I think the UI tests are failing in other PRs, so I think we can ignore here.

Zsailer:
The windows 3.8 test seems to fail reliably. I've kicked it 3 times and it keeps failing. @davidbrochart can you check if it's related to your changes?

davidbrochart:
> I'm not sure how this affects @dlqqq's PR. Perhaps some of this logic will change in #250, but for now this makes sense.

Yes if #250 gets merged it will override the changes in this PR, but until then let's release since it fixes a bug.

PR #258, opened 2024-03-20
Bump jupyter_events to 0.10.0

Zsailer:
Should fix broken tests from #246


davidbrochart (approved):

davidbrochart:
Thanks Zach!

Zsailer:
Looks like we have some flakiness in the UI tests?

davidbrochart:
Yes, I think they just always fail, we need to take care of that.

PR #259, opened 2024-03-22
Make user's cursor "tooltip-on-hover" faster and wider

jzhang20133:
When you hover over a user's cursor, it shows their name. 

This PR addresses two UX issues around the current behavior:
1. The width around the cursor where this hover action happens is quite small. 
2. The delay for when the name appears is quite long. 

While these are fairly subjective settings, I believe this PR is a better default/minimum for this behavior.

### DIFF
--- a/packages/collaboration/src/cursors.ts
+++ b/packages/collaboration/src/cursors.ts
@@ -251,7 +251,7 @@ const userHover = hoverTooltip(
           continue;
         }
         // Use some margin around the cursor to display the user.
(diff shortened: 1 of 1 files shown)


ibdafna (approved):
Change looks good to me. Not sure about the units tests - looks unrelated.

Zsailer:
Thanks @ibdafna.

Yeah, the tests are failing for unrelated reasons. They are failing in all other PRs as well.

Zsailer (approved):

Issue #270, opened 2024-03-29
Public API to get a view of the shared document model


krassowski:
### Problem

Other server extensions would like to use the shared document model to access the content of the notebook without the overhead of sending it over from the frontend, and without the risk of using an outdated (or even inaccessible) copy stored on the disk.

Currently, while this is possible as demonstrated in https://github.com/jupyterlab/jupyter-ai/pull/708, it requires:
- accessing a private `_document` attribute of the `DocumentRoom`
- re-creating the room ID using `encode_file_path` and other some black magic (which has no guarantee of working if the id format changes in a future release of `jupyter-collaboration`)

### Proposed Solution

Expose a new `get_document()` method on the `YDocExtension`. It could look like:

```python
def get_document(
    self: YDocExtension,
    path: str,
    content_type: 'notebook' | 'file',
    file_format: 'json' | 'text'
) -> YBaseDoc | None:
    file_id_manager = self.settings["file_id_manager"]
    file_id = file_id_manager.index(request.path)

    # these two lines would be moved to an utility function to avoid drift from the code used in the handler.
    encoded_path = encode_file_path(file_format, content_type, file_id)
    room_id: str = encoded_path.split("/")[-1]

    try:
        room = await collaboration.ywebsocket_server.get_room(room_id)
    except RoomNotFound:
        return None

    if isinstance(room, DocumentRoom):
        document = room._document
        return document
```

One question raised by @3coins on the server call was whether such API should return a read-only view of the document:
- In case of `YNotebook` subclass this could mean an encapsulation hiding methods such as `append_cell` and `set_cell`. This would require defining read-only variants/wrapper for each of the data structures defined in [jupyter-ydoc](https://github.com/jupyter-server/jupyter_ydoc).
- Alternatively, we could use a copy of the model as implemented by @davidbrochart for forking in https://github.com/jupyterlab/jupyter-collaboration/pull/239/. Creating a copy might have performance implications; while we still do not send the notebook over wire, we are making an in-memory copy - which should be orders of magnitude faster, but still slower than not doing it.

Task list:
- [x] benchmark the performance penalty of cloning the document as done in forking (relatively to sending it over the wire)
- [ ] establish consensus on the way forward with making the result read-only
- [x] implement the new method
- [x] add tests
- [x] add documentation

### Additional context

[...]

PR #275, opened 2024-04-01
Add a public API for getting a read-only view of the shared model

krassowski:
Closes https://github.com/jupyterlab/jupyter-collaboration/issues/270

Adds `get_document(path, content_type, file_format)` method to the `YDocExtension`.

The returned object is a fork of the original document, thus any modifications are not propagated back.

### DIFF
--- a/jupyter_collaboration/app.py
+++ b/jupyter_collaboration/app.py
@@ -158,12 +158,7 @@ class YDocExtension(ExtensionApp):
             return None
 
         if isinstance(room, DocumentRoom):
-            update = room.ydoc.get_update()
-
-            fork_ydoc = Doc()
-            fork_ydoc.apply_update(update)
-
-            return YDOCS.get(content_type, YDOCS["file"])(fork_ydoc)
+            return room._document
 
         return None
 
--- a/jupyter_collaboration/app.py
+++ b/jupyter_collaboration/app.py
@@ -142,10 +142,10 @@ class YDocExtension(ExtensionApp):
         file_format: Literal["json", "text"],
         copy: bool = True,
     ) -> YBaseDoc | None:
-        """Get a read-only view of the shared model for the matching document.
+        """Get a view of the shared model for the matching document.
 
-        The returned shared model is a fork, meaning that any changes made to it
-        will not be propagated to the shared model used by the application.
+        If `copy=True`, the returned shared model is a fork, meaning that any changes
+         made to it will not be propagated to the shared model used by the application.
         """
         file_id_manager = self.serverapp.web_app.settings["file_id_manager"]
         file_id = file_id_manager.index(path)
--- a/jupyter_collaboration/app.py
+++ b/jupyter_collaboration/app.py
@@ -3,16 +3,26 @@
 from __future__ import annotations
 
 import asyncio
+from typing import Literal
 
 from jupyter_server.extension.application import ExtensionApp
+from jupyter_ydoc import ydocs as YDOCS
+from jupyter_ydoc.ybasedoc import YBaseDoc
+from pycrdt import Doc
 from pycrdt_websocket.ystore import BaseYStore
 from traitlets import Bool, Float, Type
 
 from .handlers import DocSessionHandler, YDocWebSocketHandler
 from .loaders import FileLoaderMapping
+from .rooms import DocumentRoom
 from .stores import SQLiteYStore
-from .utils import AWARENESS_EVENTS_SCHEMA_PATH, EVENTS_SCHEMA_PATH
-from .websocketserver import JupyterWebsocketServer
+from .utils import (
+    AWARENESS_EVENTS_SCHEMA_PATH,
+    EVENTS_SCHEMA_PATH,
+    encode_file_path,
+    room_id_from_encoded_path,
+)
+from .websocketserver import JupyterWebsocketServer, RoomNotFound
 
 
 class YDocExtension(ExtensionApp):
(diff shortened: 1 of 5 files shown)


jupyter_collaboration/app.py
```diff
@@ -112,6 +122,39 @@ def initialize_handlers(self):
...
+        except RoomNotFound:
+            return None
+
+        if isinstance(room, DocumentRoom):
+            update = room.ydoc.get_update()
+
+            fork_ydoc = Doc()
+            fork_ydoc.apply_update(update)
+
+            return YDOCS.get(content_type, YDOCS["file"])(fork_ydoc)
```
davidbrochart, line 154: So this will be a snapshot of the document at the time this method is called, but don't you want a live document, with all future updates applied?

krassowski: Good point. For my use case I am fine with having just a snapshot, but maybe having a proper _view_ which gets the future updates would make more sense for others?

krassowski: > I don't know if you would be fine with having a live fork.

Maybe.

I do not _strictly_ want a snapshot. I am indifferent to that, as long as it does not force me to wait for the lock/update/etc.

> In the latter case, I'm wondering why you still need a shared document, rather than just converting it to a plain Python object?

Converting a notebook to a plain Python object takes 4 times longer (https://github.com/jupyterlab/jupyter-collaboration/issues/270#issuecomment-2027549354). 30 ms on a notebook with 100 empty cells makes me worried that this would be unsuitable for the completions use case especially when notebooks have large visualisations in outputs etc. Often we will only want last/next ~10 cells so we would not be converting full notebook to Python ever.

krassowski: I do not follow. I am not in the frontend, I am in the server extension. (10 cells is an underestimate/figure of speech, it will be probably much more, the point is it will not be all 5000 cells in a notebook so long but a fraction, maybe 50 maybe 200)

davidbrochart: Ah ok, then you could do the same in the backend: get the shared document and convert any subset of it you'd like.

krassowski: This is exactly what I am doing in https://github.com/jupyterlab/jupyter-ai/pull/708 and plan to use this PR for. The question is do we want to return:
- a frozen read-only model copy
- a live read-only model
- a live read-write model

Again, for my use case this is really indifferent. From the point of view of API design I thought that starting with the smallest possible API surface (frozen model copy) and then expanding it in backward-compatible way (adding live updates, adding write access) seems like an easy decision (well maybe that is not fully backward compatible, but it would not be breaking for most use cases).

Currently this PR returns a frozen (but not read-only) copy of the model. I believe I can make it read-only by monkey-patching some methods in the Model instance after creating the copy so that they raise an exception when write is attempted.

What is the best next step/solution in your opinion?

krassowski: > but I don't understand the benefit of having a frozen copy of the model

There is not any - other than that it is a simple way of ensuring that other extensions do not start writing to the shared model via this new public API.

> I think `get_document()` should just return the shared model, not a copy of it.

I am happy with that.

davidbrochart: What do you think about adding a parameter `get_document(copy=True)`? In the future we could also add `live=True`, which would correspond to the use-case we have in the suggestion system.

krassowski: Sounds fine to me, done in: https://github.com/jupyterlab/jupyter-collaboration/pull/275/commits/3a8ba7ddcbd179b65618bd574c306f346ead57fa

jupyter_collaboration/app.py
```diff
@@ -112,6 +122,43 @@ def initialize_handlers(self):
...
+        *,
+        path: str,
+        content_type: Literal["notebook", "file"],
+        file_format: Literal["json", "text"],
+        copy: bool = True,
+    ) -> YBaseDoc | None:
+        """Get a read-only view of the shared model for the matching document.
+
+        The returned shared model is a fork, meaning that any changes made to it
+        will not be propagated to the shared model used by the application.
```
davidbrochart, line 136:
```suggestion
        """Get a view of the shared model for the matching document.

        If `copy=True`, the returned shared model is a fork, meaning that any changes
         made to it will not be propagated to the shared model used by the application.
```

jupyter_collaboration/app.py
```diff
@@ -112,6 +122,43 @@ def initialize_handlers(self):
...
+            return None
+
+        if isinstance(room, DocumentRoom):
+            if copy:
+                update = room.ydoc.get_update()
+
+                fork_ydoc = Doc()
+                fork_ydoc.apply_update(update)
+
+                return YDOCS.get(content_type, YDOCS["file"])(fork_ydoc)
```
davidbrochart, line 156:
```suggestion
                return YDOCS.get(content_type, YDOCS[content_type])(fork_ydoc)
```

davidbrochart (changes requested):

jupyter_collaboration/app.py
krassowski: This does not make much sense to me - the second argument is the fallback for when the given `content_type` is not found. If we are not worried about it maybe we should simplify to:

```suggestion
                return YDOCS[content_type](fork_ydoc)
```

davidbrochart: Oh sorry I didn't read carefully the whole line, please discard my suggestion.

krassowski:
@davidbrochart should we merge and iterate or do you have any further suggestions to address here first?

davidbrochart (approved):
Let's merge, thanks Mike!

Issue #252, opened 2024-03-15
Adopt AnyIO


davidbrochart:
### Problem

I'm seeing this error due to [aiosqlite](https://github.com/omnilib/aiosqlite), for instance in the [CI](https://github.com/jupyterlab/jupyter-collaboration/actions/runs/8297184676/job/22708202921?pr=251#step:7:65):
```
Traceback (most recent call last):
  File "C:\hostedtoolcache\windows\Python\3.11.8\x64\Lib\threading.py", line 1045, in _bootstrap_inner
    self.run()
  File "C:\hostedtoolcache\windows\Python\3.11.8\x64\Lib\site-packages\aiosqlite\core.py", line 120, in run
    future.get_loop().call_soon_threadsafe(set_exception, future, e)
  File "C:\hostedtoolcache\windows\Python\3.11.8\x64\Lib\asyncio\base_events.py", line 807, in call_soon_threadsafe
    self._check_closed()
  File "C:\hostedtoolcache\windows\Python\3.11.8\x64\Lib\asyncio\base_events.py", line 520, in _check_closed
    raise RuntimeError('Event loop is closed')
RuntimeError: Event loop is closed
```
I also saw this kind of errors in Jupyverse, where I'm switching to [sqlite-anyio](https://github.com/davidbrochart/sqlite-anyio) ([here](https://github.com/jupyter-server/jupyverse/pull/388)).

### Proposed Solution

In general, jupyter-collaboration could benefit from structured concurrency, just like pycrdt-websocket did.

dlqqq:
Found a possibly related upstream issue: https://github.com/omnilib/aiosqlite/issues/241

PR #278, opened 2024-04-01
Workaround shutdown issue with `aiosqlite` for now

krassowski:
Addresses issue discussed in https://github.com/jupyterlab/jupyter-collaboration/issues/252 to help bring the CI green again.

The Windows tests are still failing though - are these indicative or Windows-specific bugs?

### DIFF
--- a/tests/test_documents.py
+++ b/tests/test_documents.py
@@ -62,6 +62,7 @@ async def test_room_concurrent_initialization(
 
 
 async def test_room_sequential_opening(
+    jp_serverapp,
     rtc_create_file,
     rtc_connect_doc_client,
 ):
@@ -81,3 +82,7 @@ async def test_room_sequential_opening(
     assert dt < 1
     dt = await connect(file_format, file_type, file_path)
     assert dt < 1
+
+    # workaround for a shutdown issue of aiosqlite, see
+    # https://github.com/jupyterlab/jupyter-collaboration/issues/252
+    await jp_serverapp.web_app.settings["jupyter_collaboration"].stop_extension()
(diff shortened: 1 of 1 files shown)


davidbrochart:
From [the macOS failure](https://github.com/jupyterlab/jupyter-collaboration/actions/runs/8513264844/job/23316584924?pr=278) it looks like your workaround could also be applied to `test_room_concurrent_initialization`?
The [Windows failure](https://github.com/jupyterlab/jupyter-collaboration/actions/runs/8513264844/job/23316584095?pr=278) is caused by the `jupyter_server_fileid` [manager](https://github.com/jupyter-server/jupyter_server_fileid/blob/ffd314832343b2d0ebca62925d523125b7cf3ad7/jupyter_server_fileid/manager.py#L387-L392) which does some cleanup on garbage collection, which I don't think is a good practice.

krassowski:
Well, down to one failure:

image

I guess we can try to force `jupyter_server_fileid` to delete during the test too?

davidbrochart:
But the [failure](https://github.com/jupyterlab/jupyter-collaboration/actions/runs/8520156754/job/23335715911) means that your workaround didn't work, right?

krassowski:
Not quite sure. Could it have been an interaction with `jupyter_server_fileid`?  It is all passing now?

davidbrochart:
I think it's flaky, we should rerun tests to check.

krassowski:
I will close and reopen to see if we get all green again (of note `Tests / Test Minimum Versions (pull_request)` looks stuck right now).

davidbrochart (approved):
I reran tests 2 more times, your workaround seems to work fine.
Now it's `test_dirty` that seems to be flaky :smile:

Issue #269, opened 2024-03-28
Separate out the frontend and the backend


davidbrochart:
### Problem

[jupyter-collaboration](https://github.com/jupyterlab/jupyter-collaboration) is currently under the [jupyterlab](https://github.com/jupyterlab) organization. It includes the JupyterLab frontend and the [jupyter_server](https://github.com/jupyter-server/jupyter_server) backend. The backend part used to be [jupyter_server_ydoc](https://github.com/jupyter-server/jupyter_server_ydoc), under the [jupyter-server](https://github.com/jupyter-server) organization. I think it was a good choice to separate out the frontend and the backend, since some extensions may want to have access to the shared models without bringing the frontend stuff. Also it would allow [jupyverse](https://github.com/jupyter-server/jupyverse) to not depend (or depend less) on `jupyter-server`.

### Proposed Solution

We could split this repository in two, and have only the JupyterLab frontend stuff in a new repository `jupyterlab/jupyterlab-collaboration`, which would have the backend as a dependency.
The backend part would be in `jupyter-server/jupyter-collaboration`.
What do you think?

cc @krassowski

jtpio:
> We could split this repository in two, and have only the JupyterLab frontend stuff in a new repository `jupyterlab/jupyterlab-collaboration`, which would have the backend as a dependency.
> The backend part would be in `jupyter-server/jupyter-collaboration`.

Just to understand the issue correctly. Is the issue that the current `jupyter-collaboration` package includes the `@jupyter/collaboration-extension` by default, because it's included in the wheel by default as data files?

If so, another way would be to have a Python package for the backend logic, and another Python package to include both the backend and the JupyterLab extension? That way we can avoid splitting the repos again, and reduce maintenance overhead by keeping the packages in the same repo.

jtpio:
> The backend part used to be [jupyter_server_ydoc](https://github.com/jupyter-server/jupyter_server_ydoc), under the [jupyter-server](https://github.com/jupyter-server) organization

Are we able to find the issue / PR that initiated the move to a single `jupyter-collaboration` repo? To understand the original motivations for this.

davidbrochart:
I can't find an issue or a PR, but I guess it was for ease of releasing and synchronization between the frontend and the backend.

jtpio:
Some more context in https://github.com/jupyterlab/jupyterlab/issues/13527.

[...]

PR #279, opened 2024-04-08
Support server-side execution

davidbrochart:
This currently only works with [jupyverse](https://github.com/jupyter-server/jupyverse), not yet with [jupyter-server](https://github.com/jupyter-server/jupyter_server).
You can try it with:
```bash
pip install "jupyterlab>=4.2.0b0"
pip install "jupyverse[jupyterlab, auth]>=0.4.2"
jupyverse --set kernels.require_yjs=true --set jupyterlab.server_side_execution=true
```

Supersedes https://github.com/jupyterlab/jupyterlab/pull/15448.
Closes https://github.com/jupyterlab/jupyterlab/issues/2833.
Closes https://github.com/jupyterlab/jupyterlab/issues/12867.

https://github.com/jupyterlab/jupyterlab/assets/4711805/c65b5253-da77-4203-9223-b9097e6688c6


davidbrochart:
I don't understand the error in [Test JavaScript](https://github.com/jupyterlab/jupyter-collaboration/actions/runs/8611885039/job/23599919858?pr=279):
```
@jupyter/react-components/lib/index.js:3
export * from './wrap';
^^^^^^
SyntaxError: Unexpected token 'export'
```
@krassowski or @jtpio do you have an idea?

jupyter_collaboration/app.py
```diff
@@ -57,6 +57,14 @@ class YDocExtension(ExtensionApp):
         directory.""",
     )
 
+    server_side_execution = Bool(
```
jtpio, line 60: Maybe there could also be a small note about this flag in the documentation? https://jupyterlab-realtime-collaboration.readthedocs.io/en/latest/

So it's more visible and discoverable.

jtpio: And also mentioning it only works with `jupyverse` for now

davidbrochart:
Not sure what is happening with [Test Minimum Versions](https://github.com/jupyterlab/jupyter-collaboration/actions/runs/8617194987/job/23617749048?pr=279), any idea @blink1073 ?

krassowski:
We need similar change as in https://github.com/jupyterlab/jupyterlab/pull/16105

davidbrochart:
Thanks @krassowski !

packages/collaboration-extension/src/collaboration.ts
```diff
@@ -189,3 +191,63 @@ export const userEditorCursors: JupyterFrontEndPlugin<void> = {
...
+
+export const notebookCellExecutor: JupyterFrontEndPlugin<INotebookCellExecutor> =
+  {
+    id: '@jupyter/collaboration-extension:notebook-cell-executor',
+    description:
+      'Add notebook cell executor that uses REST API instead of kernel protocol over WebSocket.',
+    autoStart: true,
+    provides: INotebookCellExecutor,
+    activate: (app: JupyterFrontEnd): INotebookCellExecutor => {
+      if (PageConfig.getOption('serverSideExecution') === 'true') {
```
jtpio, line 203: Looking at the diff for this PR, this `serverSideExecution` page config option does not seem to be set?

davidbrochart: My understanding is that setting it [here](https://github.com/jupyterlab/jupyter-collaboration/blob/a0f46107375c8ba0fb19590fae13c1ce6c78c0b5/jupyter_collaboration/app.py#L60) magically makes it appear in the `PageConfig` as `serverSideExecution`.
I checked that it can be configured by Jupyverse's CLI:
```bash
jupyverse --set kernels.require_yjs=true --set jupyterlab.server_side_execution=true
```

jtpio: > My understanding is that setting it [here](https://github.com/jupyterlab/jupyter-collaboration/blob/a0f46107375c8ba0fb19590fae13c1ce6c78c0b5/jupyter_collaboration/app.py#L60) magically makes it appear in the `PageConfig` as `serverSideExecution`.


Isn't because it is set here in `jupyverse`?

https://github.com/jupyter-server/jupyverse/blob/085705a82c77718be428ddde67ea1e52af4b1e3a/plugins/jupyterlab/fps_jupyterlab/routes.py#L127

davidbrochart: Oh indeed, I forgot about this one. So no magic :smile: 
I'll make the change in the settings, thanks.

docs/source/configuration.md
```diff
@@ -29,3 +29,20 @@ jupyter lab --YDocExtension.document_cleanup_delay=100
...
+with the kernel is then delegated to the server, and cell outputs are populated in the notebook
+shared document. The frontend gets these outputs changes and shows them live. What this means is
+that the notebook state can be fully recovered even if the frontend disconnects, because cell
+outputs are not populated frontend-side but server-side.
+
+This feature is disabled by default, and can be enabled like so:
+```bash
+pip install "jupyterlab>=4.2.0b0"
+pip install "jupyverse[jupyterlab, auth]>=0.4.2"
+jupyverse --set kernels.require_yjs=true --set jupyterlab.server_side_execution=true
```
jtpio, line 47: Should the command also set the `YDocProvider.server_side_execution` trait? Or is that automatically picked up by `jupyverser` in some way?

jtpio:
@davidbrochart maybe having a small screencast in the top comment could help have a better idea of what the feature looks like? Or maybe the one in https://github.com/jupyterlab/jupyterlab/pull/15448#issuecomment-1830079178 is still relevant?

docs/source/configuration.md
davidbrochart: Jupyverse doesn't use traits nor the Python side of `jupyter-collaboration`, so passing `--set jupyterlab.server_side_execution=true` is all that is needed.

davidbrochart:
Yes, I added that screencast to the top comment :+1:

almereyda:
This looks marvellous, congratulations. One little question about the second part, when the notebook had been reopened: Is there any other visual indication, that there is a cell or notebook running on the server, other than the intermittent changes of the tab cross when new changes are pushed and saved right away?

My immediate intuition was to look for the play button, if it turned into some kind of pause button, but it wouldn't have changed.

Or maybe the Kernel is not actually *Idle*, as displayed, but in some kind of new state JupyterLab doesn't know about, yet?

ujjawal-khare:
@davidbrochart by any chance we can release this as a pre-release. It will be really helpful for our use case.

packages/collaboration-extension/src/collaboration.ts
```diff
@@ -189,3 +191,63 @@ export const userEditorCursors: JupyterFrontEndPlugin<void> = {
     });
   }
 };
+
+export const notebookCellExecutor: JupyterFrontEndPlugin<INotebookCellExecutor> =
```
krassowski, line 195: Something to consider: do we want this plugin in a separate package in the future so that when we split frontend and backend (https://github.com/jupyterlab/jupyter-collaboration/issues/269) we can still load this plugin but not the others? This question does not block merging this PR in any way.

docs/source/configuration.md
```diff
@@ -29,3 +29,20 @@ jupyter lab --YDocExtension.document_cleanup_delay=100
 # The YStore class to use for storing Y updates (default: JupyterSQLiteYStore).
 jupyter lab --YDocExtension.ystore_class=pycrdt_websocket.ystore.TempFileYStore
 ```
+
+There is an experimental feature that is currently only supported by the
+[Jupyverse](https://github.com/jupyter-server/jupyverse) server
+(not yet with [jupyter-server](https://github.com/jupyter-server/jupyter_server)):
```
krassowski, line 35: Is there an issue with proposal to implement it in jupyter-server to allow users to track the progress (or offer to contribute)?

Was implementing this in jupyter-server previously discussed in jupyter-server calls?

docs/source/configuration.md
```diff
@@ -29,3 +29,20 @@ jupyter lab --YDocExtension.document_cleanup_delay=100
...
+server-side execution. With this, running notebook code cells is not done in the frontend through
+the low-level kernel protocol over WebSocket API, but through a high-level REST API. Communication
+with the kernel is then delegated to the server, and cell outputs are populated in the notebook
+shared document. The frontend gets these outputs changes and shows them live. What this means is
+that the notebook state can be fully recovered even if the frontend disconnects, because cell
```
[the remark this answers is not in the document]
davidbrochart: OK to remove "fully" :smile:

docs/source/configuration.md
davidbrochart: There is https://github.com/jupyter-server/jupyter_server/issues/900, and I think @Zsailer is interested in implementing it.

packages/collaboration-extension/src/collaboration.ts
davidbrochart: But does it make sense to have server-side execution without collaboration? The former is based on the latter, so in my opinion they go together.

krassowski: It would be sufficient to have the collaborative drive, right?

docs/source/configuration.md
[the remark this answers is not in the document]
krassowski:
```suggestion
that the notebook state can be recovered even if the frontend disconnects, because cell
```

davidbrochart:
> @davidbrochart by any chance we can release this as a pre-release. It will be really helpful for our use case.

Yes that's the plan, give us a little bit more time :smile: 
Also, beware that this currently only works with [Jupyverse](https://github.com/jupyter-server/jupyverse).

krassowski (approved):
Looks good to me. I propose to merge and iterate (also in interest of moving forward with repo restructuring).

davidbrochart:
I'm having a last-minute thought. With the discussions around execution state, I'm wondering if having a REST API for cell execution is the best. Instead, we could for instance take advantage of the `"execution_state"` [property](https://github.com/jupyter-server/jupyter_ydoc/pull/197) of a cell to trigger execution. A client wishing to execute a cell would set this field to `"running"`, if currently in `"idle"`. This would make server-side execution transport-agnostic (not specific to a REST API). What do you think?

krassowski:
In jupyter-server we can (in principle) forbid execution for specific users as we have an authorization layer (https://jupyter-server.readthedocs.io/en/latest/operators/security.html#authorization). It is easy to wrap a REST API with authorization. Can this be also done easily with shared model modifications without increasing its complexity?

Another concern, possibly easier to address is how to execute a cell multiple times (well of course that could be a list of requests, and each request could be numbered by the client).

davidbrochart:
But since this is an experimental feature, only working in `jupyverse`, and under a flag, maybe it's okay to merge. We can still iterate in the future.

davidbrochart:
Now that we depend on NPM packages that are in pre-release, does it imply that we should also make a pre-release, say `v2.0.12b0`, or can it just be `v2.0.12`?

jtpio:
Since it adds a new feature, maybe it can be bumped to `2.1.0b0`?

davidbrochart:
OK to bump, be should it be a beta?

jtpio:
No strong opinions. If it doesn't change the default behavior of the `jupyter-collaboration` package it could even be a `2.1.0` release directly. Otherwise an alpha would be fine, followed by a final release if everything is ok.

krassowski:
I would say a beta, but can we also get https://github.com/jupyterlab/jupyter-collaboration/pull/275 and my upcoming PR which splits backend and frontend in?

davidbrochart:
OK, let's wait then, there is  no hurry.

rishabh-dream11:
@davidbrochart Just checking to see if there is consensus on the solution and can be used, I have a product requirement pending on it, hence asking for the same. 
We use jupyter-server and have not tried jupyverse.

imerelli:
Hi, is there any update regarding this feature, which I personally consider very important? When will it be integrated into the official release of Jupyter?

krassowski:
Good question. I think with https://github.com/jupyterlab/jupyter-collaboration/pull/307 and https://github.com/datalayer/jupyter-server-nbmodel it will become at least possible to use it via an extension

vinaypoduval19:
@krassowski How can I test this feature out? The above extension seems to not be available on pypi

krassowski:
@vinaypoduval19 I am impatient too, but the `jupyter-server-nbmodel` extension is in a third-party organization and contributors here do not have a control on when it gets released. I am grateful to datalayer for developing the extension and I'm trying to be helpful by leaving a review on https://github.com/datalayer/jupyter-server-nbmodel/pull/20 and opening https://github.com/datalayer/jupyter-server-nbmodel/pull/21 but this is as much as I can do.

b10902118:
> This currently only works with [jupyverse](https://github.com/jupyter-server/jupyverse), not yet with [jupyter-server](https://github.com/jupyter-server/jupyter_server). You can try it with:
> 
> ```shell
> pip install "jupyterlab>=4.2.0b0"
> pip install "jupyverse[jupyterlab, auth]>=0.4.2"
> jupyverse --set kernels.require_yjs=true --set jupyterlab.server_side_execution=true
> ```
> 
> Supersedes [jupyterlab/jupyterlab#15448](https://github.com/jupyterlab/jupyterlab/pull/15448). Closes [jupyterlab/jupyterlab#2833](https://github.com/jupyterlab/jupyterlab/issues/2833). Closes [jupyterlab/jupyterlab#12867](https://github.com/jupyterlab/jupyterlab/issues/12867).
> 
>  Peek.2023-11-28.16-25.mp4

This seems not working now. I tried the exact version == or >=, also with `jupyter-collaboration` installed but still not working.

fkurushin:
> > This currently only works with [jupyverse](https://github.com/jupyter-server/jupyverse), not yet with [jupyter-server](https://github.com/jupyter-server/jupyter_server). You can try it with:
> > ```shell
> > pip install "jupyterlab>=4.2.0b0"
> > pip install "jupyverse[jupyterlab, auth]>=0.4.2"
> > jupyverse --set kernels.require_yjs=true --set jupyterlab.server_side_execution=true
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Supersedes [jupyterlab/jupyterlab#15448](https://github.com/jupyterlab/jupyterlab/pull/15448). Closes [jupyterlab/jupyterlab#2833](https://github.com/jupyterlab/jupyterlab/issues/2833). Closes [jupyterlab/jupyterlab#12867](https://github.com/jupyterlab/jupyterlab/issues/12867).
> > Peek.2023-11-28.16-25.mp4
> 
> This seems not working now. I tried the exact version == or >=, also with `jupyter-collaboration` installed but still not working.


I also have a problem but with dependencies versions 
```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
pycrdt-websocket 0.15.1 requires pycrdt<0.11.0,>=0.10.3, but you have pycrdt 0.9.18 which is incompatible.
jupyter-ydoc 3.0.1 requires pycrdt<0.11.0,>=0.10.1, but you have pycrdt 0.9.18 which is incompatible.
Successfully installed pycrdt-0.9.18
```

davidbrochart:
You should start from a fresh environment.

banderlog:
It is not working even from the fresh environment

```
python3 -m venv venv
./venv/bin/pip install "jupyterlab>=4.2.0b0"
./venv/bin/pip install "jupyverse[jupyterlab, auth]>=0.4.2"
./venv/bin/jupyverse --set kernels.require_yjs=true --set jupyterlab.server_side_execution=true --host=192.168.1.10
```

davidbrochart:
Thanks for reporting the issue @banderlog. This should be fixed by the following PRs:
- https://github.com/jupyterlab/jupyterlab/pull/17469
- https://github.com/jupyter-server/jupyverse/pull/480

davidbrochart:
@banderlog It should be fixed in jupyverse v0.7.8.

davidbrochart:
It looks like your db is wrong, could you try removing `.jupyter_ystore.db`?

banderlog:
I do not have such file, as far as I know:

```bash
sudo updatedb
locate .jupyter_ystore.db
# nothing found
```

banderlog:
@davidbrochart yep, now it works as intended in the first video 👍

maurerle:
Would be really cool to have this also on JupyterHub, even on single user just to let the notebook stay alive.
What is missing there? after installing jupyter-collaboration I get an error in F12 with:
`wss://jupyter.example.com/user/xyz/api/collaboration/room/JupyterLab:globalAwareness`

And Notebooks do not load anymore.

So I guess that there is still some more things needed..?

Issue #283, opened 2024-04-16
No collaborator panel in development mode


davidbrochart:
## Description

After [setting up a development environment](https://jupyterlab-realtime-collaboration.readthedocs.io/en/latest/developer/contributing.html#setting-up-a-development-environment), JupyterLab doesn't show the collaborator panel. Collaborating on a notebook works fine though.
Am I missing something?

## Reproduce

```bash
pip install -e .
pip install -e projects/jupyter-collaboration-ui -e projects/jupyter-docprovider -e projects/jupyter-server-ydoc
jupyter labextension develop --overwrite projects/jupyter-collaboration-ui
jupyter labextension develop --overwrite projects/jupyter-docprovider
```

martinRenou:
Do you have a JavaScript error message in the console?

What does `jupyter labextension list` give you?

davidbrochart:
It seems that it depends on the installation order:
```console
$ pip install -e .
$ pip install -e projects/jupyter-collaboration-ui -e projects/jupyter-docprovider -e projects/jupyter-server-ydoc
$ pip install jupyterlab
$ jupyter labextension develop --overwrite projects/jupyter-collaboration-ui
$ jupyter labextension develop --overwrite projects/jupyter-docprovider
```
`jupyter-collaboration` is not installed:
```console
$ jupyter labextension list
JupyterLab v4.1.6
/home/david/micromamba/envs/jupyter-collaboration/share/jupyter/labextensions
        jupyterlab_pygments v0.3.0 enabled OK (python, jupyterlab_pygments)
        @jupyter/docprovider-extension v3.0.0-alpha.0 enabled OK


Disabled extensions:
    @jupyterlab/filebrowser-extension:defaultFileBrowser
    @jupyterlab/notebook-extension:cell-executor
```
After reinstalling it with `jupyter labextension develop --overwrite projects/jupyter-collaboration-ui`, now it works:
```console
$ jupyter labextension list
JupyterLab v4.1.6
/home/david/micromamba/envs/jupyter-collaboration/share/jupyter/labextensions
        jupyterlab_pygments v0.3.0 enabled OK (python, jupyterlab_pygments)
        @jupyter/collaboration-extension v3.0.0-alpha.0 enabled OK
        @jupyter/docprovider-extension v3.0.0-alpha.0 enabled OK (python, jupyter-docprovider)


Disabled extensions:
    @jupyterlab/filebrowser-extension:defaultFileBrowser
[...]

PR #280, opened 2024-04-12
Split up the backend, frontend, and the docprovider

krassowski:
Closes #269

New structure:
- packages: (npm)
  - `@jupyter/collaboration` (no changes)
  - `@jupyter/collaboration-extension`: filebrowser and server-side execution plugin extracted
  - `@jupyter/docprovider` (no changes)
  - `@jupyter/docprovider-extension` (new) home for filebrowser/drive and server-side execution plugin (for now)
- projects: (python, new)
  - `jupyter-server-ydoc`: jupyter-server extension
  - `jupyter-docprovider`: JupyterLab/Notebook 7+ extension (`@jupyter/docprovider-extension`)
  - `jupyter-collaboration-ui`: JupyterLab/Notebook 7+ extension (`@jupyter/collaboration-extension`)
  - `jupyter-collaboration`: empty package which installs all the three above

### Breaking changes

We should call this a new major version as now it is necessary to import from `jupyter_server_ydoc` rather than `jupyter_collaboration` (though I am not sure if anyone was importing anything as there was no public API before?).

PR #285, opened 2024-04-16
Rename the docprovider destination name to avoid conficts

brichet:
Fix the `docprovider` extension destination path, following https://github.com/jupyterlab/jupyter-collaboration/pull/280

Fixes https://github.com/jupyterlab/jupyter-collaboration/issues/283

We should probably add some tests (I don't even know how it passed)


krassowski (approved):
Thank you @brichet!

brichet:
The `Check Release` seems to be broken on main too.

davidbrochart:
Thanks a lot Nicolas!

PR #287, opened 2024-04-17
Add the list of opened shared documents to the user awareness

### LINKED ISSUE #284: List of documents opened by collaborators
The collaborators panel display the current shared document of a collaborator.

It could be interesting to display a list of all shared documents on which a collaborator is working.

### Problem

To find the current shared document, we use the `statedb -> layout-restorer -> main` of Jupyterlab, and display the document only if the widget is a `notebook` or an `editor`.

https://github.com/jupyterlab/jupyter-collaboration/blob/f263eb13871f9867c194947a937845c53d93f789/packages/collaboration-extension/src/collaboration.ts#L110-L119

The drawbacks of this implementation are:
- the widgets that can host shared documents are hard coded
- the shared document is tracked by the `layout-restorer` (which is not related to collaboration)
- the shared document must be opened in the main area (e.g. chat which could be opened in a side panel)

In addition we only track here the current shared document in the main area, but not other shared documents opened in main area.

### Proposed Solution

We could instead update a list of documents every time a user opens or closes a shared document, from the `YDrive`.

https://github.com/jupyterlab/jupyter-collaboration/blob/f263eb13871f9867c194947a937845c53d93f789/packages/docprovider/src/ydrive.ts#L129

The collaborators panel could open the list of documents when clicking on the user (accordion widget).
Perhaps the current one in the main area could be highlighted as the main document the collaborator is working on.

brichet:
Keep track of all opened documents in the collaborators panel.

Related to #284

The file list is an panel under the user in collaboration sidepanel. The appearance of the file list is similar to that of the filebrowser (thanks @martinRenou for the idea).
The document opened in the main area of the collaborator is highlighted with a dot in front (as active document), similar to the running kernel dot in the filebrowser.

Peek 2024-04-19 16-48

Some warnings/questions:
- it uses the HTML classes (and CSS rules) from the filebrowser. If the filebrowser extension is not loaded, it may not work properly (a remix application for instance). Don't know if we should duplicate the CSS rules.
- to retrieve the "active" document, we still use the `statedb -> layout-restorer -> main`, and look for `:RTC:` in the filename:
  - it works only if the document widget is tracked by the `LayoutRestorer`
  - it should be updated if `:RTC:` is removed from the filename


martinRenou (approved):
LGTM!

brichet:
@davidbrochart I have a "merging is blocked" warning  because I have no approval by reviewers with write access.

May I bypass it or do we want to wait for it ?

davidbrochart (approved):
Thanks @brichet !

Issue #245, opened 2024-03-06
[New Issue]: Seeing Two collaborative sessions warning occasionally when creating a new file


jzhang20133:
## Description

We are occasionally observing an error dialog when opening up a new file that says there are two collaborative sessions are created for this file.
![Screenshot 2024-02-26 at 11 42 20 AM (1)](https://github.com/jupyterlab/jupyter-collaboration/assets/4554746/af1e5ce3-d34f-4a39-aee7-9b83bb42181a)

## Reproduce

We haven't found a way to reproduce this reliably. It feels a bit flaky, though I have some suspicion on where this is coming from, mentioned below. Here is are the series of logs we need:

```
[I 2024-02-26 11:40:01.627 ServerApp] Request for Y document 'TestRTCWithFolder/Untitled.ipynb' with room ID: 9bc89a74-323a-411f-96b4-bf47260a909f
[I 2024-02-26 11:40:01.765 YDocExtension] Creating FileLoader for: TestRTCWithFolder/Untitled.ipynb
[E 2024-02-26 11:40:01.766 ServerApp] Uncaught exception GET /api/collaboration/room/json:notebook:9bc89a74-323a-411f-96b4-bf47260a909f?sessionId=76415e06-5e42-460d-9356-402da6f62da3 (127.0.0.1)
    HTTPServerRequest(protocol='http', host='localhost:8888', method='GET', uri='/api/collaboration/room/json:notebook:9bc89a74-323a-411f-96b4-bf47260a909f?sessionId=76415e06-5e42-460d-9356-402da6f62da3', version='HTTP/1.1', remote_ip='127.0.0.1')
    Traceback (most recent call last):
      File "/Users/jialinzhang/miniconda3/envs/jlab4-fresh/lib/python3.10/site-packages/tornado/web.py", line 1769, in _execute
        result = await result  # type: ignore
      File "/Users/jialinzhang/miniconda3/envs/jlab4-fresh/lib/python3.10/site-packages/jupyter_collaboration/handlers.py", line 105, in prepare
        await self._websocket_server.start_room(self.room)
      File "/Users/jialinzhang/miniconda3/envs/jlab4-fresh/lib/python3.10/site-packages/pycrdt_websocket/websocket_server.py", line 88, in start_room
        await self._task_group.start(room.start)
[...]

PR #289, opened 2024-04-18
handle exception when websocket server start room failed

jzhang20133:
In this PR, we handle exception when websocket server start room failed by cleaning up fileLoader and room and user won't run into the two collaboration pop up due to websocket server task group inactive failure. 
This PR also move authentication logic ahead of room initialization. 

Resolving: 
https://github.com/jupyterlab/jupyter-collaboration/issues/291
https://github.com/jupyterlab/jupyter-collaboration/issues/245

### DIFF
--- a/projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
+++ b/projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
@@ -11,6 +11,7 @@ from typing import Any
 
 from jupyter_server.auth import authorized
 from jupyter_server.base.handlers import APIHandler, JupyterHandler
+from jupyter_server.utils import ensure_async
 from jupyter_ydoc import ydocs as YDOCS
 from pycrdt_websocket.websocket_server import YRoom
 from pycrdt_websocket.ystore import BaseYStore
@@ -70,6 +71,7 @@ class YDocWebSocketHandler(WebSocketHandler, JupyterHandler):
         task.add_done_callback(self._background_tasks.discard)
 
     async def prepare(self):
+        await ensure_async(super().prepare())
         if not self._websocket_server.started.is_set():
             self.create_task(self._websocket_server.start())
             await self._websocket_server.started.wait()
@@ -110,12 +112,26 @@ class YDocWebSocketHandler(WebSocketHandler, JupyterHandler):
                     # it is a transient document (e.g. awareness)
                     self.room = TransientRoom(self._room_id, self.log)
 
-            await self._websocket_server.start_room(self.room)
-            self._websocket_server.add_room(self._room_id, self.room)
+            try:
+                await self._websocket_server.start_room(self.room)
+            except Exception as e:
+                self.log.error("Room %s failed to start on websocket server", self._room_id)
+                # Clean room
+                await self.room.stop()
+                self.log.info("Room %s deleted", self._room_id)
+                self._emit(LogLevel.INFO, "clean", "Room deleted.")
 
-        res = super().prepare()
-        if res is not None:
-            return await res
+                # Clean the file loader in file loader mapping if there are not any rooms using it
+                _, _, file_id = decode_file_path(self._room_id)
+                file = self._file_loaders[file_id]
+                if file.number_of_subscriptions == 0 or (
+                    file.number_of_subscriptions == 1 and self._room_id in file._subscriptions
+                ):
+                    self.log.info("Deleting file %s", file.path)
+                    await self._file_loaders.remove(file_id)
+                    self._emit(LogLevel.INFO, "clean", "file loader removed.")
+                raise e
+            self._websocket_server.add_room(self._room_id, self.room)
 
     def initialize(
         self,
@@ -134,6 +150,8 @@ class YDocWebSocketHandler(WebSocketHandler, JupyterHandler):
         self._document_save_delay = document_save_delay
         self._websocket_server = ywebsocket_server
         self._message_queue = asyncio.Queue()
+        self._room_id = ""
+        self.room = None
 
     @property
     def path(self):
@@ -227,7 +245,7 @@ class YDocWebSocketHandler(WebSocketHandler, JupyterHandler):
         try:
             self.write_message(message, binary=True)
         except Exception as e:
-            self.log.debug("Failed to write message", exc_info=e)
+            self.log.error("Failed to write message", exc_info=e)
 
     async def recv(self):
         """
--- a/projects/jupyter-server-ydoc/jupyter_server_ydoc/rooms.py
+++ b/projects/jupyter-server-ydoc/jupyter_server_ydoc/rooms.py
@@ -161,7 +161,10 @@ class DocumentRoom(YRoom):
 
         Cancels the save task and unsubscribes from the file.
         """
-        await super().stop()
+        try:
+            await super().stop()
+        except RuntimeError:
+            pass
         # TODO: Should we cancel or wait ?
         if self._saving_document:
             self._saving_document.cancel()
@@ -298,4 +301,13 @@ class TransientRoom(YRoom):
         try:
             await super()._broadcast_updates()
         except asyncio.CancelledError:
+            pass
+
+    async def stop(self) -> None:
+        """
+        Stop the room.
+        """
+        try:
+            await super().stop()
+        except RuntimeError:
             pass
--- a/tests/test_handlers.py
+++ b/tests/test_handlers.py
@@ -131,3 +131,87 @@ async def test_room_handler_doc_client_should_emit_awareness_event(
     assert collected_data[1]["action"] == "leave"
     assert collected_data[1]["roomid"] == "text:file:" + fim.get_id("test.txt")
     assert collected_data[1]["username"] is not None
+
+
+async def test_room_handler_doc_client_should_cleanup_room_file(
+    rtc_create_file, rtc_connect_doc_client, jp_serverapp
+):
+    path, _ = await rtc_create_file("test.txt", "test")
+
+    event = Event()
+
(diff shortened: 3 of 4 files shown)


projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
```diff
@@ -109,9 +109,26 @@ async def prepare(self):
...
 
-            await self._websocket_server.start_room(self.room)
-            self._websocket_server.add_room(self._room_id, self.room)
+                # Clean the file loader if there are not rooms using it
+                _, _, file_id = decode_file_path(self._room_id)
+                file = self._file_loaders[file_id]
+                self.log.info("Deleting file %s", file.path)
+                await self._file_loaders.remove(file_id)
+                self._emit(LogLevel.INFO, "clean", "Loader deleted.")
+                raise e
```
davidbrochart, line 131: Should the exception be raised, or just logged?

davidbrochart (commented):
Thanks @jzhang20133, could you please add a test that checks this logic?

projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
jzhang20133: it should be raised when task group is no active.

jzhang20133:
@davidbrochart  I have added a unit test to make sure two collaborative warning event is not thrown   and room and file is deleted when both websocket requests fails due to webserver task group is no longer active.

Zsailer:
Awesome work, @jzhang20133! 

It looks like PR also fixes #291 by moving the `prepare` `super` call earlier in this override 🚀

projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
```diff
@@ -70,52 +70,71 @@ def create_task(self, aw):
...
-                        file_type,
-                        file,
-                        self.event_logger,
-                        ystore,
-                        self.log,
-                        self._document_save_delay,
-                    )
-
+        res = super().prepare()
+        if isinstance(res, Awaitable):
```
Zsailer, line 74: is there ever a scenario where this isn't an awaitable? I guess `None`? 

Can we use `ensure_async` from jupyter server?

```python
from jupyter_server.utils import ensure_async

await ensure_async(super().prepare())
```

Zsailer:
@jzhang20133 thanks for this! 

Just a minor comment about the `prepare` method. Does everything need to be under the condition that `prepare` is an awaitable? 

Maybe a simpler approach is to call `res = await ensure_async(super().prepare())`.

jzhang20133:
Addressed comments. @davidbrochart and @Zsailer  would you like to review this PR again?

projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
```diff
@@ -110,12 +112,27 @@ async def prepare(self):
...
+                    await self._websocket_server.start_room(self.room)
+                except Exception as e:
+                    self.log.error("Room %s failed to start on websocket server", self._room_id)
+                    # Clean room
+                    self.room.stop()
+                    self.log.info("Room %s deleted", self._room_id)
+                    self._emit(LogLevel.INFO, "clean", "Room deleted.")
+
+                    # if websocket server failed, then room and file will be garbage collected and file
+                    # Clean the file loader in file loader mapping if there are not rooms using it
```
Zsailer, line 125:
```suggestion
                    # Clean the file loader in file loader mapping if there are not any rooms using it
```

projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
```diff
@@ -110,12 +112,27 @@ async def prepare(self):
...
+                try:
+                    await self._websocket_server.start_room(self.room)
+                except Exception as e:
+                    self.log.error("Room %s failed to start on websocket server", self._room_id)
+                    # Clean room
+                    self.room.stop()
+                    self.log.info("Room %s deleted", self._room_id)
+                    self._emit(LogLevel.INFO, "clean", "Room deleted.")
+
+                    # if websocket server failed, then room and file will be garbage collected and file
```
Zsailer, line 124:
```suggestion
                    # if websocket server failed, then room and file will be garbage collected
```

tests/test_handlers.py
```diff
@@ -131,3 +131,87 @@ async def my_listener(logger: EventLogger, schema_id: str, data: dict) -> None:
...
+    doc = YUnicode()
+    doc.observe(_on_document_change)
+
+    async with await rtc_connect_doc_client("text", "file", path) as ws, WebsocketProvider(
+        doc.ydoc, ws
+    ):
+        await event.wait()
+        await sleep(0.1)
+
+    # kill websocketserver to minik task group inactive failure
```
Zsailer, line 156:
```suggestion
    # kill websocketserver to mimic task group inactive failure
```

projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
jzhang20133: addressed

projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
jzhang20133: removed this sentence.

tests/test_handlers.py
jzhang20133: addressed

projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
```diff
@@ -292,13 +308,15 @@ def on_close(self) -> None:
...
         # stop serving this client
         self._message_queue.put_nowait(b"")
-        if isinstance(self.room, DocumentRoom) and self.room.clients == [self]:
-            # no client in this room after we disconnect
-            # keep the document for a while in case someone reconnects
-            self.log.info("Cleaning room: %s", self._room_id)
-            self.room.cleaner = asyncio.create_task(self._clean_room())
-        if self._room_id != "JupyterLab:globalAwareness":
-            self._emit_awareness_event(self.current_user.username, "leave")
+        if hasattr(self, "room"):
```
Zsailer, line 311: I'd rather we not use `hasattr` and implicit attributes here. 

Instead, let's initialize the `room` and `_room_id` attributes in the `initialize(...)` method, even if we set them to `None` initially: https://github.com/jupyterlab/jupyter-collaboration/blob/926c4d000b26cb677aca93cf3868f6a4c11f0e8b/projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py#L120-L136

I think that's the way to do it in tornado handlers.

projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
```diff
@@ -292,13 +308,15 @@ def on_close(self) -> None:
...
+            if isinstance(self.room, DocumentRoom) and self.room.clients == [self]:
+                # no client in this room after we disconnect
+                # keep the document for a while in case someone reconnects
+                self.log.info("Cleaning room: %s", self._room_id)
+                self.room.cleaner = asyncio.create_task(self._clean_room())
+        if hasattr(self, "_room_id"):
```
Zsailer, line 317: [Same comment here. ](https://github.com/jupyterlab/jupyter-collaboration/pull/289/files#r1583880746)

Zsailer (approved):

Zsailer:
Looks great! Thanks @jzhang20133!

projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
```diff
@@ -110,12 +112,26 @@ async def prepare(self):
                     # it is a transient document (e.g. awareness)
                     self.room = TransientRoom(self._room_id, self.log)
 
-            await self._websocket_server.start_room(self.room)
-            self._websocket_server.add_room(self._room_id, self.room)
+            try:
+                await self._websocket_server.start_room(self.room)
+            except Exception as e:
```
davidbrochart, line 117: Maybe I'm missing something, but didn't we introduced the YRoom [exception handler](https://github.com/jupyter-server/pycrdt-websocket/blob/4e610460f425ef12b8e674df0304fced789aef33/pycrdt_websocket/yroom.py#L224) just for that purpose? It seems we are bypassing all the work we did in `pycrdt-websocket` and instead directly handling the exception here.

projects/jupyter-server-ydoc/jupyter_server_ydoc/rooms.py
```diff
@@ -161,7 +161,10 @@ async def stop(self) -> None:
 
         Cancels the save task and unsubscribes from the file.
         """
-        await super().stop()
+        try:
+            await super().stop()
+        except RuntimeError:
+            pass
```
davidbrochart, line 167: This should be done in `pycrdt-websocket` using the exception handler.

projects/jupyter-server-ydoc/jupyter_server_ydoc/rooms.py
```diff
@@ -299,3 +302,12 @@ async def _broadcast_updates(self):
...
             pass
+
+    async def stop(self) -> None:
+        """
+        Stop the room.
+        """
+        try:
+            await super().stop()
+        except RuntimeError:
+            pass
```
davidbrochart, line 313: This should be done in `pycrdt-websocket` using the exception handler.

davidbrochart (commented):
I think a follow-up to this PR would be to not handle exceptions directly here but in `pycrdt-websocket` using the exception handler.

Issue #290, opened 2024-04-18
Documents won't open after a user's session expires and renews. 


Zsailer:
I ran into an issue where the JupyterLab becomes completely blocked from opening _any_ documents when a user's cookie expires. Even after being re-authenticated and refreshing the page, I am unable to open any documents. The only thing that unblocks this scenario is to restart the server. 

To help debug the issue a bit, I'll give a summary of what I saw in the logs (sorry for the ugly JSON, it's what I have running 😅 ) with timestamps.   

The user cookie expires.
```
{
    "message": "Clearing invalid/expired login cookie jupyter-session-token",
    "levelname": "WARNING",
    "name": "ServerApp",
    "asctime": "2024-04-12 17:41:34,303",
    "module": "identity",
    "funcName": "_get_user",
    "pathname": "/opt/conda/envs/user-env/lib/python3.9/site-packages/jupyter_server/auth/identity.py",
    "lineno": 259
}
```
Due to expiration, the JupyterLab lost connection to the server and I think this caused the document to self clean.
```
{
    "message": "Cleaning room: json:notebook:3312ce81-fed7-44c4-a227-a21f002ad2a6",
    "levelname": "INFO",
    "name": "ServerApp",
    "asctime": "2024-04-12 17:39:07,037",
    "module": "handlers",
    "funcName": "on_close",
    "pathname": "/opt/conda/envs/user-env/lib/python3.9/site-packages/jupyter_collaboration/handlers.py",
    "lineno": 297
}
{
    "message": "Cleaning room: json:notebook:3312ce81-fed7-44c4-a227-a21f002ad2a6",
    "levelname": "INFO",
    "name": "ServerApp",
    "asctime": "2024-04-12 17:43:15,180",
    "module": "handlers",
    "funcName": "on_close",
    "pathname": "/opt/conda/envs/user-env/lib/python3.9/site-packages/jupyter_collaboration/handlers.py",
    "lineno": 297
}
```
The client is still trying to poll the server using an expired token, so a bunch of 403s appear. I'll include the first one here
```
{
    "message": "403 GET /api/collaboration/room/json:notebook:3312ce81-fed7-44c4-a227-a21f002ad2a6?sessionId=3dd37d38-6ec5-4c10-bb63-4a75d074e05b 5.46ms referer=None",
    "levelname": "WARNING",
    "name": "ServerApp",
    "asctime": "2024-04-12 17:43:17,645",
    "module": "log",
[...]

PR #294, opened 2024-04-24
Start Y Websocket Server earlier and make it resilient to crashes

Zsailer:
This prevents the server locking we see in #290

This PR makes two critical changes:

1. Instead of starting the y-websocket server when the first YWebsocketHandler request is made, this PR starts the server in the startup sequence of the ExtensionApp. 
2. Handle exceptions in the ywebsocket server using the exception logging offered in https://github.com/jupyter-server/pycrdt-websocket/pull/31

This depends on this PR in Jupyter Server: https://github.com/jupyter-server/jupyter_server/pull/1417

### DIFF
--- a/projects/jupyter-server-ydoc/jupyter_server_ydoc/app.py
+++ b/projects/jupyter-server-ydoc/jupyter_server_ydoc/app.py
@@ -10,7 +10,8 @@ from jupyter_ydoc import ydocs as YDOCS
 from jupyter_ydoc.ybasedoc import YBaseDoc
 from pycrdt import Doc
 from pycrdt_websocket.ystore import BaseYStore
-from traitlets import Bool, Float, Type
+from pycrdt_websocket.websocket_server import exception_logger
+from traitlets import Bool, Float, Type, Set, Instance
 
 from .handlers import DocSessionHandler, YDocWebSocketHandler
 from .loaders import FileLoaderMapping
@@ -23,7 +24,6 @@ from .utils import (
     room_id_from_encoded_path,
 )
 from .websocketserver import JupyterWebsocketServer, RoomNotFound
-
 
 class YDocExtension(ExtensionApp):
     name = "jupyter_server_ydoc"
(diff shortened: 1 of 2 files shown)


projects/jupyter-server-ydoc/jupyter_server_ydoc/app.py
```diff
@@ -134,6 +138,33 @@ def initialize_handlers(self):
...
+        """
+        self.log.info("Starting the Collaborative Document Server.")
+
+        def _restart_or_teardown_yserver(_):
+            """If the y websocket server task stopped due to an exception, restart it.
+            
+            If the y-websocket server was cancelled on purpose, tear it down.
+            """
+            self.ywebsocket_server._started = None
+            self.ywebsocket_server._starting = False
```
davidbrochart, line 152: This attribute doesn't exist anymore after https://github.com/jupyter-server/pycrdt-websocket/pull/24 (which was not released yet, but it will be soon).

davidbrochart (changes requested):
Thanks @Zsailer.
I think that it would be better to have an `auto_restart` parameter passed to `pycrdt-websocket`'s `WebsocketServer`, that would essentially do the same as here (restart on exception). That way, it would benefit to all users of the WebSocket server, not just `jupyter-collaboration`. What do you think?
I can open a PR there if you want.

Zsailer:
> I think that it would be better to have an auto_restart parameter passed to pycrdt-websocket's WebsocketServer

Sure! That sounds great. I'm happy to make the PR over there as well. Just let me know.

davidbrochart:
I opened https://github.com/jupyter-server/pycrdt-websocket/pull/31.

projects/jupyter-server-ydoc/jupyter_server_ydoc/app.py
davidbrochart: FYI it was released in `pycrdt-websocket` v0.13.0.

Zsailer:
This depends on a release of pycrdt-websocket (hence why the tests are failing).

Zsailer:
Closing in favor of #295

PR #295, opened 2024-04-29
Log (instead of raise) exceptions when running as a server extension

Zsailer:
This supersedes some of #294, leveraging changes in pycrdt-websocket v0.13.1. 

In #290, we see cases where the y-websocket server becomes unusable anytime an exception is raised within the y-websocket server's task groups. We made changes in pycrdt-websocket so that these exceptions could optionally be logged instead of raised, thereby keeping the server running. 

This PR leverages those changes so we can ensure that there serve is never blocked.

We need to bump the pycrdt-websocket minimum version to get these changes.


projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
```diff
@@ -105,6 +106,26 @@ async def prepare(self):
...
+                            room_id = self.room.room_id
+                        log.error(
+                            f"Document Room Exception, (room_id={room_id}: ",
+                            exc_info=exception,
+                        )
+                        return True
+
+                    # Logging exceptions, instead of raising them here to ensure
+                    # that the y-rooms stay alive even after an exception is seen.
+                    self.room.exception_handler = exception_logger
```
jzhang20133, line 127: why not update DocumentRoom's initialization method?

projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
```diff
@@ -105,6 +106,26 @@ async def prepare(self):
...
 
+                    def exception_logger(exception: Exception, log: Logger) -> bool:
+                        """A function that catches any exceptions raised in the websocket
+                        server and logs them.
+
+                        The protects the y-room's task group from cancelling
+                        anytime an exception is raised.
+                        """
+                        room_id = "unknown"
+                        if self.room.room_id:
```
jzhang20133, line 117: `self.room.room_id` can be `self._room_id`

jzhang20133 (approved):

projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
Zsailer: 👍

jzhang20133 (approved):

projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
```diff
@@ -81,6 +82,21 @@ async def prepare(self):
...
+                    server and logs them.
+
+                    The protects the y-room's task group from cancelling
+                    anytime an exception is raised.
+                    """
+                    log.error(
+                        f"Document Room Exception, (room_id={self._room_id or 'unknown'}): ",
+                        exc_info=exception,
+                    )
+                    return True
```
davidbrochart, line 98: This is almost the same function as in `websocketserver.py`, except for the logging message. I think this should be factored out, with an extra-argument for the message.

davidbrochart (approved):
Thanks @Zsailer, some comments but this should not block from merging.

projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
Zsailer: For sure. Since this is an internal definition right now, let's factor this out in a later PR. I want to proceed with merging, releasing, and testing.

PR #299, opened 2024-05-02
adding await ystore start method in prepare method

jzhang20133:
Together with https://github.com/jupyter-server/pycrdt-websocket/pull/40, resolving https://github.com/jupyter-server/pycrdt-websocket/issues/41


Zsailer:
Test failures appear to be related to: https://github.com/jupyter-server/jupyter_server_fileid/issues/75

jupyter_collaboration/handlers.py
```diff
@@ -111,6 +111,7 @@ def exception_logger(exception: Exception, log: Logger) -> bool:
                     file = self._file_loaders[file_id]
                     updates_file_path = f".{file_type}:{file_id}.y"
                     ystore = self._ystore_class(path=updates_file_path, log=self.log)
+                    await ystore.start()
```
davidbrochart, line 114: In the general case, starting a service can be blocking. Actually, after https://github.com/jupyter-server/pycrdt-websocket/pull/42 starting a YStore will be.
```suggestion
                    self.create_task(ystore.start())
                    await ystore.started.wait()
```

davidbrochart (changes requested):

jupyter_collaboration/handlers.py
davidbrochart: Actually, starting the YStore should not be done here but [in the DocumentRoom](https://github.com/jupyterlab/jupyter-collaboration/blob/1f367f58a1090ca27e920352b376edf03375b6c7/jupyter_collaboration/rooms.py#L102), since it is optional.
Also, you might want to stop the YStore when stopping the YRoom.

jzhang20133: I think adding await in the DocumentRoom is a good idea and it also help prevent starting the room at two places which could cause race condition as well. I will make that change.

Zsailer (approved):

jupyter_collaboration/rooms.py
```diff
@@ -100,6 +100,7 @@ async def initialize(self) -> None:
             # try to apply Y updates from the YStore for this document
             read_from_source = True
             if self.ystore is not None:
+                await self.ystore.started.wait()
```
davidbrochart, line 103: @jzhang20133 You didn't apply my [suggestion](https://github.com/jupyterlab/jupyter-collaboration/pull/299#discussion_r1588855135).

davidbrochart:
@Zsailer I don't think we should merge when there are requested changes that were not made, in my opinion. We should at least justify that these changes are not needed, but the best would be to wait for the reviewer's answer.
I understand that you want to iterate quickly, but reviewing is also taking a lot of my time, so it should be taken into consideration.

jupyter_collaboration/rooms.py
jzhang20133: @davidbrochart Following your suggestion, we moved the logic here in document room. By not applying your suggestion, are you referring to this line`self.create_task(ystore.start())`?  I did not add ystore.start logic here because ystore.start() is also called in yroom.broadcast_update() method. And if ystore has been started and then it is started again, [an exception could be thrown](https://github.com/jupyter-server/pycrdt-websocket/blob/main/pycrdt_websocket/ystore.py#L354).   Instead of triggering starting of ystore at multiple places and exception handling at multiple places, we choose to simply wait for the ystore to be started, which is a great idea that you have suggested.

Zsailer:
> I don't think we should merge when there are requested changes that were not made

This was my mistake, @davidbrochart, an oversight in my follow-up review. I'm so so sorry. 

I thought @jzhang20133 applied your suggestions after moving that chunk of code from the handlers to the document room. I now see that she dropped one of the lines. This one is on me for not flagging this during the review. I will take this as a note to slow down. 

> reviewing is also taking a lot of my time, so it should be taken into consideration.

:heart: absolutely, @davidbrochart. And we cannot express enough how much we appreciate you for this. 

Please know that this wasn't an attempt to bypass or ignore your review. It was an oversight on my part, and I apologize for not giving your time+review the proper attention ❤️

Zsailer:
> I understand that you want to iterate quickly

Yes, you're right the call this out. It's not a good excuse, but helpful to be transparent here. This issue was blocking us from putting on a "real" system with real people testing it, so we're trying to get these libraries in a useable state ASAP (they are pretty brittle today when running in a cloud environment where RTC is most useful). In the long run, I hope this brings helpful and powerful feedback for the project.

jupyter_collaboration/rooms.py
Zsailer: Thanks, @jzhang20133. I'll let David respond to this when he's back in office next week and we can create a follow-up PR if needed.

@davidbrochart's point below stands. If we do something different from his suggestions, we should include an explanation and wait for his response before merging. 

This one is on me folks. I missed this difference in the change before approving the PR. Let's learn from this and make sure we're properly patient going forward.

jzhang20133: @Zsailer @davidbrochart I will make sure I add explanation if change is made differently from suggested and wait for @davidbrochart 's inputs and approval on the difference in the future PRs.  Thank you both for your patience and help on reviewing PRs! Really appreciated!!!

jzhang20133: @davidbrochart  For this specific change, please feel free to let us know your thought here about not adding this line `self.create_task(ystore.start())`. I can address accordingly in another PR.

davidbrochart:
No worries Zach and Jialing, I hope my message was not too dry.
Be sure that I really appreciate the efforts you're putting into making Jupyter collaboration more robust, and keep up the great work!

jupyter_collaboration/rooms.py
jzhang20133: yes, it just wait for it to be started by yroom.broadcast_update() method. If yroom is started twice, [an exception could be thrown](https://github.com/jupyter-server/pycrdt-websocket/blob/main/pycrdt_websocket/ystore.py#L354) and the exception could kill task groups. We need to add exception handling for ystore.start() if it get started in multiple places.

PR #304, opened 2024-05-07
Fix writing settings in page_config

fcollonval:
The current code does not write the extension settings in the page config as `page_config_data` already exists.


davidbrochart (approved):
Thanks @fcollonval.
Failures are unrelated, I opened #305.

Issue #315, opened 2024-05-17
Time-to-live support not working anymore?


asteppke:
For notebooks that create a lot of output the `.jupyter_ystore.db` can get rather large and unfortunately overflows our users' quota easily.

In this example here

````python
import time
for i in range(100_000_000):
    print(f"{i}, ", end="")
    time.sleep(0.05)
````

the size of the notebook file after a runtime of a few minutes grows to 1.5 MB. On the other hand the corresponding `.jupyter_ystore.db` grows to 580 MB. 

I have read parts of the discussions around the database and I have a rough understanding of the the complications that makes solving this quite challenging. For now the time-to-live option seems like a suitable workaround to limit the growth to some extend. At the moment this does not seem to work (anymore?) though. 

When starting a new session with

````
jupyter lab --SQLiteYStore.document_ttl=600
````

I only receive the following error messages:

````
[I 2024-05-17 16:08:01.659 ServerApp] Creating new notebook in
[I 2024-05-17 16:08:01.722 ServerApp] Request for Y document 'Untitled10.ipynb' with room ID: 780564de-e0da-492a-9d14-af545441c896
[I 2024-05-17 16:08:01.913 YDocExtension] Creating FileLoader for: Untitled10.ipynb
[I 2024-05-17 16:08:01.914 YDocExtension] Watching file: Untitled10.ipynb
[I 2024-05-17 16:08:01.915 ServerApp] Initializing room json:notebook:780564de-e0da-492a-9d14-af545441c896
[I 2024-05-17 16:08:01.935 ServerApp] Content in room json:notebook:780564de-e0da-492a-9d14-af545441c896 loaded from file Untitled10.ipynb
[E 2024-05-17 16:08:01.937 ServerApp] Error initializing: Untitled10.ipynb
    TypeError("'>' not supported between instances of 'int' and 'DeferredConfigString'")
    Traceback (most recent call last):
      File "C:\tools\miniconda3\envs\data\Lib\site-packages\jupyter_collaboration\handlers.py", line 233, in open
        await self.room.initialize()
      File "C:\tools\miniconda3\envs\data\Lib\site-packages\jupyter_collaboration\rooms.py", line 151, in initialize
        await self.ystore.encode_state_as_update(self.ydoc)
      File "C:\tools\miniconda3\envs\data\Lib\site-packages\pycrdt_websocket\ystore.py", line 145, in encode_state_as_update
        await self.write(update)
[...]

PR #318, opened 2024-06-18
Workaround for later initialization, recovers document_ttl

asteppke:
This is a short fix for https://github.com/jupyterlab/jupyter-collaboration/issues/315. 

In the future this can be updated when the changed traitlets infrastructure is fully taken into account but for now this restores the `document_ttl` functionality and should have no negative side effects.

### DIFF
--- a/projects/jupyter-server-ydoc/jupyter_server_ydoc/stores.py
+++ b/projects/jupyter-server-ydoc/jupyter_server_ydoc/stores.py
@@ -30,3 +30,9 @@ class SQLiteYStore(LoggingConfigurable, _SQLiteYStore, metaclass=SQLiteYStoreMet
         help="""The document time-to-live in seconds. Defaults to None (document history is never
         cleared).""",
     )
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # Class is instantiated later, so we need to set the document_ttl here
+        self.document_ttl = int(self.document_ttl) if self.document_ttl is not None else None
+      
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,7 +3,7 @@ on:
   push:
     branches: ["main"]
   pull_request:
-
+  workflow_dispatch:
   schedule:
     - cron: "0 8 * * *"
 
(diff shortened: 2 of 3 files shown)


davidbrochart:
Thanks @asteppke.

> In the future this can be updated when the changed traitlets infrastructure is fully taken into account

What do you mean by "changed traitlets infrastructure"?

projects/jupyter-server-ydoc/jupyter_server_ydoc/stores.py
```diff
@@ -30,3 +30,9 @@ class SQLiteYStore(LoggingConfigurable, _SQLiteYStore, metaclass=SQLiteYStoreMet
         help="""The document time-to-live in seconds. Defaults to None (document history is never
         cleared).""",
     )
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # Class is instantiated later, so we need to set the document_ttl here
+        self.document_ttl = int(self.document_ttl) if self.document_ttl is not None else None
```
davidbrochart, line 37: I don't understand what this changes. Could you write a test that shows that this fixes something?

asteppke:
> Thanks @asteppke.
> 
> > In the future this can be updated when the changed traitlets infrastructure is fully taken into account
> 
> What do you mean by "changed traitlets infrastructure"?

I did not bisect this in detail but when you implemented the `document_ttl` functionality in the past I think it worked as intended. Instead of going through the normal traitlets initialization code the attributes are set directly here in jupyter_collaboration (I guess because of late initialization). The traitlet guys mention in passing in their documentation that since 5.0 there is a deferral happening of the conversion from command line arguments to traits. So therefore jupyter_collaboration is now setting the 'still-unconverted-strings' as attributes instead of traits. 

So this adding the missing conversion step. I can see if this can be shown also within the a test in the test suite.

projects/jupyter-server-ydoc/jupyter_server_ydoc/stores.py
asteppke: I added a test that checks whether the store has the correct parameter from the command line arguments. 

Without the changeset above the store writing fails immediately if document_ttl is set.

asteppke:
I added the corresponding tests for the document_ttl. Other unrelated tests fail though already without these commits.

PR #327, opened 2024-07-18
Move fixtures to `jupyter_server_ydoc.pytest_plugin`

### LINKED ISSUE #326: Extract fixtures to a pytest plugin
### Problem

For testing the collaboration mode and requires quite a specific setup to get a document room open. Since a number of extensions exists building on the shared model capabilities, it would be perfect if those could write pytest tests without duplicating the logic which is subject to change.

A few examples where python tests would be useful but are presently missing:
- https://github.com/jupyterlab/jupyter-chat/blob/main/packages/jupyterlab-collaborative-chat/jupyterlab_collaborative_chat/tests/test_handlers.py
- https://github.com/jupytercad/JupyterCAD

### Proposed Solution

Move some fixtures from `tests/conftest.py` to `projects/jupyter-server-ydoc/jupyter_server_ydoc/pytest_plugin.py`.

### Additional context

Other jupyter-server extensions (including `jupyterlab`, `jupyterlab_server`, `jupyter_server_fileid` and `jupyter_server` itself) ship with a set of fixtures in `pytest_plugin` module, e.g.:
- https://github.com/jupyter-server/jupyter_server_fileid/blob/main/jupyter_server_fileid/pytest_plugin.py
- https://github.com/jupyterlab/jupyterlab/blob/main/jupyterlab/pytest_plugin.py
- https://github.com/jupyterlab/jupyterlab_server/blob/main/jupyterlab_server/pytest_plugin.py

krassowski:
Fixes #326


krassowski:
All green! I would appreciate a review because I would like to open another PR on top of these reshuffle and release this in another beta :)

brichet (approved):
LGTM, thanks @krassowski

Issue #330, opened 2024-07-18
Frontend does not work unless jupyterlab-collaboration-ui is installed


krassowski:
## Description

Installing [jupyter-server-ydoc](https://github.com/jupyterlab/jupyter-collaboration/tree/main/projects/jupyter-server-ydoc) and [jupyter-docprovider](https://github.com/jupyterlab/jupyter-collaboration/tree/main/projects/jupyter-docprovider) in the latest beta version breaks jupyterlab. Installing jupyterlab-collaboration-ui fixes it:

![image](https://github.com/user-attachments/assets/1ce5affc-2f87-47bf-a504-a8c6b1ab7c33)

krassowski:
It looks like this is due to https://github.com/jupyterlab/jupyter-collaboration/pull/287 which introduced a bad dependency  on `@jupyter/collaboration` in `docprovider`; docprovider cannot depend on the collaboration UI. CC @brichet.

brichet:
@krassowski the only need for this dependency is for this [token](https://github.com/jupyterlab/jupyter-collaboration/blob/bcc36e1a451aaa5dbfbe020db72070149bd934bd/packages/collaboration/src/tokens.ts#L23).
Do you think that it make sense to move it to another package ? 

I'm not clear about the packaging of this repo. For instance the package `collaboration` depends on `docprovider` but seems to not use it: https://github.com/jupyterlab/jupyter-collaboration/blob/bcc36e1a451aaa5dbfbe020db72070149bd934bd/packages/collaboration/package.json#L44

And also I can reproduce the other errors you shared, about some missing dependencies of `@jupyterlab` packages....

krassowski:
Thank you for looking into it!

> Do you think that it make sense to move it to another package ?

Yes, let's move it.

> I'm not clear about the packaging of this repo. For instance the package collaboration depends on docprovider but doesn't seem to use it:

It was there before the refactor too. I guess we could remove it.

> And also I can reproduce the other errors you shared, about some missing dependencies of @jupyterlab packages....

By default you would only see the first two errors (error and a warning). If https://github.com/datalayer/jupyter-server-nbmodel is installed the additional error will also show up, because this issue also prevents it from starting.

brichet:
Thanks @krassowski, I'm opening a PR to fix it, moving the `IGlobalAwareness` token to the *docprovider* package.

PR #331, opened 2024-07-19
Remove @jupyter/collaboration dependency from @jupyter/docprovider

brichet:
Fixes https://github.com/jupyterlab/jupyter-collaboration/issues/330

This PR move the `IAwareness` token from *@jupyter/collaboration* package to *@jupyter/docpovider*, to fix a dependency issue.
It also remove a leftover dependency in *@jupyter/collaboration*.


krassowski (approved):
Thank you @brichet! Tested locally and can confirm that this fixes the issue!

PR #338, opened 2024-08-19
Implementing Collaborative Timeline Slider with Undo/Redo Functionality

Meriem-BenIsmail:
### Description:

This pull request introduces a comprehensive timeline and undo/redo functionality to the Jupyter Collaboration project. The primary goal of this feature is to enhance the collaborative experience by allowing users to track changes, restore previous versions of documents, and seamlessly undo or redo actions.

### Key Features:

1. **Timeline Slider Component:**
    - Added a component (`TimelineSliderComponent`) that provides an interactive slider to navigate through the history of document changes.
    - The slider fetches and displays timestamps of changes, allowing users to select specific versions of the document and to restore them if needed.
2. **Document Forking and Restoration:**
    - Implemented functions to fork the current state of the document and create new versions.
    - Users can restore the document to any previous state using the `restore` mode.
3. **Undo/Redo Management:**
    - Integrated [UndoManager](https://github.com/jupyter-server/pycrdt/blob/main/python/pycrdt/_undo.py#L18) to handle undo and redo operations for document edits.
    - Users can perform multiple undo/redo actions to revert or reapply changes.
4. **API Enhancements:**
    - Extended the existing API to support timeline fetching and document restoration.
    - Introduced new endpoints for creating document forks and managing undo/redo operations.
5. **Notifications:**
    - Implemented user notifications to provide feedback on actions such as successful document restoration or errors.

### Usage:

1. **Accessing the Timeline:**
    - Click on the history icon in the status bar to open the timeline slider.
    - Use the slider to navigate through different versions of the document.
2. **Restoring a Version:**
    - Select a desired timestamp from the slider and click the "Restore version" button to revert the document to the selected state.
    - A notification will confirm the successful restoration or display an error if the operation fails.
3. **Undo/Redo Actions:**
    - Use the timeline slider to perform undo/redo operations by moving back and forth between timestamps.

Note: The timeline is tested  with various document types, including notebooks and files in different directory structures and also [YJCad](https://github.com/jupytercad/JupyterCAD/blob/main/python/jupytercad_core/jupytercad_core/jcad_ydoc.py) files of [JupyterCAD](https://github.com/jupytercad/JupyterCAD).

Here is a demo of the functionality
timeline-rtc-feature


Meriem-BenIsmail:
> Thanks for working on this @Meriem-BenIsmail ! I just tried your branch, and I'm not able to go back in time:
> 
>  Peek.2024-08-20.14-53.mp4 
> Is there something I'm doing wrong?

That's weird I tried to recreate your vid and it works for me 
prb

Thanks for the feedback though, I will have a look at what might be causing the problem.
Note: when the timeline is opened and we start using the slider the document displayed is a fork so any changes made after using the slider aren't saved.

packages/docprovider/src/ydrive.ts
```diff
@@ -57,10 +60,57 @@ export class YDrive extends Drive implements ICollaborativeDrive {
...
+
+  setTimelineWidget(widget: TimelineWidget) {
+    this._timelineWidget = widget;
+  }
+
+  getTimelineWidget(): TimelineWidget | null {
+    return this._timelineWidget || null;
+  }
+
+  async updateTimelineForNotebook(notebookPath: string): Promise<void> {
```
krassowski, line 78: It looks like it supports not only notebook but any documents, right? Later below I propose to rewrite this procedure with side effects into a more generic function, let's call it `getProviderFor(path: string): Promise<WebSocketProvider>`

packages/docprovider/src/ydrive.ts
```diff
@@ -220,6 +270,29 @@ export class YDrive extends Drive implements ICollaborativeDrive {
...
+          options.contentType,
+          options.format
+        );
+      } else {
+        this.updateTimelineForNotebook(options.path);
+        this._timelineWidget.update();
+      }
+      const elt = document.getElementById('slider-status-bar');
+      if (elt && !this._timelineWidget.isAttached) {
+        Widget.attach(this._timelineWidget, elt);
```
krassowski, line 292: Conceptually, it is not a responsibility of the drive to know anything about widgets. Usually we would create a new "model" (like `timelineModel`) to interface between one object knowing about multiple things and the widgets, but here a simpler solution might be possible.

Practically, querying by ID is a code smell.

Here is a proposal:
- let's move instationation and attachement of `TimelineWidget` to the `@jupyter/docprovider-extension:statusBarTimeline` plugin (where it will be a breeze because you have the status bar element there widget reference)
- replace `updateTimelineForNotebook` method on `YDrive` with `getProviderForPath(path: string): Promise<IForkProvider>` where:
  ```ts
  interface IForkProvider {
    connectToFork: (arguments) => Promise<someType>;
    contentType: string;
    format: string;
  }
  ```
  and mark `WebSocketProvider` as implementing `IForkProvider` interface
- if absolutely needed add a signal on `YDrive` such as `modelCreated: Signal<YDrive, IModelCreatedArgs>` so that you can call `getProviderForPath()` immediately once this `_onCreate()` is called, where:
  ```ts
  interface IModelCreatedArgs {
     options: Contents.ISharedFactoryOptions;
     sharedModel: YDocument<DocumentChange>;
  }
  ```
   but maybe you could just use document tracker instead.

packages/docprovider/src/yprovider.ts
```diff
@@ -88,8 +95,18 @@ export class WebSocketProvider implements IDocumentProvider {
...
   }
+  get sharedModel(): YDocument<DocumentChange> {
+    return this._sharedModel;
+  }
+  setPath(path: string) {
+    this._path = path;
+  }
+  setSharedModel(sharedModel: YDocument<DocumentChange>) {
+    this._sharedModel = sharedModel;
+  }
```
krassowski, line 109:
```suggestion
```

Is this used anywhere? It would be a significant expansion to the public API, let's keep it small to make it managable to maintain.

projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
```diff
@@ -459,3 +466,140 @@ async def put(self, path):
...
+    def initialize(
+        self, ystore_class: type[BaseYStore], ywebsocket_server: JupyterWebsocketServer
+    ) -> None:
+        self.ystore_class = ystore_class
+        self.ywebsocket_server = ywebsocket_server
+
+    async def get(self, path: str) -> None:
+
+        file_id_manager = self.settings["file_id_manager"]
+        file_id = file_id_manager.get_id("/".join(self.request.path.split("/")[4:]))
```
krassowski, line 481: This looks odd. The code removes four parts of the path - why? I think that elsewhere we just access the path as-is. I would guess that it was meant to remove four letters `RTC:` if present but this is just a guess.

packages/docprovider/style/slider.css
```diff
@@ -0,0 +1,36 @@
...
+}
+
+.timestamp-display {
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  gap: 6px;
+}
+
+.restore-btn button {
```
krassowski, line 28: Bare elements (without e.g. a class name) on the right-most side are bad for performance because browsers match CSS selectors from right to left. A browser would need to check rest of the rule against all buttons on the page which might be many.

krassowski (changes requested):
This would be super exciting to have, it would be a revolution compared to the existing `.ipynb_checkpoints` mechanism!

I've tried to review that as if this PR was ready for review, I hope it is not too early to give fairly detailed feedback on the implementation.

packages/docprovider/src/ydrive.ts
krassowski: Maybe `getProviderForPath` → `getForkProvider`?

davidbrochart:
I am still unable to get this PR working. Here is what I did:
```console
micromamba create -n jupyter-collaboration
micromamba activate jupyter-collaboration
micromamba install pip nodejs
pip install -e ".[dev]"
pip install -e projects/jupyter-collaboration-ui -e projects/jupyter-docprovider -e projects/jupyter-server-ydoc
pip install jupyterlab
jupyter labextension develop --overwrite projects/jupyter-collaboration-ui
jupyter labextension develop --overwrite projects/jupyter-docprovider
```
@Meriem-BenIsmail can you confirm that it works on your side in a fresh environment?
@krassowski did it work on your side?

davidbrochart:
Ah it seems that it only works the first time a notebook is opened, but not after closing and reopening it:

https://github.com/user-attachments/assets/527df096-ca26-4f16-95d8-d513b22410a9

Meriem-BenIsmail:
> Ah it seems that it only works the first time a notebook is opened, but not after closing and reopening it:
> 
>  Peek.2024-08-27.11-12.mp4

Oh I see I haven't noticed this. It works also when you reload the page btw.
I will look into this.

projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
```diff
@@ -28,13 +29,19 @@
...
+    encode_file_path,
     room_id_from_encoded_path,
 )
-from .websocketserver import JupyterWebsocketServer
+from .websocketserver import JupyterWebsocketServer, RoomNotFound
 
 YFILE = YDOCS["file"]
+YNOTEBOOK = YDOCS["notebook"]
+YJCAD = YDOCS.get("jcad", None)
+
```
davidbrochart, line 40:
```suggestion
```

projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
```diff
@@ -459,3 +466,139 @@ async def put(self, path):
...
+                        FORK_DOCUMENTS[idx] = YJCAD(fork_ydoc)
+
+                undo_manager = FORK_DOCUMENTS[idx].undo_manager
+
+                UNDO_MANAGERS[idx] = undo_manager
+                for item in updates:
+                    fork_ydoc.apply_update(item)
+
+                fork_room = YRoom()
+                fork_room.ydoc = fork_ydoc
```
davidbrochart, line 539: It would be better to pass a ydoc to `YRoom()`, see [here](https://github.com/jupyter-server/pycrdt-websocket/pull/18/files#diff-d5f2a35bb64e14b167e41fb7ba849cb6b9cc45197bffb9f11278c457f5848b73R55) if you want to open a PR in `jupyter-ydoc`.

davidbrochart:
It seems that the changes don't appear at the same place on the slider depending on whether you're undoing or redoing:

https://github.com/user-attachments/assets/bfc2664c-6138-4739-afd0-6c84db67ca50

projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
```diff
@@ -459,3 +464,133 @@ async def put(self, path):
...
+                    {
+                        "sessionId": SERVER_SESSION,
+                        "roomId": idx,
+                    }
+                )
+                self.set_status(200)
+                return self.finish(data)
+            elif mode == "fork":
+                for _key, value in UNDO_MANAGERS.items():
+                    undo_manager = value
```
davidbrochart, line 545: I don't quite understand what this does, is your intention to get the latest undo manager? If so, since `UNDO_MANAGERS` is a global variable, you are introducing some state in the server, which shouldn't be. For instance, what happens if several clients are using the timeline at the same time?

projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
```diff
@@ -459,3 +464,133 @@ async def put(self, path):
...
+                updates = [item[0] async for item in root_room.ystore.read()]
+                fork_ydoc = Doc()
+
+                ydoc_factory = YDOCS.get(fileType)
+                if ydoc_factory is not None:
+                    FORK_DOCUMENTS[idx] = ydoc_factory(fork_ydoc)
+
+                undo_manager = FORK_DOCUMENTS[idx].undo_manager
+
+                UNDO_MANAGERS[idx] = undo_manager
```
davidbrochart, line 526: I don't see any entry being deleted from `UNDO_MANAGERS`, which means memory will grow indefinitely. Isn't there a time where an undo manager is not needed anymore, maybe when the user chooses to restore a document to some previous state?

Meriem-BenIsmail:
> It seems that the changes don't appear at the same place on the slider depending on whether you're undoing or redoing:
> 
>  Peek.2024-08-27.16-40.mp4

I think that has to do with the fact that the undo manager doesn't track empty updates, it only tracks relevant modifications to the document. But when fetching the updates/timestamps of a document from the ystore, I see that some of the updates aren't modifications to the file itself I'm not sure about the reason.

davidbrochart:
What about setting the slider's step number to the initial undo stack's length?

Meriem-BenIsmail:
> What about setting the slider's step number to the initial undo stack's length?
Will I also get the timestamps to display them in the frontend from the undostack that way ?

davidbrochart:
I think so yes, for instance when you apply all the updates to the fork, you can check if the document's undo stack increases after each applied update, and build a timestamp list that correspond to each stack item this way.

projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
```diff
@@ -28,13 +29,17 @@
...
 )
-from .websocketserver import JupyterWebsocketServer
+from .websocketserver import JupyterWebsocketServer, RoomNotFound
 
 YFILE = YDOCS["file"]
 
+
 SERVER_SESSION = str(uuid.uuid4())
+UNDO_MANAGERS = {}
+FORK_DOCUMENTS = {}
```
davidbrochart, line 42: I don't think we need to keep track of fork documents and undo managers in `FORK_DOCUMENTS` and `UNDO_MANAGERS`. They can both be accessed from the room: `yroom.ydoc` and `yroom.ydoc.undo_manager`.

Meriem-BenIsmail: I think the `yroom.ydoc` doesn't have an `undo_manager` but this does https://github.com/jupyter-server/jupyter_ydoc/blob/main/jupyter_ydoc/ybasedoc.py#L10 right ? so to access the `undo_manager` , i do this for example ` fork_document = ydoc_factory(fork_ydoc) `

Meriem-BenIsmail: > I think the `yroom.ydoc` doesn't have an `undo_manager` but this does https://github.com/jupyter-server/jupyter_ydoc/blob/main/jupyter_ydoc/ybasedoc.py#L10 right ? so to access the `undo_manager` , i do this for example `fork_document = ydoc_factory(fork_ydoc)`

I could remove UNDO_MANAGERS but I think I need to keep FORK_DOCUMENTS so i could access the undo manager

davidbrochart: Good point, yes you can at least remove `UNDO_MANAGERS`, or have a subclass of `YRoom` for your `fork_room` where you add a property for the fork document (that way you can get rid of `FORK_DOCUMENTS` too).

davidbrochart:
There is something weird happening when after a document was restored to a previous state:
- when restored, the dirty indicator should be cleared, meaning that the file has been saved.
- after reopening the file, the timeline seems to be broken:

https://github.com/user-attachments/assets/7fe1274b-0832-45ad-b796-1f98eba582f1

davidbrochart:
> I think the dirty indicator is not cleared because even if we could restore the document state, the displayed version is a still a fork so maybe that's the cause.

You don't reconnect to the root document after restoring?

Meriem-BenIsmail:
> > I think the dirty indicator is not cleared because even if we could restore the document state, the displayed version is a still a fork so maybe that's the cause.
> 
> You don't reconnect to the root document after restoring?

I could do that but I would still need to reload the document if I want to use the timeline again.

davidbrochart:
The document would reload by itself, just by reconnecting to the root.

Meriem-BenIsmail:
> The document would reload by itself, just by reconnecting to the root.

but i still need to refetch the new timeline so i would need to reload the browser and i don't think that happens automatically if i reconnect.

davidbrochart:
Refreshing the browser is not an option but you can surely update the timeline.

davidbrochart:
Thanks for keeping up the great work @Meriem-BenIsmail, this is starting to look really nice! I like the user experience.
One thing though is that I see you don't allow to undo the very first change [here](https://github.com/jupyterlab/jupyter-collaboration/pull/338/files#diff-c9386f522b08fc1b0e605b66df3233869f8566afd42575fee4528d8504d696a1R576):
```py
if action == "undo" and len(undo_manager.undo_stack) > 1:
    undo_manager.undo()
```
I suppose this has something to do with the fact that JupyterLab doesn't like empty notebooks (see [this issue](https://github.com/jupyterlab/jupyterlab/issues/16718))? For text files however, this means that we cannot go back to to an empty file:

https://github.com/user-attachments/assets/de33a367-564e-4e91-a31a-9a9bff134e56

projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
```diff
@@ -459,3 +463,125 @@ async def put(self, path):
...
+
+            elif action == "redo" and undo_manager.can_redo():
+                undo_manager.redo()
+            else:
+                break
+
+    async def _cleanup_undo_manager(self, room_id: str) -> None:
+        if room_id in FORK_DOCUMENTS:
+            del FORK_DOCUMENTS[room_id]
+            print(f"Fork Document for {room_id} has been removed.")
```
davidbrochart, line 587: You should use the logger instead of printing.

projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
```diff
@@ -459,3 +463,125 @@ async def put(self, path):
...
+        format = str(self.request.query_arguments.get("format")[0].decode("utf-8"))
+        content_type = str(self.request.query_arguments.get("type")[0].decode("utf-8"))
+        encoded_path = encode_file_path(format, content_type, file_id)
+        try:
+            room_id = room_id_from_encoded_path(encoded_path)
+            room: YRoom = await self.ywebsocket_server.get_room(room_id)
+            fork_ydoc = Doc()
+
+            ydoc_factory = YDOCS.get(content_type)
+            if ydoc_factory is not None:
```
davidbrochart, line 489: The handler should return an error when there is no document factory.

projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
```diff
@@ -459,3 +463,125 @@ async def put(self, path):
...
+                    await self._perform_undo_or_redo(undo_manager, "redo", steps)
+                    self.set_status(200)
+                    return self.finish({"status": "redone"})
+                else:
+                    return self.finish({"error": "No more redo operations available"})
+            elif action == "restore":
+                try:
+                    root_room = await self._websocket_server.get_room(room_id)
+                    original_ydoc = root_room.ydoc
+                    original_ydoc.apply_update(fork_document.ydoc.get_update())
```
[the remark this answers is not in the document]
Meriem-BenIsmail: This is done before reconnecting to the root though, this is done when clicking on the restore button making it the new state of the original doc so it needs to be done in the backend.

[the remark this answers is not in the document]
davidbrochart: I don't think it matters if changes to the root document are applied before or after reconnecting to the root, does it?

davidbrochart:
> Do you have an idea if this issue only appears in notebooks ?

Potentially yes, but it can be considered as a bug if a document doesn't allow an "empty state", so let's keep it like that for now, and revisit the issue if/when https://github.com/jupyterlab/jupyterlab/issues/16718 is fixed.

projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
```diff
@@ -459,3 +463,141 @@ async def put(self, path):
...
+            result_timestamps = []
+
+            for update, timestamp in updates_and_timestamps:
+                undo_stack_len = len(undo_manager.undo_stack)
+                fork_ydoc.apply_update(update)
+                if len(undo_manager.undo_stack) > undo_stack_len:
+                    result_timestamps.append(timestamp)
+
+            fork_room = YRoom()
+            fork_room.ydoc = fork_ydoc
```
davidbrochart, line 513: With `pycrdt-websocket` v0.14.2, this can be replaced with:
```suggestion
            fork_room = YRoom(fork_ydoc)
```

projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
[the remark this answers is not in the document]
Meriem-BenIsmail: Sure yeah I removed it and it works just the same. Thanks for the suggestion.

davidbrochart:
You need to require `pycrdt-websocket>=0.14.2,<0.15.0` [here](https://github.com/jupyterlab/jupyter-collaboration/blob/1b74c437fec5475548e8cf08233c0992dcfdd4f8/projects/jupyter-server-ydoc/pyproject.toml#L34).

packages/docprovider-extension/src/filebrowser.ts
```diff
@@ -91,7 +94,7 @@ export const ynotebook: JupyterFrontEndPlugin<void> = {
   optional: [ISettingRegistry],
   activate: (
     app: JupyterFrontEnd,
-    drive: ICollaborativeDrive,
+    drive: YDrive,
```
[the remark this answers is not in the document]
davidbrochart: @Meriem-BenIsmail What do you think about @krassowski's comment?

packages/docprovider/src/TimelineSlider.tsx
```diff
@@ -0,0 +1,61 @@
...
+    try {
+      const parsedURL = new URL(url);
+      const pathname = parsedURL.pathname;
+      const segments = pathname.split('/');
+      return segments[segments.length - 1];
+    } catch (error) {
+      console.error('Invalid URL:', error);
+      return '';
+    }
+  }
```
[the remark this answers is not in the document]
davidbrochart: Ping @Meriem-BenIsmail

packages/docprovider-extension/src/filebrowser.ts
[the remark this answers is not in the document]
Meriem-BenIsmail: Yes I agree I changed it back to `ICollaborativeDrive` and then added a type guard as he suggested.

davidbrochart:
I think this is looking good. I'll give more time for review, and merge tomorrow otherwise.

davidbrochart:
@krassowski were all your requested changes applied?

davidbrochart:
Thanks again @Meriem-BenIsmail!

krassowski:
Sorry for not getting back here earlier, got sick and did not remember to comment back - most of the things were addressed so good call to merge!

Issue #341, opened 2024-08-29
Populating the path property of the ydoc


brichet:
## Description

The path property of the ydoc is not populated.

This is probably a regression included in https://github.com/jupyterlab/jupyter-collaboration/pull/174.
The path used to be set here: https://github.com/jupyterlab/jupyter-collaboration/blob/547667399adeefcf2729a2c76b4526e752efdbf6/jupyter_server_ydoc/ydoc.py#L127

## Reproduce

Open a document in collaborative mode and try to get the path from the ydoc.

## Expected behavior

The path to be populated.

## Suggestion

Set the path should be trivial at initialization, i.e. at https://github.com/jupyterlab/jupyter-collaboration/blob/b8fafc0646e03ce2dcbd27b71267b1b99c976547/projects/jupyter-server-ydoc/jupyter_server_ydoc/rooms.py#L44

Then we should probably update the path if it has changed, probably from https://github.com/jupyterlab/jupyter-collaboration/blob/b8fafc0646e03ce2dcbd27b71267b1b99c976547/projects/jupyter-server-ydoc/jupyter_server_ydoc/loaders.py#L183

PR #335, opened 2024-08-12
Bump dependencies to fix CI

krassowski:
Fixes failing CI spotted in https://github.com/jupyterlab/jupyter-collaboration/pull/334#pullrequestreview-2232542708

PR #342, opened 2024-08-29
Initialize and update the ydoc path property

brichet:
This PR fixes https://github.com/jupyterlab/jupyter-collaboration/issues/341

### NOTE:
~~In this implementation, the document is reloaded when the path change, as if the file content has changed.~~
~~Another approach could be to add a listener on the path only in the `FileLoader`, which only update the path property of the ydoc on change.~~

### DIFF
--- a/projects/jupyter-server-ydoc/jupyter_server_ydoc/loaders.py
+++ b/projects/jupyter-server-ydoc/jupyter_server_ydoc/loaders.py
@@ -42,6 +42,7 @@ class FileLoader:
 
         self._watcher = asyncio.create_task(self._watch_file()) if self._poll_interval else None
         self.last_modified = None
+        self._current_path = self.path
 
     @property
     def file_id(self) -> str:
@@ -205,8 +206,13 @@ class FileLoader:
         """
         do_notify = False
         async with self._lock:
+            path = self.path
+            if self._current_path != path:
+                self._current_path = path
+                do_notify = True
+
             # Get model metadata; format and type are not need
-            model = await ensure_async(self._contents_manager.get(self.path, content=False))
+            model = await ensure_async(self._contents_manager.get(path, content=False))
 
             if self.last_modified is not None and self.last_modified < model["last_modified"]:
                 do_notify = True
--- a/projects/jupyter-server-ydoc/jupyter_server_ydoc/loaders.py
+++ b/projects/jupyter-server-ydoc/jupyter_server_ydoc/loaders.py
@@ -39,6 +39,7 @@ class FileLoader:
 
         self._log = log or getLogger(__name__)
         self._subscriptions: dict[str, Callable[[], Coroutine[Any, Any, None]]] = {}
+        self._filepath_subscriptions: dict[str, Callable[[], Coroutine[Any, Any, None]]] = {}
 
         self._watcher = asyncio.create_task(self._watch_file()) if self._poll_interval else None
         self.last_modified = None
@@ -80,7 +81,12 @@ class FileLoader:
             except asyncio.CancelledError:
                 self._log.info(f"file watcher for '{self.file_id}' is cancelled now")
 
-    def observe(self, id: str, callback: Callable[[], Coroutine[Any, Any, None]]) -> None:
+    def observe(
+        self,
+        id: str,
+        callback: Callable[[], Coroutine[Any, Any, None]],
+        filepath_callback: Callable[[], Coroutine[Any, Any, None] | None] | None = None
+    ) -> None:
         """
         Subscribe to the file to get notified about out-of-band file changes.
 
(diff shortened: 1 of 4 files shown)


davidbrochart:
Thanks @brichet.
I tried to rename an opened file out-of-band, and it seems to crash?

brichet:
> Thanks @brichet. I tried to rename an opened file out-of-band, and it seems to crash?

Yes, it doesn't fix the issue on renaming the file out-of-band. It works when renaming the file using jupyterlab.
The issue you mentioned is related to `jupyter_server_fileid`: https://github.com/jupyter-server/jupyter_server_fileid/issues/80

brichet:
> > which only update the path property of the ydoc on change
> 
> I feel like that would be a better approach, as any reloading from disk is a risk of race condition in RTC context, right?

Any opinion on how to do it ?

The 2 options I can see are:
- adding new methods, something like `path_observer()` and `path_unobserver()`
- adding an optional argument (`Callable`) to the current `observe()` method

projects/jupyter-server-ydoc/jupyter_server_ydoc/loaders.py
```diff
@@ -204,15 +215,26 @@ async def maybe_notify(self) -> None:
         Notifies subscribed rooms about out-of-band file changes.
         """
         do_notify = False
+        filepath_change = False
         async with self._lock:
+            path = self.path
```
davidbrochart, line 220: Any reason to have this new `path` variable?

davidbrochart:
I think you can now remove the note in the [top comment](https://github.com/jupyterlab/jupyter-collaboration/pull/342#issue-2494423409)?

projects/jupyter-server-ydoc/jupyter_server_ydoc/loaders.py
brichet: To avoid having to call `self.path` several times, this function queries the database and checks certain information in the file system.

davidbrochart: Thanks!

davidbrochart (approved):
Thanks @brichet.

brichet:
Is it expected that the new test fails in the context of `Test Minimal Versions` and `test Prerelease` ?

davidbrochart:
You should be able to reproduce with `Test Minimal Versions`'s environment on you side?

brichet:
> You should be able to reproduce with `Test Minimal Versions`'s environment on you side?

It works on my side with python 3.8

davidbrochart:
I think it's not only Python, but all dependencies. You can probably use `uv` for installing minimal versions.

brichet:
> I think it's not only Python, but all dependencies. You can probably use `uv` for installing minimal versions.

Not sure about that, I checked the packages and they are similar to mine.
However, I wonder if there is not an issue with the collaboration python package versions:
[`Using cached jupyter_server_ydoc-1.0.0b1-py3-none-any.whl`](https://github.com/jupyterlab/jupyter-collaboration/actions/runs/10631063478/job/29476591052?pr=342#step:4:35)

It should be the `jupyter_server_ydoc-1.0.0b3`, no ?

krassowski:
Huh:

https://github.com/jupyterlab/jupyter-collaboration/blob/b8fafc0646e03ce2dcbd27b71267b1b99c976547/projects/jupyter-collaboration/pyproject.toml#L32-L35

https://github.com/jupyterlab/jupyter-collaboration/blob/b8fafc0646e03ce2dcbd27b71267b1b99c976547/projects/jupyter-collaboration/pyproject.toml#L3

I thought I fixed it with https://github.com/jupyterlab/jupyter-collaboration/pull/335 but clearly it did not work as intended and instead of replacing the old ones just added a new line on top:

image

which means the top one gets last one gets used

davidbrochart:
Maybe exclude `scripts/bump_version.py` for `mypy` in `pyproject.toml`?
```toml
[tool.mypy]
exclude=[
    "^binder/jupyter_config\\.py$",
    "^scripts/bump_version\\.py$",
    "/setup\\.py$",
]
```

brichet:
> Maybe exclude `scripts/bump_version.py` for `mypy` in `pyproject.toml`?

Thanks @davidbrochart.

Maybe we could exclude it, but do you know why it fails in this PR ? It seems to pass on recent ones...

davidbrochart:
> do you know why it fails in this PR ? It seems to pass on recent ones...

I don't know, I've seen that in other repos too.

brichet:
And it seems there are some other issue after the release:
[Using cached jupyter_collaboration_ui-1.0.0b3-py3-none-any.whl](https://github.com/jupyterlab/jupyter-collaboration/actions/runs/10938097933/job/30365432304?pr=342#step:4:306), it should be 1.0.0b6 I think.

I will first remove the test to be able to merge this one.

davidbrochart:
It looks like `check-links` is still failing, but let's merge this one.

brichet:
> It looks like `check-links` is still failing, but let's merge this one.

Yes, it is not related

PR #343, opened 2024-08-30
Fix bump file

brichet:
Fix the dependencies in `jupyter_collaboration` metapackage, and fix the bump script.

Initially discussed at https://github.com/jupyterlab/jupyter-collaboration/pull/342#issuecomment-2321265257

cc. @krassowski

### DIFF
--- a/projects/jupyter-collaboration/pyproject.toml
+++ b/projects/jupyter-collaboration/pyproject.toml
@@ -1,6 +1,5 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
-dependencies = ["jupyter_collaboration>=3.0.0b8", "jupyter_collaboration_ui>=1.0.0b8", "jupyter_docprovider>=1.0.0b8", "jupyter_server_ydoc>=1.0.0b8"]
 
 [build-system]
 build-backend = "hatchling.build"
@@ -30,9 +29,9 @@ classifiers = [
     "Framework :: Jupyter :: JupyterLab :: Extensions :: Prebuilt",
 ]
 dependencies = [
-    "jupyter-server-ydoc>=1.0.0b1",
-    "jupyter-collaboration-ui>=1.0.0b1",
-    "jupyter-docprovider>=1.0.0b1"
+    "jupyter-server-ydoc>=1.0.0b8",
+    "jupyter-collaboration-ui>=1.0.0b8",
+    "jupyter-docprovider>=1.0.0b8"
 ]
 dynamic = ["version"]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,7 +57,7 @@ packages = [
 allow-direct-references = true
 
 [tool.jupyter-releaser]
-skip = ["check-links"]
+skip = ["check-links", "check-python"]
 
 [tool.jupyter-releaser.options]
 # `--skip-if-dirty` is a workaround for https://github.com/jupyter-server/jupyter_releaser/issues/567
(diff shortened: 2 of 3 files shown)


brichet:
Any clue on the check-release failure ?

I can't reproduce the steps locally, so it's hard to debug.

jtpio:
> Any clue on the check-release failure ?

This seems to be failing as part of the `check-python` step, and related to https://github.com/jupyter-server/jupyter_releaser/issues/499.

Which was also noticed in https://github.com/voila-dashboards/voici/pull/106.

A workaround for now could be to disable the `check-python` step. Until there is a way to skip the `pip install` in the releaser, or be able to install all packages of the monorepo at once in this check.

brichet:
Thanks for your feedback @jtpio.

> A workaround for now could be to disable the `check-python` step. Until there is a way to skip the `pip install` in the releaser, or be able to install all packages of the monorepo at once in this check.

I'll see if I can find a way to maintain this check, otherwise we'll probably ignore it until there's a fix on the releaser.

krassowski:
So instead of skipping the `check-python` I wonder if we can tell pip to include a local registry using `--extra-index-url` or `--find-links` pointing to the local file system on the runner. This too may require a change in releaser, but that would be probably beneficial to many other packages (like `jupyter-ai` and `viola`) so I think it is worth it.

davidbrochart (approved):

davidbrochart:
It looks like we are missing the pre-release specifier:
```
ERROR: Could not find a version that satisfies the requirement jupyter-collaboration-ui>=1.0.0 (from jupyter-collaboration) (from versions: 1.0.0a0, 1.0.0a1, 1.0.0a2, 1.0.0b0, 1.0.0b1, 1.0.0b2, 1.0.0b3, 1.0.0b4, 1.0.0b5, 1.0.0b6)
ERROR: No matching distribution found for jupyter-collaboration-ui>=1.0.0
```

brichet:
> It looks like we are missing the pre-release specifier:
> 
> ```
> ERROR: Could not find a version that satisfies the requirement jupyter-collaboration-ui>=1.0.0 (from jupyter-collaboration) (from versions: 1.0.0a0, 1.0.0a1, 1.0.0a2, 1.0.0b0, 1.0.0b1, 1.0.0b2, 1.0.0b3, 1.0.0b4, 1.0.0b5, 1.0.0b6)
> ERROR: No matching distribution found for jupyter-collaboration-ui>=1.0.0
> ```

Yes, we still have an issue because of the release that do not handle installing local python package with pip (AFAIK): https://github.com/jupyterlab/jupyter-collaboration/pull/343#issuecomment-2325954113

davidbrochart:
So do you think we should merge anyway?

jtpio:
It would be better to get that `check-release` check green, to give more confidence when making releases and catching potential other issues.

davidbrochart:
@brichet do you think we should merge (after rebase)?

brichet:
Do you know if the `check-python` test has been updated on the releaser ?
Otherwise we need to ignore this test to avoid all the following PR to fail, but I'm not sure there is a consensus about that.

davidbrochart:
I don't know. Was the error the following?
```
ERROR: Could not find a version that satisfies the requirement jupyter-collaboration-ui>=1.0.0 (from jupyter-collaboration) (from versions: 1.0.0a0, 1.0.0a1, 1.0.0a2, 1.0.0b0, 1.0.0b1, 1.0.0b2, 1.0.0b3, 1.0.0b4, 1.0.0b5, 1.0.0b6, 1.0.0b8)
ERROR: No matching distribution found for jupyter-collaboration-ui>=1.0.0
```

jtpio:
https://github.com/jupyter-server/jupyter_releaser/issues/499 is still opened and there is no work-in-progress to fix this on the releaser currently.

The easiest workaround for now would likely be to skip the `check-python` step (and open a new issue to track removing it), to make sure `check_release` passes on CI.

davidbrochart:
Thanks!

PR #345, opened 2024-09-05
Register RTC content provider

davidbrochart:
Needs https://github.com/jupyterlab/jupyterlab/pull/16744.

This PR gets rid of the custom drive, and instead uses the default JupyterLab drive where it registers a content provider that uses the RTC protocol.


krassowski:
Closing as merged in https://github.com/jupyterlab/jupyter-collaboration/pull/418

Issue #352, opened 2024-09-13
Expose generic document provider API


davidbrochart:
### Problem

Currently the document providers are kept [private](https://github.com/jupyterlab/jupyter-collaboration/blob/1492678f11f0fca9555224376ed3a51701e41fad/packages/docprovider/src/ydrive.ts#L258) inside the drive. But there can be a need to programmatically create providers and have access to them from outside the drive, like in JupyterCAD where the kernel wants to connect to a shared model that is synchronized with a file.
Also, it would be nice to have a transport-agnostic provider interface, so that a client can "connect to a file" without knowing if it's using e.g. [jupyter-collaboration](https://github.com/jupyterlab/jupyter-collaboration) or [jupyter-shared-drive](https://github.com/davidbrochart/jupyter-shared-drive) (which use WebSockets and WebRTC, respectivelly).

### Proposed Solution

Extract out a document provider API and have it use a transport-agnostic interface.

davidbrochart:
Actually, the drive has a `sharedModelFactory` that can be used to create a shared model through its `createNew()` method, passing it the file name.
A first [PR in JupyterCAD](https://github.com/jupytercad/JupyterCAD/pull/425) is using that, rather than dealing with the internals of the drive (WebSocket provider, etc.). Now I think that the `ICollaborativeDrive` should not belong to `jupyter-collaboration`, but should instead be extracted out to an independent package, say `jupyter-collaborative-drive`, that `jupyter-collaboration` would depend on. Other implementations of collaborative drives, like [jupyter-shared-drive](https://github.com/davidbrochart/jupyter-shared-drive), could provide an `ICollaborativeDrive`, which JupyterCAD could use instead of `jupyter-collaboration`. Because `jupyter-share-drive` doesn't need a backend, this would allow JupyterCAD to work in JupyterLite.

PR #353, opened 2024-09-16
Extract out ICollaborativeDrive to `@jupyter/collaborative-drive`

davidbrochart:
See https://github.com/jupyterlab/jupyter-collaboration/issues/352#issuecomment-2353809355.


davidbrochart:
Should there be a new `projects/jupyter-collaborativedrive`?

davidbrochart:
I'm wondering if [this failure](https://github.com/jupyterlab/jupyter-collaboration/actions/runs/10919722949/job/30308001718?pr=353):
```
Error: src/yprovider.ts(6,35): error TS2307: Cannot find module '@jupyter/collaborativedrive' or its corresponding type declarations.
Error: src/ydrive.ts(16,8): error TS2307: Cannot find module '@jupyter/collaborativedrive' or its corresponding type declarations.
```
is caused by the fact that `@jupyter/collaborativedrive` is not yet published to NPM?
I cannot reproduce it, it works locally.

davidbrochart:
@krassowski What do you think about getting this in before releasing v3?
We would need to publish `@jupyter/collaborativedrive` to NPM.

davidbrochart:
@jtpio do you have the rights to publish `@jupyter/collaborativedrive` to NPM?

packages/collaborativedrive/package.json
```diff
@@ -0,0 +1,55 @@
+{
+  "name": "@jupyter/collaborativedrive",
```
jtpio, line 2: Maybe we should rename the package to the following, so it reads better, before publishing it?

```suggestion
  "name": "@jupyter/collaborative-drive",
```

davidbrochart: Fine with me :+1:

davidbrochart:
@krassowski @jtpio ok to merge and publish `@jupyter/collaborative-drive`?

davidbrochart:
I'm planning to merge today.

.github/workflows/test.yml
```diff
@@ -71,6 +71,7 @@ jobs:
           pip install "jupyterlab>=4.0.0,<5"
           pip install -e .
           jlpm
+          jlpm build
```
krassowski, line 74: Why do we need it here?

davidbrochart: I don't remember the reason, I'll remove it to see what happens.

davidbrochart: Maybe it will be fine once `@jupyter/collaborative-drive` is published?

jtpio: Ah, maybe I can help publish a placeholder version to see if it helps.

packages/collaborativedrive/package.json
jtpio, line 1: Let's rename the folder from:

```
packages/collaborativedrive
```

To

```
packages/collaborative-drive
```

So it's consistent with the package name?

.github/workflows/test.yml
jtpio: Just published the package: https://www.npmjs.com/package/@jupyter/collaborative-drive

jtpio: It looks like a version with the previous name was also published earlier: https://www.npmjs.com/package/@jupyter/collaborativedrive

Maybe it can be removed from `npm` so it's less confusing?

davidbrochart: Now I'm confused, who published that? Do we use it somewhere?

jtpio:
> I guess the published package is useless without code.

What do you mean by "without code"? The package was published from this branch directly.

davidbrochart:
Oh my bad, yes the code is there. Then I don't know what is wrong.

.github/workflows/test.yml
jtpio: > Now I'm confused, who published that? Do we use it somewhere?

Looks like you did?

image

jtpio: I guess you published it to see if that would help fix the CI?

jtpio:
Having to add `jlpm run build` makes it look like something is wrong in the setup of the new package, since it was not needed before.

yarn.lock
```diff
@@ -2122,10 +2123,22 @@ __metadata:
   languageName: unknown
   linkType: soft
 
+"@jupyter/collaborative-drive@^3.0.0-beta.6, @jupyter/collaborative-drive@workspace:packages/collaborative-drive":
```
jtpio, line 2126: This line looks different than for the other packages.

davidbrochart: Good catch, let's try to make it similar.

davidbrochart: So it's Yarn that adds `@jupyter/collaborative-drive@^3.0.0-beta.6`.

davidbrochart:
> > I guess the published package is useless without code.
> 
> What do you mean by "without code"? The package was published from this branch directly.

Actually yes, there is no code [here](https://www.npmjs.com/package/@jupyter/collaborative-drive?activeTab=code).

davidbrochart:
I published `@jupyter/collaborative-drive` v3.0.0-beta.7, and there is no `IDocumentProvider` in `lib/tokens.js` [here](https://www.npmjs.com/package/@jupyter/collaborative-drive/v/3.0.0-beta.7?activeTab=code). It is in `lib/tokens.d.ts` but it's not enough, right?

davidbrochart:
In the end, I just think that `jlpm build` is needed because the test in [docprovider](https://github.com/jupyterlab/jupyter-collaboration/blob/08df1514354abef1e0351738814804f5091977e5/packages/docprovider/src/__tests__/yprovider.spec.ts) now needs `@jupyter/collaborative-drive` through [yprovider.ts](https://github.com/jupyterlab/jupyter-collaboration/blob/4c685cab15d845911e0f6f608322cd3351bca0fc/packages/docprovider/src/yprovider.ts#L6), so it has to be compiled.

davidbrochart:
I'll take the risk to merge this PR and release jupyter-collaboration v3.0.0beta7. We can still revert the changes and release v3.0.0beta8 if it doesn't work.

jtpio:
Arf, I wanted to help you more on this yesterday but didn't have time.

At the very least it would have been great to leave a TODO, or open an issue...

Maybe check the `tsconfig.json` to see if it needs a `references`?

PR #355, opened 2024-09-18
Add missing aria label to the file browser

trungleduc:
This PR fixes the issue of the custom file browser and Galata helpers. The labels are taken from here https://github.com/jupyterlab/jupyterlab/blob/main/packages/filebrowser-extension/src/index.ts#L341


trungleduc:
cc @davidbrochart

PR #356, opened 2024-09-18
Remove the unused `ISettingRegistry` token from the file browser plugin

trungleduc:
3.0.0b5 is broken due to this token.


jtpio (approved):
Good catch, thanks!

jtpio:
`check-links` seems to be failing consistently and can likely be tracked separately.

trungleduc:
Thanks @jtpio, could you trigger the release action? The current version is broken

davidbrochart:
Done.

krassowski:
In an unlikely circumstance if anyone else finds them bitten by this - the error on the server side would be:

```
tornado.web.HTTPError: HTTP 404: Not Found (Settings not found (jupyterlab))
```

because the `.load('jupyter')` method was called on settings registry, not on translator. In UI the consequence was that the file browser was not showing up.

Thanks for the quick fix!

Should we add a snapshot test for filebrowser showing up?

PR #358, opened 2024-09-19
Fix mypy

brichet:
This PR should fix mypy which fails with the `bump_version.py` file.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -60,7 +60,7 @@ repos:
     rev: v0.991
     hooks:
       - id: mypy
-        exclude: "(^binder/jupyter_config\\.py$)|(/setup\\.py$)"
+        exclude: "(^binder/jupyter_config\\.py$)|(^scripts/bump_version\\.py$)|(/setup\\.py$)"
         args: ["--config-file", "pyproject.toml"]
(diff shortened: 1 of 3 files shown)


davidbrochart (approved):
Thanks, failures are unrelated, let's fix them in another PR.

davidbrochart:
Wait, pre-commit is still failing? It means mypy is still analysing this file? Also, we don't need `# type: ignore` anymore:
```py
from pkg_resources import parse_version  # type: ignore
```

brichet:
> Wait, pre-commit is still failing? It means mypy is still analysing this file? 

Yes, I saw that...

> Also, we don't need `# type: ignore` anymore:
> 
> ```python
> from pkg_resources import parse_version  # type: ignore
> ```

The 2 first commits that I reverted fixed mypy for that file.

brichet:
Thanks @davidbrochart.

Should we remove 
```exclude=[
    "^binder/jupyter_config\\.py$",
    "^scripts/bump_version\\.py$",
    "/setup\\.py$",
]
```
from *pyproject.toml* to avoid confusion, or is this useful for local run ?

davidbrochart:
Yes I think it's still useful outside of pre-commit.

PR #365, opened 2024-09-25
Provide the `YDrive` as the default drive

jtpio:
Alternative to https://github.com/jupyterlab/jupyter-collaboration/pull/345

Needs https://github.com/jupyterlab/jupyterlab/pull/16794

Register the `YDrive` as a the default drive in JupyterLab, relying on the split of the service manager plugins in https://github.com/jupyterlab/jupyterlab/pull/16794.

One of the nice side effect is the removal of the `RTC:` prefix, which should help fix some issues reported on this repo.

[rtc-swap-default-drive.webm](https://github.com/user-attachments/assets/ec6e3262-e80f-4af5-b92f-e7d44e12453c)

- Based on https://github.com/jupyterlab/jupyterlab/pull/16794
- Linking to the upstream `@jupyterlab/services` package with `yalc` for testing

### DIFF
--- a/packages/docprovider-extension/src/filebrowser.ts
+++ b/packages/docprovider-extension/src/filebrowser.ts
@@ -38,6 +38,12 @@ import {
 } from '@jupyter/docprovider';
 import { Awareness } from 'y-protocols/awareness';
 import { URLExt } from '@jupyterlab/coreutils';
+import {
+  IDefaultDrive,
+  IUserManager,
+  ServiceManagerPlugin,
+  User
+} from '@jupyterlab/services';
 
 /**
  * The command IDs used by the file browser plugin.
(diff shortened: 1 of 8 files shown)


packages/docprovider-extension/src/filebrowser.ts
```diff
@@ -50,21 +56,33 @@ const DOCUMENT_TIMELINE_URL = 'api/collaboration/timeline';
...
-    app.serviceManager.contents.addDrive(drive);
-    return drive;
+    ydrive: YDrive,
+    globalAwareness: Awareness | null,
+    translator: ITranslator | null
+  ): void => {
+    ydrive.translator = (translator ?? nullTranslator).load('jupyterlab');
+    if (globalAwareness) {
+      ydrive.globalAwareness = globalAwareness;
+    }
```
jtpio, line 85: Because the service manager plugins are activated first and their `app` is still `null` during the first phase, we extract setting the `translator` and `globalAwareness` to a separate plugin, so it is activated as a regular plugin.

jtpio:
Closing as replaced by #418.

PR #367, opened 2024-10-10
Update jupyter_ydoc and pycrdt_websocket dependencies

brichet:
This PR updates dependencies:
- `jupyter_ydoc >=2.1.2`
- `pycrdt_websocket >=0.15.0`
- indirectly `pycrdt >=0.10.1,<0.11.0`

These changes allow the use of document awareness on the server side, to handle states on server. 

- add the awareness to the `YDocument` (after https://github.com/jupyter-server/jupyter_ydoc/pull/277)
- this awareness is updated with the remote clients changes (https://github.com/jupyter-server/pycrdt-websocket/pull/76)
- the changes in its local state are broadcasted to remote clients (https://github.com/jupyter-server/jupyter_ydoc/pull/277)

### DIFF
--- a/projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
+++ b/projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
@@ -286,31 +286,6 @@ class YDocWebSocketHandler(WebSocketHandler, JupyterHandler):
         """
         message_type = message[0]
 
-        if message_type == YMessageType.AWARENESS:
-            # awareness
-            skip = False
-            changes = self.room.awareness.get_changes(message[1:])
-            added_users = changes["added"]
-            removed_users = changes["removed"]
-            for i, user in enumerate(added_users):
-                u = changes["states"][i]
-                if "user" in u:
-                    name = u["user"]["name"]
-                    self._websocket_server.connected_users[user] = name
-                    self.log.debug("Y user joined: %s", name)
-            for user in removed_users:
-                if user in self._websocket_server.connected_users:
-                    name = self._websocket_server.connected_users[user]
-                    del self._websocket_server.connected_users[user]
-                    self.log.debug("Y user left: %s", name)
-            # filter out message depending on changes
-            if skip:
-                self.log.debug(
-                    "Filtered out Y message of type: %s",
-                    YMessageType(message_type).name,
-                )
-                return skip
-
         if message_type == MessageType.CHAT:
             msg = message[2:].decode("utf-8")
 
--- a/projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
+++ b/projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
@@ -137,6 +137,9 @@ class YDocWebSocketHandler(WebSocketHandler, JupyterHandler):
                         exception_handler=exception_logger,
                     )
 
+                    # Listen for the changes in GlobalAwareness to update users
+                    self.room.awareness.observe(self._on_global_awareness_event)
+
             try:
                 await self._websocket_server.start_room(self.room)
             except Exception as e:
@@ -379,6 +382,31 @@ class YDocWebSocketHandler(WebSocketHandler, JupyterHandler):
                 await self._file_loaders.remove(file_id)
                 self._emit(LogLevel.INFO, "clean", "Loader deleted.")
             del self._room_locks[self._room_id]
+
+    def _on_global_awareness_event(self, type: str, changes: tuple[dict[str, Any], Any]) -> None:
+        """
+        Update the users when the global awareness changes.
+
+
+            Parameters:
+                type: 'update' or 'change' (change is triggered only if the states are modified)
+                changes: the changes
+        """
+        if type != "change":
+            return
+        added_users = changes[0]["added"]
+        removed_users = changes[0]["removed"]
+        for user in added_users:
+            u = self.room.awareness.states[user]
+            if "user" in u:
+                name = u["user"]["name"]
+                self._websocket_server.connected_users[user] = name
+                self.log.debug("Y user joined: %s", name)
+        for user in removed_users:
+            if user in self._websocket_server.connected_users:
+                name = self._websocket_server.connected_users[user]
+                del self._websocket_server.connected_users[user]
+                self.log.debug("Y user left: %s", name)
 
     def check_origin(self, origin):
         """
--- a/projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
+++ b/projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
@@ -8,7 +8,7 @@ import json
 import time
 import uuid
 from logging import Logger
-from typing import Any
+from typing import Any, Literal
 from uuid import uuid4
 
 from jupyter_server.auth import authorized
@@ -384,7 +384,9 @@ class YDocWebSocketHandler(WebSocketHandler, JupyterHandler):
                 self._emit(LogLevel.INFO, "clean", "Loader deleted.")
             del self._room_locks[self._room_id]
 
-    def _on_global_awareness_event(self, topic: str, changes: tuple[dict[str, Any], Any]) -> None:
+    def _on_global_awareness_event(
+        self, topic: Literal["change", "update"], changes: tuple[dict[str, Any], Any]
+    ) -> None:
         """
         Update the users when the global awareness changes.
 
@@ -392,7 +394,7 @@ class YDocWebSocketHandler(WebSocketHandler, JupyterHandler):
                 topic (str): `"update"` or `"change"` (`"change"` is triggered only if the states are modified).
                 changes (tuple[dict[str, Any], Any]): The changes and the origin of the changes.
         """
-        if type != "change":
+        if topic != "change":
             return
         added_users = changes[0]["added"]
         removed_users = changes[0]["removed"]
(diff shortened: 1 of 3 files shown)


projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
```diff
@@ -137,6 +137,9 @@ def exception_logger(exception: Exception, log: Logger) -> bool:
                         exception_handler=exception_logger,
                     )
 
+                    # Listen for the changes in GlobalAwareness to update users
+                    self.room.awareness.observe(self._on_global_awareness_event)
```
davidbrochart, line 141: This room is not necessarily the global awareness. Should we check that the room ID is `"JupyterLab:globalAwareness"` before observing awareness changes?

projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
```diff
@@ -405,6 +383,31 @@ async def _clean_room(self) -> None:
                 self._emit(LogLevel.INFO, "clean", "Loader deleted.")
             del self._room_locks[self._room_id]
 
+    def _on_global_awareness_event(self, type: str, changes: tuple[dict[str, Any], Any]) -> None:
+        """
+        Update the users when the global awareness changes.
+
+
```
davidbrochart, line 390:
```suggestion
```

projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
```diff
@@ -405,6 +383,31 @@ async def _clean_room(self) -> None:
                 self._emit(LogLevel.INFO, "clean", "Loader deleted.")
             del self._room_locks[self._room_id]
 
+    def _on_global_awareness_event(self, type: str, changes: tuple[dict[str, Any], Any]) -> None:
```
davidbrochart, line 386: `type` is a keyword in Python, better to use something else.
```suggestion
    def _on_global_awareness_event(self, topic: str, changes: tuple[dict[str, Any], Any]) -> None:
```

projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
```diff
@@ -405,6 +383,31 @@ async def _clean_room(self) -> None:
...
+        """
+        Update the users when the global awareness changes.
+
+
+            Parameters:
+                type: 'update' or 'change' (change is triggered only if the states are modified)
+                changes: the changes
```
davidbrochart, line 393:
```suggestion
                topic (str): `"update"` or `"change"` (`"change"` is triggered only if the states are modified).
                changes (tuple[dict[str, Any], Any]): The changes and the origin of the changes.
```

projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
```diff
@@ -405,6 +383,31 @@ async def _clean_room(self) -> None:
...
+        for user in added_users:
+            u = self.room.awareness.states[user]
+            if "user" in u:
+                name = u["user"]["name"]
+                self._websocket_server.connected_users[user] = name
+                self.log.debug("Y user joined: %s", name)
+        for user in removed_users:
+            if user in self._websocket_server.connected_users:
+                name = self._websocket_server.connected_users[user]
+                del self._websocket_server.connected_users[user]
```
davidbrochart, line 408:
```suggestion
                name = self._websocket_server.connected_users.pop(user)
```

projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
```diff
@@ -405,6 +383,31 @@ async def _clean_room(self) -> None:
...
+
+
+            Parameters:
+                type: 'update' or 'change' (change is triggered only if the states are modified)
+                changes: the changes
+        """
+        if type != "change":
+            return
+        added_users = changes[0]["added"]
+        removed_users = changes[0]["removed"]
```
davidbrochart, line 398: What about `"updated"` users?

davidbrochart (changes requested):

projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
brichet: Actually I wonder if the `connected_users` of the `_websocket_server` is even used.
I copied that code from the previous message handler, but I don't know if we need that function.
If we keep it, we should indeed handle the "updated" users (removing the former name and add the new one).

brichet: > If we keep it, we should indeed handle the "updated" users (removing the former name and add the new one).

Thinking again about that, we don't have the previous name, the changes contain only the client ids.
We should rebuild the full list from the global awareness when a user is updated, because we don't know if the user name has been updated.

davidbrochart: Right, the goal of this code was to check in the backend that the awareness information from the frontend is correct. For instance, we don't want a student to take the user name of the teacher. That's why there is this [skip](https://github.com/jupyterlab/jupyter-collaboration/blob/fea37f00c743d504ce1a2361552942c284a36ca6/projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py#L307) variable that was supposed to filter out an awareness message, but this was never put to actual use.
I'm wondering how we can filter out a message with your changes though?

brichet: I don't know if we can filter out a message with the current information we have.
For example when you reload the page, it adds a new client id with the same user information. The old client id is removed in a second step (probably when a client has a lack of update from it ?).
The same user is duplicated over a period of time, with 2 different client IDs. If we allow this behavior, I don't know how we can filter out someone trying to cheat.

The following image shows some logs when a remote client reloads the page. When a change is received, the change and the current users in the awareness are printed.
Screenshot from 2024-10-15 16-10-45

davidbrochart: The backend is the authority for user identities, and the frontend can get its identity at [/api/me](https://github.com/jupyter-server/jupyter_server/blob/74655ce66f36ed85a83591e6658e70ba91232580/jupyter_server/services/api/handlers.py#L76), so we should be able to check that they match.

davidbrochart: I agree that it's out of scope for this PR, but I'm wondering if we still get the possibility to discard an awareness update?

brichet: If possible (I don't know how we can check authentication), the simplest is probably to do it in the YRoom directly, before updating the awareness ([here](https://github.com/jupyter-server/pycrdt-websocket/blob/449ba9f8eb15daba4e0946608079f6bb3112213c/pycrdt_websocket/yroom.py#L294-L311))
Otherwise, we can keep the previous logic, if the `message_type` is awareness and should be skipped, the message is not put in the queue.

davidbrochart: The logic to filter out messages is [here](https://github.com/jupyter-server/pycrdt-websocket/blob/449ba9f8eb15daba4e0946608079f6bb3112213c/pycrdt_websocket/yroom.py#L269-L275) already, but how can we connect it to the changes here?

davidbrochart: I think that what's wrong is that with this PR, we apply the awareness updates coming from clients to the awareness in the backend, and then we observe the changes and decide if the updates should be forwarded to other clients, but it's already too late, right? We should not apply the awareness in the first place if it's not correct.

davidbrochart: I think I never finished the work, but `self.on_message` should be a callback that would basically be [this](https://github.com/jupyterlab/jupyter-collaboration/blob/1492678f11f0fca9555224376ed3a51701e41fad/projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py#L283-L312).

brichet: > We should not apply the awareness in the first place if it's not correct.

This is what I meant with "*the simplest is probably to do it in the YRoom directly, before updating the awareness*"

> I think that what's wrong is that with this PR, we apply the awareness updates coming from clients to the awareness in the backend, and then we observe the changes and decide if the updates should be forwarded to other clients, but it's already too late, right? 

For what I understand the message is forwarded to client at the same time the awareness in the backend is updated, [here](https://github.com/jupyter-server/pycrdt-websocket/blob/449ba9f8eb15daba4e0946608079f6bb3112213c/pycrdt_websocket/yroom.py#L302-L311), and then the user list is updated in the `_websocket_server`.
But probably I miss something...

davidbrochart: It's actually used [in jupyverse](https://github.com/jupyter-server/jupyverse/blob/fff0274a49c4f2c3666a3669c5c396a1d1f5e271/plugins/yjs/fps_yjs/routes.py#L176).

davidbrochart: > For what I understand the message is forwarded to the clients at the same time the awareness in the backend is updated, [here](https://github.com/jupyter-server/pycrdt-websocket/blob/449ba9f8eb15daba4e0946608079f6bb3112213c/pycrdt_websocket/yroom.py#L302-L311), and then the user list is updated in the `_websocket_server`.

Yes but we should not execute this code if the awareness update is "invalid".

davidbrochart (approved):
Thanks @brichet, let's merge for now?

PR #373, opened 2024-10-15
Fix errors in contributing.rst for setup

JasonWeill:
Fix errors in setup docs: removes a redundant `pip` command, and updates the repository's punctuation.

### DIFF
--- a/docs/source/developer/contributing.rst
+++ b/docs/source/developer/contributing.rst
@@ -31,14 +31,12 @@ Installing JupyterLab Real-Time Collaboration
 
 The development version of the server requires `node <https://nodejs.org/en/download/>`_ and `pip <https://pip.pypa.io/en/stable/installing/>`_.
 
-Once you have installed the dependencies mentioned above, use the following
-steps:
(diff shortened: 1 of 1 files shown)


davidbrochart:
> removes a redundant `pip` command

Why is it redundant?

JasonWeill:
Line 32 mentions that users need to install both `node` and `pip` before starting with installation. Line 34 starts, "Once you have installed the dependencies above …". Either we need to delete line 32 and start the instructions with installing both `node` and `pip`, or the documentation needs to presume, at the start of the code block, that the user has already installed both `node` and `pip`. The previous version of this file included instructions to install `pip`, but not `node`.

davidbrochart:
But `pip install --upgrade pip` *upgrades* `pip`, right?

JasonWeill:
It does; is that required? As long as the user has supported versions of `pip` and `node`, they should be OK to start with the command after the "upgrade `pip`" command.

jtpio (approved):
Thanks!

PR #378, opened 2024-10-18
Widget to display users of a collaborative document

brichet:
This PR provides a widget that can be used in a collaborative document to display current users.

This widget is a refactoring of a widget initially created in [jupyterCAD](https://github.com/jupytercad/JupyterCAD).
Porting it to `jupyter_collaboration` has been initially discussed in https://github.com/jupytercad/JupyterCAD/issues/496.

PR #379, opened 2024-10-21
Users item toolbar

brichet:
*Copy of https://github.com/jupyterlab/jupyter-collaboration/pull/378, to try to fix the `check_release` workflow.*

This PR provides a widget that can be used in a collaborative document to display current users.

This widget is a refactoring of a widget initially created in [jupyterCAD](https://github.com/jupytercad/JupyterCAD).
Porting it to `jupyter_collaboration` has been initially discussed in https://github.com/jupytercad/JupyterCAD/issues/496.

**EDIT:**

If the widget is integrated to a notebook for example, it looks like:
image

<details>
<summary>Code for the notebook widget (in Jupyterlab)</summary>

The Notebook is a special case because the document model can be null when first displayed. We need to handle the `modelChange` event.

- packages/notebook-extension/src/index.ts

  ```ts
    toolbarRegistry.addFactory<NotebookPanel>(FACTORY, 'usersItem', panel => {
      const container = new Panel();
      const onModelChanged = () => {
        container.widgets.forEach(w => container.layout?.removeWidget(w));
        container.addWidget(UsersItem.createWidget({ model: panel.model }));
      }
      onModelChanged();
      panel.content.modelChanged.connect(onModelChanged);
      return container;
    });
  ```

- packages/notebook-extension/schema/panel.json

   ```json
    {
      "name": "usersItem",
      "rank": 1005
    }
   ```
</details>


packages/docprovider/src/users-item.tsx
```diff
@@ -0,0 +1,192 @@
+/*
```
martinRenou, line 1: Is docprovider the right place for this, or should be in the `collaboration` package?

brichet: I wondered the same, and I put it here because it is really a document related widget, it supposed to handle the document current collaborators.
But I don't have a strong opinion on it, and can move it to `collaboration`.

jtpio:
Should this be merged now while `jupyter-collaboration` is currently in RC, with `3.0.0` final to be released in the coming days?

Or should this be part of a `3.1.0` release, since it's more of a new feature?

brichet:
> Should this be merged now while `jupyter-collaboration` is currently in RC, with `3.0.0` final to be released in the coming days?
> 
> Or should this be part of a `3.1.0` release, since it's more of a new feature?

Right, we should probably wait for the next release.

jtpio:
Looks like this repo does not make active use of GitHub milestones, but it could be useful to start using them to track such issues and improvements.

I just created the following two milestones:
- `3.0.x`: https://github.com/jupyterlab/jupyter-collaboration/milestone/4
- `3.1.0`: https://github.com/jupyterlab/jupyter-collaboration/milestone/3

Feel free to move it back to the `3.0.x` if you think this should be released sooner.

jtpio:
btw @brichet if you have a screenshot / screencast around that could be posted in the description that would be really useful, thanks! (and we could also add it to the changelog when releasing this)

brichet:
> btw @brichet if you have a screenshot / screencast around that could be posted in the description that would be really useful, thanks! (and we could also add it to the changelog when releasing this)

Thanks, That helped me to finish this PR, I tested it only in jupyter-collaboration v2 and forgot to copy the styles in there.

martinRenou (approved):
Thanks!

brichet:
Should we merge this PR, since we are currently in a RC release ?
It was supposed to be in 3.1.0: https://github.com/jupyterlab/jupyter-collaboration/milestone/3

davidbrochart:
We can release v3.1.0 and then merge this PR, what do you think?

brichet:
> We can release v3.1.0 and then merge this PR, what do you think?

Make sense, then we can probably release a 3.1.1, since this PR simply provide a new widget.

martinRenou:
If we're strictly following semver, that should be a 3.2.0

davidbrochart:
I don't mind having this PR in v3.2.0, but I think the release cycle is a bit overkill here (releasing alphas, betas, RCs...).
@krassowski what do you think about changing the releaser so that we can directly bump a minor version, without going through the full release cycle?

martinRenou:
You should be able to run prep-release with `3.1.0` instead of `next` if you want to directly release `3.1.0`?

davidbrochart:
No, see the scheme to bump the version:
https://github.com/jupyterlab/jupyter-collaboration/blob/b3d8d29e3ea6e9cf3c56bada86b15b612355339a/scripts/bump_version.py#L15-L52

martinRenou:
Ok, it seems I'm used to releasing jupytercad which does what I said: https://github.com/jupytercad/JupyterCAD/blob/main/python/jupytercad/scripts/bump-version.py#L49

krassowski:
If there is a self-contained addition which warrants minor bump but not much else I am in favour of skipping some of the pre-release stages and e.g. going straight to RC (while bumping the minor) over just bumping the patch version.

davidbrochart:
In the meantime, I think we can release v3.1.0, if everybody agrees.

davidbrochart:
Sorry @brichet, but could you rebase one more time?

davidbrochart:
Tests on Ubuntu all fail in `main` too :eyes:

davidbrochart:
I cannot reproduce locally on my Debian.

brichet:
Me neither on Ubuntu

davidbrochart (approved):
Let's merge and address the CI issue on Ubuntu separately.
Thanks @brichet.

jtpio:
> In the meantime, I think we can release v3.1.0, if everybody agrees.

FYI @davidbrochart next time when making a final release, you can check the following box when triggering the workflow

image


Otherwise the release does not show any merged PR, which is misleading: https://github.com/jupyterlab/jupyter-collaboration/releases/tag/v3.1.0

image

davidbrochart:
Right, I forgot about that. I'm wondering it should not be checked by default.

brichet:
Is this possible to release this change in `v3`, or are we blocked by #436 ?

andrii-i:
Hi everyone. I've opened a PR https://github.com/jupyterlab/jupyter-collaboration/pull/439 that rebases 3.x onto main up to v3.1.0 which will unblock backporting to 3.x including this PR. Please give #439 a look when possible.

Alternative simpler option would be to hard reset 3.x to https://github.com/jupyterlab/jupyter-collaboration/commit/47305e4ed8414300ac74bd7dc64acd54f6f1c0ba but then any commits that are part of 3.x and are not part of main up to https://github.com/jupyterlab/jupyter-collaboration/commit/47305e4ed8414300ac74bd7dc64acd54f6f1c0ba would be lost.

brichet:
@meeseeksdev please backport to 3.x

PR #382, opened 2024-10-23
Add forking API

davidbrochart:
The forking API consists of the following endpoints:
- `PUT /api/collaboration/fork/{root_roomid}`: creates a fork of a root document (optionally synchronizing with the root document).
- `DELETE /api/collaboration/fork/{fork_roomid}`: deletes a fork of a root document (optionally merging back in the root document).
- `GET /api/collaboration/fork/{root_roomid}`: returns the fork IDs of a root document given by its ID.

Jupyter [events](https://github.com/jupyter/jupyter_events) are emitted when creating and deleting forks.

### DIFF
--- a/projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
+++ b/projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
@@ -39,7 +39,7 @@ YFILE = YDOCS["file"]
 
 SERVER_SESSION = str(uuid.uuid4())
 FORK_DOCUMENTS = {}
-FORK_ROOMS = {}
+FORK_ROOMS: dict[str, str] = {}
 
 
 class YDocWebSocketHandler(WebSocketHandler, JupyterHandler):
@@ -606,8 +606,9 @@ class UndoRedoHandler(APIHandler):
 class DocForkHandler(APIHandler):
     """
     Jupyter Server handler to:
-    - create a fork of a document (optionally synchronizing with the root document),
-    - delete a fork of a document (optionally merging in the root document).
+    - create a fork of a root document (optionally synchronizing with the root document),
+    - delete a fork of a root document (optionally merging back in the root document).
+    - get fork IDs of a root document.
     """
 
     auth_resource = "contents"
@@ -617,6 +618,22 @@ class DocForkHandler(APIHandler):
         ywebsocket_server: JupyterWebsocketServer,
     ) -> None:
         self._websocket_server = ywebsocket_server
+
+    @web.authenticated
+    @authorized
+    async def get(self, root_roomid):
+        """
+        Returns a dictionary of given root room ID to the root's fork room IDs.
+        """
+        self.write(
+            {
+                root_roomid: [
+                    fork_roomid
+                    for fork_roomid, _root_roomid in FORK_ROOMS.items()
+                    if _root_roomid == root_roomid
+                ]
+            }
+        )
 
     @web.authenticated
     @authorized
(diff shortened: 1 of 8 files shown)


projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
```diff
@@ -600,3 +602,95 @@ async def _cleanup_undo_manager(self, room_id: str) -> None:
...
+
+    @web.authenticated
+    @authorized
+    async def delete(self, fork_roomid):
+        """
+        Deletes a forked document, and optionally merges it back in the root document.
+        """
+        root_roomid = FORK_ROOMS[fork_roomid]
+        del FORK_ROOMS[fork_roomid]
+        if int(self.get_query_argument("merge")):
```
trungleduc, line 676: using `?merge=true` might be nicer than `?merge=1` ?

trungleduc:
Downstream users need to interact with the REST API and the Event handlers directly or will you introduce the JS APIs later?

projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
davidbrochart: Yes, done in https://github.com/jupyterlab/jupyter-collaboration/pull/382/commits/08276639c3d27d592f17fcf4f95c79291e33d242.

trungleduc (approved):

davidbrochart:
> Downstream users need to interact with the REST API and the Event handlers directly or will you introduce the JS APIs later?

Yes we can add higher-level JavaScript functions, maybe in another PR?

On a different note, I'm not sure if a fork's room ID should just be `uuid`, or `format:type:uuid`. Currently it's the former, because if it were the latter it would be interpreted as a stored document, but we don't want forks to be stored, do we?

trungleduc:
> Yes we can add higher-level JavaScript functions, maybe in another PR?

Yes, let's merge this one then I can work on this

davidbrochart:
> On a different note, I'm not sure if a fork's room ID should just be `uuid`, or `format:type:uuid`. Currently it's the former, because if it were the latter it would be interpreted as a stored document, but we don't want forks to be stored, do we?

We need to answer this question before merging.

trungleduc:
if it's only about changing the ID format, Isn't it safer to store the forks and we can remove them later if we don't have to?

davidbrochart:
But a stored document currently means that it goes through the Jupyter file ID service and that it is a file-backed document. So that would mean adding logic to manage (temporary) files for these forks.

davidbrochart:
If it's a stored document yes, but it's not at the moment.

trungleduc:
creating a file-less document where all data is stored in the `.jupyter_ystore.db` file could also be useful for the chat application?

davidbrochart:
Yes. But does it make sense to store suggestions?

trungleduc:
> Yes. But does it make sense to store suggestions?

Google Docs keeps the suggestions till we explicitly delete them so I say yes.

davidbrochart:
I'm thinking that we should maybe create a new branch to work on forking/suggestions, instead of taking the risk to block `main` or release completely WIP code.

davidbrochart:
But then it won't help you if you want to use the new features in JupyterCAD?

trungleduc:
> But then it won't help you if you want to use the new features in JupyterCAD?

there is no problem, I can link my work with your PR

jtpio:
Wouldn't the forks need be saved also in the case of server restarts? Users would probably expect their suggestions to be there if they resume their work later?

> But then it won't help you if you want to use the new features in JupyterCAD?

Maybe there could be some pre-releases made from the new branch? Or one could download the artifacts built by the releaser to use the latest version.

trungleduc:
@davidbrochart could you rebase this PR to bump its version to 3.0.0?

trungleduc (approved):

davidbrochart:
Superseded by #394.

PR #383, opened 2024-10-24
Add check for user value in CollaboratorsPanel

martinRenou:
Follow up https://github.com/jupyterlab/jupyter-collaboration/pull/367.

The server awareness has no user by default, to avoid seeing it as a collaborator if not required.


davidbrochart (approved):
LGTM, thanks.

brichet (approved):
Thanks @martinRenou

brichet:
@martinRenou I added a description.

PR #395, opened 2024-11-04
Add js api

trungleduc:
## Code Changes:

- New token  `IForkManagerToken` in `@jupyter/docprovider-extension`.
- This token implements the following interface:
```js
    /**
     * Creates a new fork for a given document.
     *
     * @param options.rootId - The ID of the root document to fork.
     * @param options.synchronize - A flag indicating whether the fork should be kept
     * synchronized with the root document.
     * @param options.title - An optional label for the fork.
     * @param options.description - An optional description for the fork.
     *
     * @returns A promise that resolves to an `IForkCreationResponse` if the fork
     * is created successfully, or `undefined` if the creation fails.
     */
    createFork(options: {
        rootId: string;
        synchronize: boolean;
        title?: string;
        description?: string;
    }): Promise<IForkCreationResponse | undefined>;
    /**
     * Retrieves all forks associated with a specific document.
     *
     * @param documentId - The ID of the document for which forks are to be retrieved.
     *
     * @returns A promise that resolves to an `IAllForksResponse` containing information about all forks.
     */
    getAllForks(documentId: string): Promise<IAllForksResponse>;
    /**
     * Deletes a specified fork and optionally merges its changes.
     *
     * @param options - Options for deleting the fork.
     * @param options.forkId - The ID of the fork to be deleted.
     * @param options.merge - A flag indicating whether changes from the fork should be merged back into the root document.
     *
     * @returns A promise that resolves when the fork is successfully deleted.
     */
    deleteFork(options: {
        forkId: string;
        merge: boolean;
    }): Promise<void>;
    /**
     * Signal emitted when a new fork is added.
     *
     * @event forkAdded
     * @type ISignal<IForkManager, IForkChangedEvent>
     */
    forkAdded: ISignal<IForkManager, IForkChangedEvent>;
    /**
     * Signal emitted when a fork is deleted.
     *
     * @event forkDeleted
     * @type ISignal<IForkManager, IForkChangedEvent>
     */
    forkDeleted: ISignal<IForkManager, IForkChangedEvent>;
```


packages/docprovider/src/requests.ts
```diff
@@ -36,6 +38,45 @@ export interface ISessionModel {
...
 }
 
+/**
+ * Call the API extension
+ *
+ * @param endPoint API REST end point for the extension
+ * @param init Initial values for the request
+ * @returns The response body interpreted as JSON
+ */
+export async function requestAPI<T = any>(
```
davidbrochart, line 48: I'm not sure about the name of this function. It's basically a wrapper around `ServerConnection.makeRequest`, so should it be called `makeRequest` too?
Also, it seems that it could be used by other functions in this file, although we could do that in another PR.

davidbrochart (approved):
Thanks so much @trungleduc, this is great.
I just have a minor comment.

packages/docprovider/src/requests.ts
trungleduc: I copied this helper from other Jupyter projects

davidbrochart: Should it be part of JupyterLab then?

trungleduc: I don't know, this function is also defined in the extension template https://github.com/jupyterlab/extension-template/blob/5696924f5b7adf07cf07ea9d3510c66914fbb770/template/src/%7B%25%20if%20kind%20%3D%3D%20'server'%20%25%7Dhandler.ts%7B%25%20endif%20%25%7D.jinja#L12

Issue #380, opened 2024-10-21
Remove the extra `jlpm run build` in the workflow


jtpio:
An extra `jlpm run build` was added to the CI workflow in https://github.com/jupyterlab/jupyter-collaboration/pull/353:

https://github.com/jupyterlab/jupyter-collaboration/blob/c52fa207510ab6062e3ca9400cc4e71d9676c87c/.github/workflows/test.yml#L74

Although it was not clear why this was needed, as discussed here: https://github.com/jupyterlab/jupyter-collaboration/pull/353#pullrequestreview-2357605672

It would be great to check if it could be removed. Maybe it's hiding another issue of how the repo is currently set up.

cc @davidbrochart @krassowski

trungleduc:
> It would be great to check if it could be removed. Maybe it's hiding another issue of how the repo is currently set up.

it can not be removed because of the current setup. The lab extension source is not a part of the python package so calling `pip install ` in the python package does not build the lab extension.

PR #396, opened 2024-11-05
Update CI

trungleduc:
## Code changes

- Reorganize test workflow

Screenshot 2024-11-05 144138

- Drop python 3.8
- Remove uneeded steps in the pre-commit job
- Run UI test with packages generated from current PR, not from PYPI.
- Relax Playwright UI test condition.
- Add dev install script: calling `python scripts/dev_install.py` or ` yarn dev` is enough to set up the dev enviroment. 
- Add update snapshot workflow


trungleduc:
@jtpio I'm getting this error in the **Check Release** job: `ValueError: Tag v3.0.0 already exists`. Do you know what happened?

jtpio:
Looks like it may be related to the changes to the `yarn.lock` file?

It seems it's not able to bump the version to `3.0.1` in the workflow, which could explain the existing tag issue:

image

jtpio:
Probably not related but something that should also be fixed on CI: https://github.com/jupyterlab/jupyter-collaboration/issues/380

trungleduc:
> Looks like it may be related to the changes to the `yarn.lock` file?
> 
> It seems it's not able to bump the version to `3.0.1` in the workflow, which could explain the existing tag issue:
> 
> image

Thanks!

trungleduc:
> Probably not related but something that should also be fixed on CI: #380

Ok i will take care of this

trungleduc:
Failed UI test looks weird but i'm not sure it relates to this PR

- Actual

initialization-open-notebook-guest-actual

- Expected

initialization-open-notebook-guest-expected

davidbrochart (approved):
Thanks @trungleduc, that's great.
Since you dropped Python 3.8 in the CI, do you want to update the dependency [in pyproject.toml](https://github.com/jupyterlab/jupyter-collaboration/blob/b3d8d29e3ea6e9cf3c56bada86b15b612355339a/pyproject.toml#L10) too?
Also, can you reproduce the new lines in the cell stream output locally?

trungleduc:
> do you want to update the dependency in pyproject.toml too?

isn't it a breaking change? I thought that `main` is targeting 3.1.x

davidbrochart:
I don't think we should make a major release for that. Python 3.8 has reached end of life, I think all Jupyter projects should drop support for it too.

jtpio:
Right it's probably fine to drop support for a Python version in a minor release. For example JupyterLab dropped support for Python 3.6 in JupyterLab 3.3: https://jupyterlab.readthedocs.io/en/latest/getting_started/changelog.html#id298

As long as it's visible in the changelog and/or the docs.

jtpio:
So the UI tests seem to be failing consistently after a restart. It would be worth looking into it as otherwise we would have to live with a failing CI, which is going to be annoying very quickly.

trungleduc:
@jtpio, how should we proceed with this PR? Should we wait for a new JupyterLab patch version or update the snapshot?

trungleduc:
> Does this behavior happen for real users of `jupyter-collaboration`, not just in the UI tests?
> 
> If so, it will probably make sense to update to the "wrong" snapshot for now to indicate this is an expected issue. And then fix it when a new JupyterLab release with the fix is available?

I can reproduce it locally https://github.com/jupyterlab/jupyterlab/issues/16934

jtpio:
OK then let's update the snapshot so the PR can be merged?

Thanks @trungleduc for looking into this!

trungleduc:
it's green!

PR #399, opened 2024-11-07
Update CI (#396)

trungleduc:
Backport #396


davidbrochart:
Should we just rebase the `suggestions` branch on `main` instead of creating a PR?

trungleduc:
> Should we just rebase the `suggestions` branch on `main` instead of creating a PR?

I don't have the right to rebase this branch. It would be great if you could set up the backport bot.

davidbrochart:
But even if we set up a backport bot, we will end up with a bunch of commits that are different from `main`, right?

trungleduc:
yep, but at least I don't need to ask you to rebase this branch every time.

trungleduc:
@davidbrochart let's merge this?

jtpio:
@meeseeksdev hello

(to test if the bot is enabled here)

davidbrochart:
> @davidbrochart let's merge this?

I'm just a bit worried about when we merge branch `suggestions` into `main`. Won't we get a bunch of conflicts?

trungleduc:
> > @davidbrochart let's merge this?
> 
> I'm just a bit worried about when we merge branch `suggestions` into `main`. Won't we get a bunch of conflicts?

Theoretically no since we resolve conflicts on each backport PR.

davidbrochart:
I agree that the diff should only be our code, but I think the backport PRs will still be seen as conflicts.
Anyway, let's see with this one as you said :+1:

Issue #392, opened 2024-11-01
Timeline slider is broken when using "Open from path"


krassowski:
## Description

```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'split')
    at n (544.20c789155db30fb50597.js?v=20c789155db30fb50597:1:2052)
    at 544.20c789155db30fb50597.js?v=20c789155db30fb50597:1:2494
```

## Reproduce

1. Launch https://gist.github.com/krassowski/d4f0030f95af9f63c1dea288fc7e00dd
2. Create a notebook
3. Go to File menu → Open From Path
4. Enter "Untitled.ipynb"
5. See error in console:

```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'split')
    at n (544.20c789155db30fb50597.js?v=20c789155db30fb50597:1:2052)
    at 544.20c789155db30fb50597.js?v=20c789155db30fb50597:1:2494
```

This comes from:

https://github.com/jupyterlab/jupyter-collaboration/blob/20a3735e2e4fc81de5d94f1ea88ff16240d58779/packages/docprovider-extension/src/filebrowser.ts#L157-L170

of note, `documentPath` slicing result is wrong in this case too

## Expected behavior

No error

## Context

- Browser and version: Chrome
- JupyterLab version: 4.2.x and 4.3.0

PR #401, opened 2024-11-18
Fix open path error with timeline feature

Meriem-BenIsmail:
This PR fixes https://github.com/jupyterlab/jupyter-collaboration/issues/392

### DIFF
--- a/packages/docprovider-extension/src/filebrowser.ts
+++ b/packages/docprovider-extension/src/filebrowser.ts
@@ -158,35 +158,37 @@ export const statusBarTimeline: JupyterFrontEndPlugin<void> = {
         documentPath: string,
         documentId: string
       ) => {
-        if (drive) {
-          // Remove 'RTC:' from document path
-          documentPath = documentPath.slice(drive.name.length + 1);
-          // Dispose of the previous timelineWidget if it exists
-          if (timelineWidget) {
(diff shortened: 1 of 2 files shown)


davidbrochart:
I just rebases on `main`.

Meriem-BenIsmail:
I have a clarification request concerning this PR for fixing the https://github.com/jupyterlab/jupyter-collaboration/issues/392 issue. 
If we choose to open a notebook with the 'Open from Path' option, it doesn't open the notebook in collaborative mode (without the RTC prefix) so the timeline slider (also the history icon) shouldn't even be visible if it's not collaborative. That's what I did in this PR so that the history icon wouldn't be shown and the error described in this issue https://github.com/jupyterlab/jupyter-collaboration/issues/392 wouldn't occur. I wanted to check if that's the expected behavior that the timeline should have.

davidbrochart:
I agree that if we "Open from Path" e.g. "Untitled.ipynb", the file should not be opened in collaborative mode, and thus no timeline should be available. But if we "Open from Path" e.g. "RTC:Untitled.ipynb", the file should be opened in collaborative mode, and the timeline should be available.

davidbrochart (approved):
Thanks @Meriem-BenIsmail.

Issue #393, opened 2024-11-01
Timeline slider not shown when base_url is set


ykazakov:
# See below for an [updated report](https://github.com/jupyterlab/jupyter-collaboration/issues/393#issuecomment-2462517066)

## OLD Description

When Jupyter Lab is started as as single-user server in Jupyter Hub, the document timeline widget uses an incorrect url, which can be seen from the browser console when clicking on the timeline icon:

<img width="924" alt="Screenshot 2024-11-01 at 22 23 53" src="https://github.com/user-attachments/assets/c23cddf6-dab4-49cb-a0ea-fcbbd308f055">

As can be seen from this screenshot, the url has the part `/collaboration/timeline` two times.

If I  manually change the url by removing one duplicate and open it in a browser, it appears to be working fine:
```
{"roomId": "json:notebook:1acd8c55-cfea-4e70-aa54-f7ad73df5511", "timestamps": [1730491695.928937], "forkRoom": "426ef1f297924abfb3af89c85f1ccf82", "sessionId": "cff41fa8-722f-4dc2-b52b-07db54815389"}
```
## Reproduce

1. Install the required packages in a fresh python environment: `conda install -c conda-forge jupyterhub jupyterlab jupyter-collaboration`
2. Start jupyterhub: `jupyterhub --JupyterHub.authenticator_class='dummy' --JupyterHub.spawner_class='simple'`
3. Open the url `http://localhost:8000` and enter any user name, e.g., `bob`
4. Open a new notebook
5. Open the browser console and click on the timeline icon
6. Observe the above error message in the console

## Expected behavior

A slider should appear next to the timeline icon.

## Context

- Operating System and version: MacOS 15.0.1
- Browser and version: Safari 18.0.1
- JupyterLab version: 4.3.0 
```
% pip list | grep jupyter
jupyter_client            8.6.3
jupyter-collaboration     3.0.0
jupyter-collaboration-ui  1.0.0
jupyter_core              5.7.2
jupyter-docprovider       1.0.0
jupyter-events            0.10.0
jupyter-lsp               2.2.5
jupyter_server            2.14.2
jupyter_server_fileid     0.9.2
jupyter_server_terminals  0.5.3
jupyter-server-ydoc       1.0.0
jupyter-ydoc              3.0.0
jupyterhub                5.2.1
jupyterlab                4.3.0
jupyterlab_pygments       0.3.0
jupyterlab_server         2.27.3
```

[...]

PR #402, opened 2024-11-18
Fix timeline feature error when base url is set

Meriem-BenIsmail:
This PR fixes https://github.com/jupyterlab/jupyter-collaboration/issues/393


davidbrochart:
Failures seem to be unrelated (see #403).
I will look at that.

davidbrochart (commented):
Since #393 seems to suggest an issue with the server base URL, should you make use of it in this PR? The `baseUrl` is available in JupyterLab's page config.

Meriem-BenIsmail:
> Since #393 seems to suggest an issue with the server base URL, should you make use of it in this PR? The `baseUrl` is available in JupyterLab's page config.

The problem isn't actually with the baseUrl, it's just that the document name was being incorrectly extracted.

davidbrochart:
Could you write a test that shows that this was indeed the cause of problem? In particular, I'd like to see that your fix works even if "api/collaboration/timeline" is present in the base URL.

Meriem-BenIsmail:
Currently, the updated logic ensures that only the last occurrence of `api/collaboration/timeline` is picked up, so it should handle the base URL scenario as expected. Given this, the fix seems okay for the mentioned issue.

Would you still like me to add a specific test to verify this behavior?

I also want to mention that there’s another open PR ([#401](https://github.com/jupyterlab/jupyter-collaboration/pull/401)) that creates a test file for the timeline slider. If I add a test here, it would create or modify the same file, potentially leading to conflicts later. Let me know how you'd like to proceed—whether to hold off on tests for this PR until #401 is merged or go ahead and add them now.

packages/docprovider/src/component.tsx
```diff
@@ -20,6 +20,7 @@ type Props = {
   contentType: string;
   format: string;
 };
+const DOCUMENT_TIMELINE_URL = 'api/collaboration/timeline';
```
davidbrochart, line 23: We already have it [here](https://github.com/jupyterlab/jupyter-collaboration/blob/3126a81b78174cbc32c5f05d4678458f2d01f3eb/packages/docprovider/src/requests.ts#L15) and [here](https://github.com/jupyterlab/jupyter-collaboration/blob/3126a81b78174cbc32c5f05d4678458f2d01f3eb/packages/docprovider-extension/src/filebrowser.ts#L46). Should it be in one place?

davidbrochart:
Yes it would be nice to have a test. I merged #401.

packages/docprovider/src/component.tsx
Meriem-BenIsmail: Yes I will fix that. Thanks for the feedback !

ui-tests/package.json
```diff
@@ -5,7 +5,9 @@
   "private": true,
   "scripts": {
     "start": "jupyter lab --config jupyter_server_test_config.py",
-    "test": "npx playwright test",
+    "start:timeline": "jupyter lab --config jupyter_server_test_config_timeline.py",
```
davidbrochart, line 8: You can get rid of `ui-tests/jupyter_server_test_config_timeline.py` and do:
```suggestion
    "start:timeline": "jupyter lab --ServerApp.base_url=/api/collaboration/timeline/",
```

ui-tests/tests/timeline-slider.spec.ts
```diff
@@ -23,17 +23,51 @@ async function openNotebook(page: Page, notebookPath: string) {
...
 
-  test('should fail if there are console errors', async ({ page, tmpPath }) => {
+const isTimelineEnv = process.env.TIMELINE_FEATURE || "0";
+const isTimeline = parseInt(isTimelineEnv)
+
+test.describe('Timeline Slider', () => {
+
+  isTimeline && test.use({ autoGoto: false })
+
+  test('should fail if there are console errors when opening from path', async ({ page, tmpPath }) => {
```
davidbrochart, line 34: I don't understand what this test does. It seems that you do nothing if `isTimeline`?

ui-tests/tests/timeline-slider.spec.ts
```diff
@@ -23,17 +23,51 @@ async function openNotebook(page: Page, notebookPath: string) {
...
-    await page.notebook.save();
     await page.notebook.close();
 
+    if(!isTimeline)
     await openNotebook(page, `${tmpPath}/Untitled.ipynb`);
 
     expect(pageErrors).toHaveLength(0);
   });
+
+  test('should display in status bar without console errors when baseUrl is set', async ({ page, baseURL }) => {
```
davidbrochart, line 50: Same comment as above.

ui-tests/tests/timeline-slider.spec.ts
Meriem-BenIsmail: There is no need to execute this test in the added timeline test environment (with the set baseurl)
but this is the test for this PR that was merged https://github.com/jupyterlab/jupyter-collaboration/pull/401

ui-tests/tests/timeline-slider.spec.ts
```diff
@@ -23,17 +23,51 @@ async function openNotebook(page: Page, notebookPath: string) {
...
+      console.log('Skipping this test.');
+      return;
+    }
     const pageErrors = await capturePageErrors(page);
 
     await page.notebook.createNew();
-    await page.notebook.save();
     await page.notebook.close();
 
+    if(!isTimeline)
```
davidbrochart, line 44:
```suggestion
```

ui-tests/tests/timeline-slider.spec.ts
```diff
@@ -23,17 +23,51 @@ async function openNotebook(page: Page, notebookPath: string) {
...
 
-test.describe('Open from Path', () => {
 
-  test('should fail if there are console errors', async ({ page, tmpPath }) => {
+const isTimelineEnv = process.env.TIMELINE_FEATURE || "0";
+const isTimeline = parseInt(isTimelineEnv)
+
+test.describe('Timeline Slider', () => {
+
+  isTimeline && test.use({ autoGoto: false })
```
davidbrochart, line 32: What is it for?

Meriem-BenIsmail: So I found that even with different configuration, somehow playwright overwrites the baseurl and goes to the default one (`http://localhost:8888`) by default so we need to use this `test.use({ autoGoto: false })`

Meriem-BenIsmail: isTimeline is the environment variable used to determine whether we are testing with the new environment or the default one. So we need to do this `test.use({ autoGoto: false })` only if we are testing with the new env (when `isTimeline` is true like in here `TIMELINE_FEATURE=1 jlpm test:timeline` )

davidbrochart: Could you explain what you mean by "new environment" and "default environment"?

Meriem-BenIsmail: The "new environment" refers to the testing setup, which uses a custom `baseUrl` (`/api/collaboration/timeline/`) as opposed to the "default environment," where tests run with the standard `baseUrl` (`/lab` or `/`). The new environment is specifically configured to validate the Timeline functionality in isolation, whereas the default environment is used for general tests without the Timeline-specific context.

`isTimeline` is used to identify which environment is active. If `isTimeline` is true, we disable the automatic navigation (`test.use({ autoGoto: false })`) that Playwright performs, as this navigation doesn’t respect the custom `baseUrl `and defaults to `http://localhost:8888`. Disabling it allows us to manually control the navigation to the appropriate custom environment.

davidbrochart (approved):
Thanls @Meriem-BenIsmail.

PR #412, opened 2024-12-02
Update with jupyter-ydoc v3.0.2

davidbrochart:
This should use jupyter-ydoc [v3.0.2](https://github.com/jupyter-server/jupyter_ydoc/releases/tag/v3.0.2), and UI tests should fail because we updated the snapshots [here](https://github.com/jupyterlab/jupyter-collaboration/pull/396#issuecomment-2457755251).


trungleduc:
please update snapshots

krassowski:
It looks like the tests did not fail here even though 3.0.2 was installed?

I see jupyter_ydoc-3.0.2-py3-none-any.whl is used.

davidbrochart:
Indeed, I don't understand.

davidbrochart:
Oh but running UI tests locally I see that the text is rendered in one line, maybe we changed the snapshots again after #396?

trungleduc:
it looks ok now

trungleduc:
please update snapshots

davidbrochart:
> it looks ok now

But there are 4 failed tests now?

davidbrochart:
It's hard to believe that lowering `maxDiffPixelRatio` makes the test fail, otherwise we were not testing anything before? The difference was huge, because we had a text in one column instead of one line.

davidbrochart:
The updated snapshots look good:
image

But isn't it revealing an issue in the tests themselves?

trungleduc:
> It's hard to believe that lowering `maxDiffPixelRatio` makes the test fail, otherwise we were not testing anything before? The difference was huge, because we had a text in one column instead of one line.

the diff is computed by the number of changed pixels, so maybe for the text, the diff is big but not too much in terms of pixel number.

trungleduc:
> Fair enough, but then tests should likely not have failed in #396 too.

yeah, flakiness is the beauty of the UI tests.

krassowski:
What do you think about excluding 3.0.0 and 3.0.1 from dependencies using `!=` in:

https://github.com/jupyterlab/jupyter-collaboration/blob/ed544ba982f55d1fd4a28d5124fbddc3043bbf89/projects/jupyter-server-ydoc/pyproject.toml#L32

davidbrochart:
Good point, let's do it.

Issue #202, opened 2023-10-13
RTC breaks notebook links


thesuperzapper:
## Description

When you click the "notebook" button, you get a 404:

- ![Screenshot 2023-10-12 at 21 05 06](https://github.com/jupyterlab/jupyter-collaboration/assets/5735406/234203d0-528e-4733-95e5-5b2dc88f6b16)

This is because of the extra `"RTC:"` which is added to the file path (which is not actually there):

- ![Screenshot 2023-10-12 at 21 05 59](https://github.com/jupyterlab/jupyter-collaboration/assets/5735406/1d080ae6-c569-4730-a9a2-72786624105c)

## Reproduce

1. Ensure `pip install notebook` is installed
2. Try clicking the `notebook` button
3. See 404

jrouly:
This issue also impacts the [`voila-dashboards`](https://github.com/voila-dashboards) extension. The URL seems to prepend the filename with `RTC:` which looks like it breaks a number of expectations.

jochym:
This also affects jupyterlab-latex (https://github.com/jupyterlab/jupyterlab-latex/issues/224) for the same reason. It is even worse - the RTC is present even if the RTC is disabled with `--YDocExtension.disable_rtc=True` while the extension is enabled (just RTC is disabled). 
Please treat it as high priority - it renders multiple use cases impossible.

jrouly:
Based on discussion in https://github.com/jupyterlab/jupyter-collaboration/issues/183 I suspect now that the downstream plugins breaking may be considered a feature, not a bug.

However the original issue report here as to a broken feature within JupyterLab itself is pretty damning.

Podolyakofs:
Also this affects [jupyterlab-h5web#121](https://github.com/silx-kit/jupyterlab-h5web/issues/121), same file path error

frankebersoll:
This also affects [jupyterlab_rise](https://github.com/jupyterlab-contrib/rise): Slideshows won't work when RTC is enabled because of invalid links, you get 404 all over the place. Same for "File > Save and export notebook as..."

betolink:
I ran into the same issue with https://github.com/silx-kit/jupyterlab-h5web/issues/121  Is this a problem with the extensions?

@fcollonval @davidbrochart

davidbrochart:
> Is this a problem with the extensions?

Yes it probably doesn't support custom drives.

krassowski:
[...]

PR #337, opened 2024-08-18
Fix spurious "File Changed" dialog in JupyterLab 4.1+/Notebook 7.1+

krassowski:
- fixes https://github.com/jupyterlab/jupyter-collaboration/issues/312
- fixes https://github.com/jupyterlab/jupyter-collaboration/issues/313
- fixes https://github.com/jupyterlab/jupyter-collaboration/issues/317
- requires a PR to `jupyter_ydoc` to add `hash` setter https://github.com/jupyter-server/jupyter_ydoc/pull/262
- requires to set the hash on frontend; we can emit `fileChanged` signal from drive and this gets passed down to the context which stores the hash; this requires https://github.com/jupyterlab/jupyterlab/pull/16695

PR #418, opened 2024-12-16
Use content providers to remove RTC prefix

krassowski:
- Replaces https://github.com/jupyterlab/jupyter-collaboration/pull/345 (which has quote a few conflicts)
- Fixes https://github.com/jupyterlab/jupyter-collaboration/issues/202

We still need to decide what to do where `RTC:` prefix was used in the actual logic.

### DIFF
--- a/packages/collaborative-drive/src/tokens.ts
+++ b/packages/collaborative-drive/src/tokens.ts
@@ -1,8 +1,8 @@
 // Copyright (c) Jupyter Development Team.
 // Distributed under the terms of the Modified BSD License.
 
-import { DocumentChange, IAwareness, YDocument } from '@jupyter/ydoc';
-import { Contents } from '@jupyterlab/services';
+import { IAwareness } from '@jupyter/ydoc';
+import { Contents, SharedDocumentFactory } from '@jupyterlab/services';
 import { IDisposable } from '@lumino/disposable';
 
 import { Token } from '@lumino/coreutils';
@@ -10,9 +10,10 @@ import { Token } from '@lumino/coreutils';
 /**
  * The collaborative drive.
  */
-export const ICollaborativeDrive = new Token<ICollaborativeDrive>(
-  '@jupyter/collaboration-extension:ICollaborativeDrive'
-);
+export const ICollaborativeContentProvider =
+  new Token<ICollaborativeContentProvider>(
+    '@jupyter/collaboration-extension:ICollaborativeContentProvider'
+  );
 
 /**
  * The global awareness token.
@@ -21,18 +22,7 @@ export const IGlobalAwareness = new Token<IAwareness>(
   '@jupyter/collaboration:IGlobalAwareness'
 );
 
-/**
- * A document factory for registering shared models
- */
-export type SharedDocumentFactory = (
-  options: Contents.ISharedFactoryOptions
-) => YDocument<DocumentChange>;
-
-/**
- * A Collaborative implementation for an `IDrive`, talking to the
- * server using the Jupyter REST API and a WebSocket connection.
- */
-export interface ICollaborativeDrive extends Contents.IDrive {
+export interface ICollaborativeContentProvider {
   /**
    * SharedModel factory for the YDrive.
    */
--- a/packages/docprovider-extension/src/filebrowser.ts
+++ b/packages/docprovider-extension/src/filebrowser.ts
@@ -4,45 +4,44 @@
  */
 
 import {
-  ILabShell,
-  IRouter,
   JupyterFrontEnd,
   JupyterFrontEndPlugin
 } from '@jupyterlab/application';
 import { Dialog, showDialog } from '@jupyterlab/apputils';
 import { DocumentWidget, IDocumentWidget } from '@jupyterlab/docregistry';
 import { Widget } from '@lumino/widgets';
-import {
-  FileBrowser,
-  IDefaultFileBrowser,
-  IFileBrowserFactory
-} from '@jupyterlab/filebrowser';
-import { IStatusBar } from '@jupyterlab/statusbar';
 
-import { IEditorTracker } from '@jupyterlab/fileeditor';
+import { IStatusBar } from '@jupyterlab/statusbar';
+import { ContentsManager } from '@jupyterlab/services';
+
+import {
+  IEditorTracker,
+  IEditorWidgetFactory,
+  FileEditorFactory
+} from '@jupyterlab/fileeditor';
 import { ILogger, ILoggerRegistry } from '@jupyterlab/logconsole';
-import { INotebookTracker } from '@jupyterlab/notebook';
+import {
+  INotebookTracker,
+  INotebookWidgetFactory,
+  NotebookWidgetFactory
+} from '@jupyterlab/notebook';
 import { ISettingRegistry } from '@jupyterlab/settingregistry';
 import { ITranslator, nullTranslator } from '@jupyterlab/translation';
-
-import { CommandRegistry } from '@lumino/commands';
 
 import { YFile, YNotebook } from '@jupyter/ydoc';
 
 import {
-  ICollaborativeDrive,
+  ICollaborativeContentProvider,
   IGlobalAwareness
 } from '@jupyter/collaborative-drive';
-import { IForkProvider, TimelineWidget, YDrive } from '@jupyter/docprovider';
+import {
+  IForkProvider,
+  TimelineWidget,
+  RtcContentProvider
+} from '@jupyter/docprovider';
 import { Awareness } from 'y-protocols/awareness';
 import { URLExt } from '@jupyterlab/coreutils';
 
-/**
- * The command IDs used by the file browser plugin.
- */
-namespace CommandIDs {
-  export const openPath = 'filebrowser:open-path';
-}
 const DOCUMENT_TIMELINE_URL = 'api/collaboration/timeline';
 
 const TWO_SESSIONS_WARNING =
@@ -50,26 +49,43 @@ const TWO_SESSIONS_WARNING =
   'This is not supported. Please close this view; otherwise, ' +
   'some of your edits may not be saved properly.';
 
-/**
- * The default collaborative drive provider.
- */
-export const drive: JupyterFrontEndPlugin<ICollaborativeDrive> = {
-  id: '@jupyter/docprovider-extension:drive',
-  description: 'The default collaborative drive provider',
-  provides: ICollaborativeDrive,
-  requires: [ITranslator],
-  optional: [IGlobalAwareness],
-  activate: (
-    app: JupyterFrontEnd,
-    translator: ITranslator,
-    globalAwareness: Awareness | null
-  ): ICollaborativeDrive => {
-    const trans = translator.load('jupyter_collaboration');
-    const drive = new YDrive(app.serviceManager.user, trans, globalAwareness);
-    app.serviceManager.contents.addDrive(drive);
-    return drive;
-  }
-};
+export const rtcContentProvider: JupyterFrontEndPlugin<ICollaborativeContentProvider> =
+  {
+    id: '@jupyter/docprovider-extension:content-provider',
+    description: 'The RTC content provider',
+    provides: ICollaborativeContentProvider,
+    requires: [ITranslator],
+    optional: [IGlobalAwareness],
+    activate: (
+      app: JupyterFrontEnd,
+      translator: ITranslator,
+      globalAwareness: Awareness | null
+    ): ICollaborativeContentProvider => {
+      const trans = translator.load('jupyter_collaboration');
+      const defaultDrive = (app.serviceManager.contents as ContentsManager)
+        .defaultDrive;
+      if (!defaultDrive) {
+        throw Error(
+          'Cannot initialize content provider: default drive property not accessible on contents manager instance.'
+        );
+      }
+      const registry = defaultDrive.contentProviderRegistry;
+      if (!registry) {
+        throw Error(
+          'Cannot initialize content provider: no content provider registry.'
+        );
+      }
+      const rtcContentProvider = new RtcContentProvider({
+        apiEndpoint: '/api/contents',
+        serverSettings: defaultDrive.serverSettings,
+        user: app.serviceManager.user,
+        trans,
+        globalAwareness
+      });
+      registry.register('rtc', rtcContentProvider);
+      return rtcContentProvider;
+    }
+  };
 
 /**
  * Plugin to register the shared model factory for the content type 'file'.
@@ -79,13 +95,20 @@ export const yfile: JupyterFrontEndPlugin<void> = {
   description:
     "Plugin to register the shared model factory for the content type 'file'",
   autoStart: true,
-  requires: [ICollaborativeDrive],
-  optional: [],
-  activate: (app: JupyterFrontEnd, drive: ICollaborativeDrive): void => {
+  requires: [ICollaborativeContentProvider, IEditorWidgetFactory],
+  activate: (
+    app: JupyterFrontEnd,
+    contentProvider: ICollaborativeContentProvider,
+    editorFactory: FileEditorFactory.IFactory
+  ): void => {
     const yFileFactory = () => {
       return new YFile();
     };
-    drive.sharedModelFactory.registerDocumentFactory('file', yFileFactory);
+    contentProvider.sharedModelFactory.registerDocumentFactory(
+      'file',
+      yFileFactory
+    );
+    editorFactory.contentProviderId = 'rtc';
   }
 };
 
@@ -97,11 +120,12 @@ export const ynotebook: JupyterFrontEndPlugin<void> = {
   description:
     "Plugin to register the shared model factory for the content type 'notebook'",
   autoStart: true,
-  requires: [ICollaborativeDrive],
+  requires: [ICollaborativeContentProvider, INotebookWidgetFactory],
   optional: [ISettingRegistry],
   activate: (
     app: JupyterFrontEnd,
-    drive: YDrive,
+    contentProvider: ICollaborativeContentProvider,
+    notebookFactory: NotebookWidgetFactory.IFactory,
     settingRegistry: ISettingRegistry | null
   ): void => {
     let disableDocumentWideUndoRedo = true;
@@ -131,10 +155,11 @@ export const ynotebook: JupyterFrontEndPlugin<void> = {
         disableDocumentWideUndoRedo
       });
     };
-    drive.sharedModelFactory.registerDocumentFactory(
+    contentProvider.sharedModelFactory.registerDocumentFactory(
       'notebook',
       yNotebookFactory
     );
+    notebookFactory.contentProviderId = 'rtc';
   }
 };
 /**
@@ -144,11 +169,11 @@ export const statusBarTimeline: JupyterFrontEndPlugin<void> = {
   id: '@jupyter/docprovider-extension:statusBarTimeline',
   description: 'Plugin to add a timeline slider to the status bar',
   autoStart: true,
-  requires: [IStatusBar, ICollaborativeDrive],
+  requires: [IStatusBar, ICollaborativeContentProvider],
   activate: async (
     app: JupyterFrontEnd,
     statusBar: IStatusBar,
-    drive: ICollaborativeDrive
+    contentProvider: ICollaborativeContentProvider
   ): Promise<void> => {
     try {
       let sliderItem: Widget | null = null;
@@ -158,10 +183,9 @@ export const statusBarTimeline: JupyterFrontEndPlugin<void> = {
         documentPath: string,
         documentId: string
       ) => {
-        if (documentId && documentPath.split(':')[0] === 'RTC') {
-          if (drive) {
-            // Remove 'RTC:' from document path
-            documentPath = documentPath.slice(drive.name.length + 1);
+        // TODO: check if documentId uses RTC
+        if (documentId) {
+          if (contentProvider) {
             // Dispose of the previous timelineWidget if it exists
             if (timelineWidget) {
               timelineWidget.dispose();
@@ -169,7 +193,7 @@ export const statusBarTimeline: JupyterFrontEndPlugin<void> = {
             }
 
             const [format, type] = documentId.split(':');
-            const provider = drive.providers.get(
+            const provider = contentProvider.providers.get(
               `${format}:${type}:${documentPath}`
             ) as unknown as IForkProvider;
             const fullPath = URLExt.join(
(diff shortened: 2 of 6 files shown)


packages/docprovider/src/ydrive.ts
```diff
@@ -31,55 +37,43 @@ export interface IForkProvider {
...
+    this._globalAwareness = options.globalAwareness;
+    this._serverSettings = options.serverSettings;
     this.sharedModelFactory = new SharedModelFactory(this._onCreate);
+    this._providers = new Map<string, WebSocketProvider>();
+    /*
     super.fileChanged.connect((_, change) => {
       // pass through any events from the Drive superclass
       this._ydriveFileChanged.emit(change);
     });
+    */
```
krassowski, line 65: I am pretty sure this is no longer needed, but wanted to double check before removal:

```suggestion
```

krassowski:
CI is expected to fail until https://github.com/jupyterlab/jupyterlab/pull/17092 is merged and released.

davidbrochart:
I installed https://github.com/jupyterlab/jupyterlab/pull/17092 and then this PR, but I have an error:
```
> @jupyter/collaborative-drive:"install:extension"
src/tokens.ts(5,20): error TS2305: Module '"@jupyterlab/services"' has no exported member 'SharedDocumentFactory'.
```
Am I missing something?

krassowski:
`SharedDocumentFactory` is definitely exported in https://github.com/jupyterlab/jupyterlab/pull/17092.

Did you link the packages from https://github.com/jupyterlab/jupyterlab/pull/17092? I used:

```
jlpm link /path-to-jupyterlab-packages/ --all
jlpm
jlpm dlx yarn-berry-deduplicate
```

I had to add `"skipLibCheck": true,` to silence some react version warnings and added some resolutions to pin versions, see below:

<details>

```diff
diff --git a/package.json b/package.json
index 8520d4e..6860e66 100644
--- a/package.json
+++ b/package.json
@@ -66,5 +66,116 @@
     "stylelint-prettier": "^3.0.0",
     "typedoc": "~0.23.28",
     "typescript": "~5.0.4"
+  },
+  "resolutions": {
+    "@jupyterlab/services": "portal:/home/krassowski/jupyterlab/packages/services",
+    "@jupyterlab/docmanager": "portal:/home/krassowski/jupyterlab/packages/docmanager",
+    "@jupyterlab/builder": "portal:/home/krassowski/jupyterlab/builder",
+    "@jupyterlab/buildutils": "portal:/home/krassowski/jupyterlab/buildutils",
+    "@jupyterlab/template": "portal:/home/krassowski/jupyterlab/buildutils/template",
+    "@jupyterlab/galata": "portal:/home/krassowski/jupyterlab/galata",
+    "@jupyterlab/application": "portal:/home/krassowski/jupyterlab/packages/application",
+    "@jupyterlab/application-extension": "portal:/home/krassowski/jupyterlab/packages/application-extension",
+    "@jupyterlab/apputils": "portal:/home/krassowski/jupyterlab/packages/apputils",
+    "@jupyterlab/apputils-extension": "portal:/home/krassowski/jupyterlab/packages/apputils-extension",
+    "@jupyterlab/attachments": "portal:/home/krassowski/jupyterlab/packages/attachments",
+    "@jupyterlab/cell-toolbar": "portal:/home/krassowski/jupyterlab/packages/cell-toolbar",
+    "@jupyterlab/cell-toolbar-extension": "portal:/home/krassowski/jupyterlab/packages/cell-toolbar-extension",
+    "@jupyterlab/cells": "portal:/home/krassowski/jupyterlab/packages/cells",
+    "@jupyterlab/celltags-extension": "portal:/home/krassowski/jupyterlab/packages/celltags-extension",
+    "@jupyterlab/codeeditor": "portal:/home/krassowski/jupyterlab/packages/codeeditor",
+    "@jupyterlab/codemirror": "portal:/home/krassowski/jupyterlab/packages/codemirror",
+    "@jupyterlab/codemirror-extension": "portal:/home/krassowski/jupyterlab/packages/codemirror-extension",
+    "@jupyterlab/completer": "portal:/home/krassowski/jupyterlab/packages/completer",
+    "@jupyterlab/completer-extension": "portal:/home/krassowski/jupyterlab/packages/completer-extension",
+    "@jupyterlab/console": "portal:/home/krassowski/jupyterlab/packages/console",
+    "@jupyterlab/console-extension": "portal:/home/krassowski/jupyterlab/packages/console-extension",
+    "@jupyterlab/coreutils": "portal:/home/krassowski/jupyterlab/packages/coreutils",
+    "@jupyterlab/csvviewer": "portal:/home/krassowski/jupyterlab/packages/csvviewer",
+    "@jupyterlab/csvviewer-extension": "portal:/home/krassowski/jupyterlab/packages/csvviewer-extension",
+    "@jupyterlab/debugger": "portal:/home/krassowski/jupyterlab/packages/debugger",
+    "@jupyterlab/debugger-extension": "portal:/home/krassowski/jupyterlab/packages/debugger-extension",
+    "@jupyterlab/docmanager-extension": "portal:/home/krassowski/jupyterlab/packages/docmanager-extension",
+    "@jupyterlab/docregistry": "portal:/home/krassowski/jupyterlab/packages/docregistry",
+    "@jupyterlab/documentsearch": "portal:/home/krassowski/jupyterlab/packages/documentsearch",
+    "@jupyterlab/documentsearch-extension": "portal:/home/krassowski/jupyterlab/packages/documentsearch-extension",
+    "@jupyterlab/extensionmanager": "portal:/home/krassowski/jupyterlab/packages/extensionmanager",
+    "@jupyterlab/extensionmanager-extension": "portal:/home/krassowski/jupyterlab/packages/extensionmanager-extension",
+    "@jupyterlab/filebrowser": "portal:/home/krassowski/jupyterlab/packages/filebrowser",
+    "@jupyterlab/filebrowser-extension": "portal:/home/krassowski/jupyterlab/packages/filebrowser-extension",
+    "@jupyterlab/fileeditor": "portal:/home/krassowski/jupyterlab/packages/fileeditor",
+    "@jupyterlab/fileeditor-extension": "portal:/home/krassowski/jupyterlab/packages/fileeditor-extension",
+    "@jupyterlab/help-extension": "portal:/home/krassowski/jupyterlab/packages/help-extension",
+    "@jupyterlab/htmlviewer": "portal:/home/krassowski/jupyterlab/packages/htmlviewer",
+    "@jupyterlab/htmlviewer-extension": "portal:/home/krassowski/jupyterlab/packages/htmlviewer-extension",
+    "@jupyterlab/hub-extension": "portal:/home/krassowski/jupyterlab/packages/hub-extension",
+    "@jupyterlab/imageviewer": "portal:/home/krassowski/jupyterlab/packages/imageviewer",
+    "@jupyterlab/imageviewer-extension": "portal:/home/krassowski/jupyterlab/packages/imageviewer-extension",
+    "@jupyterlab/inspector": "portal:/home/krassowski/jupyterlab/packages/inspector",
+    "@jupyterlab/inspector-extension": "portal:/home/krassowski/jupyterlab/packages/inspector-extension",
+    "@jupyterlab/javascript-extension": "portal:/home/krassowski/jupyterlab/packages/javascript-extension",
+    "@jupyterlab/json-extension": "portal:/home/krassowski/jupyterlab/packages/json-extension",
+    "@jupyterlab/launcher": "portal:/home/krassowski/jupyterlab/packages/launcher",
+    "@jupyterlab/launcher-extension": "portal:/home/krassowski/jupyterlab/packages/launcher-extension",
+    "@jupyterlab/logconsole": "portal:/home/krassowski/jupyterlab/packages/logconsole",
+    "@jupyterlab/logconsole-extension": "portal:/home/krassowski/jupyterlab/packages/logconsole-extension",
+    "@jupyterlab/lsp": "portal:/home/krassowski/jupyterlab/packages/lsp",
+    "@jupyterlab/lsp-extension": "portal:/home/krassowski/jupyterlab/packages/lsp-extension",
+    "@jupyterlab/mainmenu": "portal:/home/krassowski/jupyterlab/packages/mainmenu",
+    "@jupyterlab/mainmenu-extension": "portal:/home/krassowski/jupyterlab/packages/mainmenu-extension",
+    "@jupyterlab/markdownviewer": "portal:/home/krassowski/jupyterlab/packages/markdownviewer",
+    "@jupyterlab/markdownviewer-extension": "portal:/home/krassowski/jupyterlab/packages/markdownviewer-extension",
+    "@jupyterlab/markedparser-extension": "portal:/home/krassowski/jupyterlab/packages/markedparser-extension",
+    "@jupyterlab/mathjax-extension": "portal:/home/krassowski/jupyterlab/packages/mathjax-extension",
+    "@jupyterlab/mermaid": "portal:/home/krassowski/jupyterlab/packages/mermaid",
+    "@jupyterlab/mermaid-extension": "portal:/home/krassowski/jupyterlab/packages/mermaid-extension",
+    "@jupyterlab/metadataform": "portal:/home/krassowski/jupyterlab/packages/metadataform",
+    "@jupyterlab/metadataform-extension": "portal:/home/krassowski/jupyterlab/packages/metadataform-extension",
+    "@jupyterlab/metapackage": "portal:/home/krassowski/jupyterlab/packages/metapackage",
+    "@jupyterlab/nbconvert-css": "portal:/home/krassowski/jupyterlab/packages/nbconvert-css",
+    "@jupyterlab/nbformat": "portal:/home/krassowski/jupyterlab/packages/nbformat",
+    "@jupyterlab/notebook": "portal:/home/krassowski/jupyterlab/packages/notebook",
+    "@jupyterlab/notebook-extension": "portal:/home/krassowski/jupyterlab/packages/notebook-extension",
+    "@jupyterlab/observables": "portal:/home/krassowski/jupyterlab/packages/observables",
+    "@jupyterlab/outputarea": "portal:/home/krassowski/jupyterlab/packages/outputarea",
+    "@jupyterlab/pdf-extension": "portal:/home/krassowski/jupyterlab/packages/pdf-extension",
+    "@jupyterlab/pluginmanager": "portal:/home/krassowski/jupyterlab/packages/pluginmanager",
+    "@jupyterlab/pluginmanager-extension": "portal:/home/krassowski/jupyterlab/packages/pluginmanager-extension",
+    "@jupyterlab/property-inspector": "portal:/home/krassowski/jupyterlab/packages/property-inspector",
+    "@jupyterlab/rendermime": "portal:/home/krassowski/jupyterlab/packages/rendermime",
+    "@jupyterlab/rendermime-extension": "portal:/home/krassowski/jupyterlab/packages/rendermime-extension",
+    "@jupyterlab/rendermime-interfaces": "portal:/home/krassowski/jupyterlab/packages/rendermime-interfaces",
+    "@jupyterlab/running": "portal:/home/krassowski/jupyterlab/packages/running",
+    "@jupyterlab/running-extension": "portal:/home/krassowski/jupyterlab/packages/running-extension",
+    "@jupyterlab/settingeditor": "portal:/home/krassowski/jupyterlab/packages/settingeditor",
+    "@jupyterlab/settingeditor-extension": "portal:/home/krassowski/jupyterlab/packages/settingeditor-extension",
+    "@jupyterlab/settingregistry": "portal:/home/krassowski/jupyterlab/packages/settingregistry",
+    "@jupyterlab/shortcuts-extension": "portal:/home/krassowski/jupyterlab/packages/shortcuts-extension",
+    "@jupyterlab/statedb": "portal:/home/krassowski/jupyterlab/packages/statedb",
+    "@jupyterlab/statusbar": "portal:/home/krassowski/jupyterlab/packages/statusbar",
+    "@jupyterlab/statusbar-extension": "portal:/home/krassowski/jupyterlab/packages/statusbar-extension",
+    "@jupyterlab/terminal": "portal:/home/krassowski/jupyterlab/packages/terminal",
+    "@jupyterlab/terminal-extension": "portal:/home/krassowski/jupyterlab/packages/terminal-extension",
+    "@jupyterlab/testing": "portal:/home/krassowski/jupyterlab/packages/testing",
+    "@jupyterlab/theme-dark-extension": "portal:/home/krassowski/jupyterlab/packages/theme-dark-extension",
+    "@jupyterlab/theme-dark-high-contrast-extension": "portal:/home/krassowski/jupyterlab/packages/theme-dark-high-contrast-extension",
+    "@jupyterlab/theme-light-extension": "portal:/home/krassowski/jupyterlab/packages/theme-light-extension",
+    "@jupyterlab/toc": "portal:/home/krassowski/jupyterlab/packages/toc",
+    "@jupyterlab/toc-extension": "portal:/home/krassowski/jupyterlab/packages/toc-extension",
+    "@jupyterlab/tooltip": "portal:/home/krassowski/jupyterlab/packages/tooltip",
+    "@jupyterlab/tooltip-extension": "portal:/home/krassowski/jupyterlab/packages/tooltip-extension",
+    "@jupyterlab/translation": "portal:/home/krassowski/jupyterlab/packages/translation",
+    "@jupyterlab/translation-extension": "portal:/home/krassowski/jupyterlab/packages/translation-extension",
+    "@jupyterlab/ui-components": "portal:/home/krassowski/jupyterlab/packages/ui-components",
+    "@jupyterlab/ui-components-extension": "portal:/home/krassowski/jupyterlab/packages/ui-components-extension",
+    "@jupyterlab/vega5-extension": "portal:/home/krassowski/jupyterlab/packages/vega5-extension",
+    "@jupyterlab/workspaces": "portal:/home/krassowski/jupyterlab/packages/workspaces",
+    "@jupyterlab/workspaces-extension": "portal:/home/krassowski/jupyterlab/packages/workspaces-extension",
+    "@jupyterlab/testutils": "portal:/home/krassowski/jupyterlab/testutils",
+    "@jupyter/ydoc": "3.0.0",
+    "@types/react": "18.0.26",
+    "react": "18.2.0",
+    "yjs": "13.5.49",
+    "y-protocols": "1.0.5"
   }
 }
diff --git a/tsconfig.json b/tsconfig.json
index 40989c6..b2695bc 100644
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,6 +1,7 @@
 {
   "compilerOptions": {
     "allowSyntheticDefaultImports": true,
+    "skipLibCheck": true,
     "composite": true,
     "declaration": true,
     "esModuleInterop": true,
```

</details>

krassowski:
Maybe you did not re-complie on the jupyterlab lab side?

davidbrochart:
> Maybe you did not re-complie on the jupyterlab lab side?

Yes I did, but your instructions [here](https://github.com/jupyterlab/jupyter-collaboration/pull/418#issuecomment-2548525429) solved the issue for me. This is definitely not my zone of comfort :thinking:

davidbrochart:
Works fine on my side, and the `RTC:` prefix is gone :tada:

packages/docprovider/src/ydrive.ts
davidbrochart: I'm not sure, it looks like you added this change in #337?

krassowski: Yup, I went ahead and removed it, I believe it is not needed as the default drive will emit these.

davidbrochart (approved):

davidbrochart:
One thing I'm wondering: I was expecting to be able to right click e.g. on a notebook and be able to open it with the RTC content provider or with the REST content provider, but it seems I have no choice but using RTC.

krassowski:
Yes, this is a choice we have, either:
- register a new widget factory for notebook and allow users to choose (via "Open with" or in settings), or
- overwrite the content provider on the default factory

I chose the latter for simplicity and to avoid user confusion. We can change it.

davidbrochart:
> I chose the latter for simplicity and to avoid user confusion.

I think we can start with that, as it was previously like that.
Actually there was a possibility to open a file in non-RTC, if it had been opened before jupyter-collaboration was installed, because it was in the workspace. But this led to a lot of confusion, and this PR makes it clearer because it will always be opened in RTC so it's even better :+1:

packages/collaborative-drive/src/tokens.ts
```diff
@@ -21,18 +22,7 @@ export const IGlobalAwareness = new Token<IAwareness>(
...
- */
-export type SharedDocumentFactory = (
-  options: Contents.ISharedFactoryOptions
-) => YDocument<DocumentChange>;
-
-/**
- * A Collaborative implementation for an `IDrive`, talking to the
- * server using the Jupyter REST API and a WebSocket connection.
- */
-export interface ICollaborativeDrive extends Contents.IDrive {
```
jtpio, line 35: Looks like we'll want to log that somewhere in the changelog as API breaking change, since other extensions may currently depend on `ICollaborativeDrive`?

For instance it's the case of JupyterCAD and JupyterGIS:
- https://github.com/jupytercad/JupyterCAD/blob/c246bf28480b8df1cbe2d20a0bbcec40aa881a5e/packages/base/src/panelview/rightpanel.tsx#L15
- https://github.com/geojupyter/jupytergis/blob/3bfd970aa7a513ebad30d31acf24adef11b6cf32/python/jupytergis_core/src/factory.ts#L2

packages/docprovider-extension/src/filebrowser.ts
```diff
@@ -4,72 +4,88 @@
...
+      const trans = translator.load('jupyter_collaboration');
+      const defaultDrive = (app.serviceManager.contents as ContentsManager)
+        .defaultDrive;
+      if (!defaultDrive) {
+        throw Error(
+          'Cannot initialize content provider: default drive property not accessible on contents manager instance.'
+        );
+      }
+      const registry = defaultDrive.contentProviderRegistry;
+      if (!registry) {
```
jtpio, line 73: Will that make `jupyter-collaboration` only compatible with JupyterLab 4.4 or higher?

packages/docprovider-extension/src/filebrowser.ts
```diff
@@ -131,10 +155,11 @@ export const ynotebook: JupyterFrontEndPlugin<void> = {
         disableDocumentWideUndoRedo
       });
     };
-    drive.sharedModelFactory.registerDocumentFactory(
+    contentProvider.sharedModelFactory.registerDocumentFactory(
       'notebook',
       yNotebookFactory
     );
+    notebookFactory.contentProviderId = 'rtc';
```
jtpio, line 162: This seems to indicate that multiple extensions could try to set their content provider on the default notebook and editor factories? 

What is the expected behavior if two extensions try to do this?

packages/docprovider-extension/src/filebrowser.ts
krassowski: Yes, I would propose that, the next version of `jupyter-collaboration` will require JupyterLab 4.4.

packages/docprovider-extension/src/filebrowser.ts
krassowski: > What is the expected behavior if two extensions try to do this?

Currently the last one wins. Alternative, we could log a warning or reject a second use of setter.

jtpio: > Currently the last one wins. Alternative, we could log a warning or reject a second use of setter.

Maybe a hard failure would be preferable for end users, so it's easier to debug and report?

krassowski: Make sense. Done in https://github.com/jupyterlab/jupyterlab/pull/17092/commits/1173ca7ec995df071142bcb1170781a0897be246.

packages/collaborative-drive/src/tokens.ts
krassowski: It looks like the transition will be smooth because:
- both extensions used `ICollaborativeDrive` as an optional dependency (after all it was just added)
- neither of the extensions actually relied on the `Drive` part of `ICollaborativeDrive` - the same properties that they needed will be available on `ICollaborativeContentProvider`

So when cutting the release we will just need to add a note that migration from `ICollaborativeDrive` to `ICollaborativeContentProvider` token for access to shared factory and forks will be required. I do not see a good way to do that right now as the changelog for next version is not there yet - is it ok if we add it during/just after cutting the release?

jtpio:
Just tried the latest changes of this PR and https://github.com/jupyterlab/jupyterlab/pull/17092 locally with a dev install and it seems to be working fine. Also when opening files with other packages like `voila` and `notebook` (just opening the link, `notebook` would still need to update to the latest packages:

image

tsconfig.json
```diff
@@ -13,7 +13,8 @@
     "noUnusedLocals": true,
     "preserveWatchOutput": true,
     "resolveJsonModule": true,
-		"strict": true,
+	"strict": true,
+    "skipLibCheck": true,
```
krassowski, line 17: We should be able to drop `skipLibCheck` once https://github.com/jupyterlab/jupyterlab/pull/17141 is in and a new alpha is released; we could probably just merge and release for testing already though.

jtpio: Curious if we could double check the TypeScript build also fails for this PR if we remove `skipLibCheck`? 

Otherwise we could indeed keep it for now and remove it when doing the next update.

krassowski: It was failing for me locally. Let me fix it by bumping the TypeScript version here instead.

jtpio: Yes, linking to https://github.com/jupyterlab/jupyterlab/pull/17141#issuecomment-2574846099 for more info.

krassowski:
It looks like there is no version pins on JupyterLab in the distribution tarballs. I opened https://github.com/jupyterlab/jupyter-collaboration/issues/425 to discuss this.

tsconfig.json
krassowski: Done in e2c87ad8ab92f88f61fb23125077b16e19c2ed94.

packages/docprovider/src/ydrive.ts
```diff
@@ -245,9 +237,11 @@ export class YDrive extends Drive implements ICollaborativeDrive {
 
   private _user: User.IManager;
   private _trans: TranslationBundle;
-  private _providers: Map<string, WebSocketProvider>;
   private _globalAwareness: Awareness | null;
+  private _providers: Map<string, WebSocketProvider>;
   private _ydriveFileChanged = new Signal<this, Contents.IChangedArgs>(this);
+  private _serverSettings: ServerConnection.ISettings;
+  private _prefix = 'RTC'; // TODO
```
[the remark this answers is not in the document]
krassowski: I ended up removing it in 4e69d61095db82fed31e2db30b217a377cd50361 on the basis that the JS assets would usually be served by the same server which hols the awareness information so most likely everyone would be using the new format. It was also needed to make opening documents work without needlessly complicating the codebase.

krassowski:
All green :green_heart:! Good to merge and release an alpha version for testing?

jtpio:
I just tried locally in a new environment with the latest JupyterLab 4.4.0a2 and the wheels from https://github.com/jupyterlab/jupyter-collaboration/actions/runs/12651599033?pr=418, and it seems to be working well :+1:

jtpio:
> I think this would be a major version to signify that it is not backwards compatible.

Right, a new major release sounds like a safer approach :+1:

PR #422, opened 2025-01-01
Fix duplicate collaborator entries in collaboration panel

nzinfo:
## Problem
When the same user connects from multiple browsers, the collaborator panel may show duplicate entries even when they are viewing the same file. This causes unnecessary clutter and React key warnings.

## Solution
- Add deduplication logic in `_onAwarenessChanged` using a Set to track unique combinations of username and current file
- Only add a collaborator to the list if their unique key (username + current file) hasn't been seen before
- Use consistent unique keys in both the awareness change handler and React component rendering

## Changes
- Add `collaborators_keys` Set to track unique collaborator entries
- Create unique keys using username and current file path
- Filter out duplicate entries before adding to collaborators list
- Update React key prop to use the same unique key format

## Testing
1. Open the same notebook in multiple browser windows with the same user
2. Verify only one entry shows per unique (username, current file) combination
3. Check browser console - no more React key warnings

### DIFF
--- a/packages/collaboration/src/collaboratorspanel.tsx
+++ b/packages/collaboration/src/collaboratorspanel.tsx
@@ -98,14 +98,19 @@ export class CollaboratorsPanel extends Panel {
   private _onAwarenessChanged = () => {
     const state = this._awareness.getStates() as any;
     const collaborators: ICollaboratorAwareness[] = [];
-
+    const collaborators_keys: Set<string> = new Set();
     state.forEach((value: Partial<ICollaboratorAwareness>, key: any) => {
       if (
         this._currentUser.isReady &&
         value.user &&
         value.user.username !== this._currentUser.identity!.username
       ) {
-        collaborators.push(value as ICollaboratorAwareness);
+        const uniqueKey = `${value.user.username}-${value.current || 'no-current'}`;
+
+        if (!collaborators_keys.has(uniqueKey)) { 
+          collaborators.push(value as ICollaboratorAwareness);
+          collaborators_keys.add(uniqueKey);
+        }
       }
     });
     this._collaboratorsChanged.emit(collaborators);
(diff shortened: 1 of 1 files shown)


nzinfo:
## The warning

Warning: Each child in a list should have a unique "key" prop.

Check the render method of `CollaboratorsBody`. See https://reactjs.org/link/warning-keys for more information.
Collaborator@http://localhost:8888/lab/extensions/@jupyter/collaboration-extension/static/collaboration_lib_index_js.0471f9a278b7578b8646.js:105:76
CollaboratorsBody@http://localhost:8888/lab/extensions/@jupyter/collaboration-extension/static/collaboration_lib_index_js.0471f9a278b7578b8646.js:96:94 react.development.js:209
    React 5
        printWarning
        error
        validateExplicitKey
        validateChildKeys
        createElementWithValidation
    CollaboratorsBody collaboratorspanel.js:75

davidbrochart (commented):
Thanks for the PR @nzinfo.
How about making `collaborators` a `Map`, instead of having the keys in `collaborators_keys` and the values in `collaborators`?

nzinfo:
@davidbrochart Thank you for the review.  I've changed collaborators into a map.

packages/collaboration/src/collaboratorspanel.tsx
```diff
@@ -132,9 +136,11 @@ export function CollaboratorsBody(props: {
 
   return (
     <div className={COLLABORATORS_LIST_CLASS}>
-      {collaborators.map((collaborator, i) => {
+      {collaborators.map((collaborator) => {
+        const uniqueKey = `${collaborator.user.username}-${collaborator.current || 'no-current'}`;
         return (
           <Collaborator
+            key={uniqueKey}
```
davidbrochart, line 143: Is it used?

krassowski: I believe it is needed for React to know which entry is which.

nzinfo: Yes, and It's the key that can avoid the warning "Warning: Each child in a list should have a unique "key" prop."

davidbrochart (approved):
Thanks @nzinfo.

Issue #437, opened 2025-02-05
Make the share button actually share something


Hyrla:
_Disclaimer: I've been using Jupyter for only 3 weeks. Some of my comments and suggestions may be irrelevant._

### Problem

If I understand it properly, as of today you can't collaborate with someone without:  
a) Having an impersonation token from him/her  
b) Having the Hub admin explicitly giving you the right to access all servers or at least this specific user's server  

The share button that jupyter-collaboration provides only works in one of these two ways. The problem is (at least for me) that it is not practical as a Hub admin to manage "collaboration accounts" (as suggested in [Jupyter Hub documentation](https://jupyterhub.readthedocs.io/en/5.2.1/tutorial/collaboration-users.html)) nor to give all my users the ability to access everybody's server.  

### Proposed Solution

Jupyter Hub 5.0 provides a "shared server" feature to let a user dynamically share his/her server with anyone else without involving admin trickery. As of today, it is only achievable through the API. I think we should change the share button behavior (which was relevant before Jupyter Hub 5.0) to a new UI that would let a non-admin user invite other users to his/her own server and revoke access later if needed using the shared server feature [documented here](https://jupyterhub.readthedocs.io/en/5.2.1/tutorial/collaboration-users.html). I tested the API myself and I can confirm that I managed to give another non-admin user the ability to access my server without any other permission needed, using my token to access the Hub API.  

Here is the very basic Python code I used to test this feature:
```
import requests
api_url = "https://jupyter.esiea.fr/hub/api"
token = "Owner token"
owner_username = "owner@esiea.fr" 
shared_user = "other-user@et.esiea.fr"

headers = {"Authorization": f"token {token}", "Content-Type": "application/json"}
data = {"user": shared_user}
response = requests.post(f"{api_url}/shares/{owner_username}/", headers=headers, json=data)

print(response.status_code, response.json())
```

Here is the API reply:
`200 {'server': {'user': {'name': 'owner@esiea.fr'}, 'name': '', 'url': '/user/owner@esiea.fr/', 'full_url': None, 'ready': False}, 'scopes': ['access:servers!server=owner@esiea.fr/'], 'user': {'name': 'other-user@et.esiea.fr'}, 'group': None, 'kind': 'user', 'created_at': '2025-02-04T15:34:51.611747Z'}`

After this API request, "other-user@esiea.fr" was able to access "owner@esiea.fr" without any impersonation token, just by typing the server URL.

### Additional context

[...]

PR #438, opened 2025-02-08
Add a UI to let users share their servers to other users and groups

Hyrla:
Hello,  
I've added the necessary code related to my issue [#437](https://github.com/jupyterlab/jupyter-collaboration/issues/437).  

image  

It replaces the default link share UI in cases where Jupyter is running behind a Hub that allows the creation of shares. It enables users to:

- Create a share to the current server for a user or a group using a search box.
- Revoke a share for a user or a group.

The permissions needed (both in Hub and Lab) are:

- `"shares!user"` to manage users' shares.
- `"read:users:name"` to create shares by designating a user by their name _(TODO: check if it's really necessary if `list:users` is already in scope?)_.
- `"list:users"` (not mandatory) to list users for the search box.
- `"list:groups"` (not mandatory) to list groups for the search box.
- `"servers!user"` (not mandatory) to allow other users to start and stop the shared server

My code **does not**:

- Work if neither `list:users` nor `list:groups` is in scope. I wanted to add an option to invite someone by typing their exact username, without having to search.
- Manage [share codes](https://jupyterhub.readthedocs.io/en/5.2.1/reference/rest-api.html#operation/get-share-codes-server), which may be useful in some scenarios.

This development has been made for my own internal purposes, but I think it could be useful to everyone. Comments and critiques are welcome!


ykazakov:
Amazing work!
I think determining the server owner should be more robust:

https://github.com/Hyrla/jupyter-collaboration/blob/1743de5c50dcd6372dd7926cfc72b96970544f02/packages/collaboration/src/sharedlink.ts#L48

This does not work if `c.JupyterHub.base_url` contains `/`.

Hyrla:
> Amazing work! I think determining the server owner should be more robust:
> 
> https://github.com/Hyrla/jupyter-collaboration/blob/1743de5c50dcd6372dd7926cfc72b96970544f02/packages/collaboration/src/sharedlink.ts#L48
> 
> This does not work if `c.JupyterHub.base_url` is set.

Thank you for your feedback. Do you have any suggestion to determine the server owner with a cleaner way?

ykazakov:
I think, if the server is running under JupyterHub, one can extract `JupyterHub.base_url` from `PageConfig.getOption('hubPrefix')` by removing the trailing `/hub/`.

ykazakov:
There appears to be an option `M.PageConfig.getOption("hubServerUser")`, which is probably what you need?

ykazakov:
Similarly, `serverName` can be extracted from `PageConfig.getOption('hubServerName')`.

Hyrla:
> You can see all available options of `PageConfig` by inspecting the `jupyter-config-data` tag of the HTML source of the lab page. See https://jupyterlab.readthedocs.io/en/latest/api/functions/coreutils.PageConfig.getOption.html

Omg thank you, I searched hours for this list of PageConfig

Hyrla:
@ykazakov I've pushed the fix, thanks again for your feedback. Could you please try if it works on your Hub? It works on mine but c.JupyterHub.base_url is not set there.

ykazakov:
@Hyrla I tested, now it works with `c.JupyterHub.base_url` set!
However, these options are probably available only if running under JupyterHub, so the variables should probably be set after the check is done:

https://github.com/jupyterlab/jupyter-collaboration/blob/0f9e77ed4d839a59757295f3e499a1810dc8a3d8/packages/collaboration/src/sharedlink.ts#L56-L61

Also, if running behind a JupyterHub, shouldn't *all* these options be set, not just one of them? 
I would suggest to instead check the presence of (all) options that you really use for the dialog.

Hyrla:
@ykazakov thank you for your feedback. I am pushing a fix concerning the two getOptions() that are misplaced. Concerning the detection of JupyterHub, not all these variables are necessarily set. I did set back the original condition as seen before my pull request: https://github.com/jupyterlab/jupyter-collaboration/blob/1c23b688ba22c1911d3ee93542e1da6e67f0514a/packages/collaboration/src/sharedlink.ts#L53

ykazakov:
I have been further testing, and in general, the dialog works great! 👍

Some (mostly minor) comments:
1. In the search list one cannot tell users and groups apart. If there is a group named exactly as the user, it is not clear which one is wich.
2. Clicking on the item in the list immediately creates a share. Perhaps it would make sense to have a confirmation dialogue to prevent unintentional shares (particularly for groups).
3. What happens if the list is very long (hundreds of users)? Perhaps wait until something is typed in the search box and the number of matches is narrowed down. See the [nextcloud share dialogue](https://nextcloud.com/blog/sharing-in-nextcloud/).
4. If the user does not have a permission to share a server (example: user Alice shares her server with user Bob, and user Bob tries to re-share the server of Alice), the share button results in the standard dialog, which would not work even if to include the token. Perhaps instead, one should show a dialog that the user does not have permission to share this server (if in doubt, ask admin), or even better, not to show the share icon at all.

Hyrla:
> I have been further testing, and in general, the dialog works great! 👍

Thank you!

> Some (mostly minor) comments:
> 
>     1. In the search list one cannot tell users and groups apart. If there is a group named exactly as the user, it is not clear which one is wich.

Yes you're right, I will add a "Group" prefix like in the shares list

>     2. Clicking on the item in the list immediately creates a share. Perhaps it would make sense to have a confirmation dialogue to prevent unintentional shares (particularly for groups).

Do you have any example snippets of code to create such a dialogue ? I wasn't able to find one (I am not an expert at all in TS nor Jupyter extension)

>     3. What happens if the list is very long (hundreds of users)? Perhaps wait until something is typed in the search box and the number of matches is narrowed down. See the [nextcloud share dialogue](https://nextcloud.com/blog/sharing-in-nextcloud/).

On my Hub there are about 350 users and it works like a charm. I am using the API with pagination, so there is no big query with every users at a single time. But if there is thousands of users it could be a problem. As far as I know, there is no Hub API to search for users or group?

>     4. If the user does not have a permission to share a server (example: user Alice shares her server with user Bob, and user Bob tries to re-share the server of Alice), the share button results in the standard dialog, which would not work even if to include the token. Perhaps instead, one should show a dialog that the user does not have permission to share this server (if in doubt, ask admin), or even better, not to show the share icon at all.

Like for 2., I don't know (yet) a way to create such dialogue but I will add it asap

Hyrla:
> This should be where the page config vars originate: https://github.com/jupyterhub/jupyterhub/blob/5.2.1/jupyterhub/singleuser/extension.py#L216-L234

Thank you!

packages/collaboration/src/sharedlink.ts
```diff
@@ -45,22 +45,458 @@ export async function showSharedLinkDialog({
...
+  }
+
+  // If we can create a share, open the proper UI
+  if (canCreateShare) {
+    let readableServerName = serverName;
+    if (readableServerName === '') {
+      readableServerName = 'default';
+    }
+    return showDialog({
+      title: trans.__('Share Jupyter Server ') + readableServerName,
```
krassowski, line 141: Different languages have different order of words

```suggestion
      title: trans.__('Share Jupyter Server %1', readableServerName),
```

packages/collaboration/src/sharedlink.ts
```diff
@@ -45,22 +45,458 @@ export async function showSharedLinkDialog({
...
+        }
+      );
+      const data = await usersResponse.json();
+      usersData = usersData.concat(data);
+      hasMore = data.length === limit;
+      offset += limit;
+      }
+
+      const sharedUserNames = new Set(
+      this._shares
```
krassowski, line 252: It looks like indentation is missing, I guess the linter did not run.

packages/collaboration/src/sharedlink.ts
```diff
@@ -45,22 +45,458 @@ export async function showSharedLinkDialog({
...
+  protected onBeforeDetach(msg: Message): void {
+    super.onBeforeDetach(msg);
+  }
+
+  private async loadUsers(): Promise<void> {
+    // If possible, download the users list for the UI
+    if (this._canListUsers) {
+      let offset = 0;
+      const limit = 200;
+      let usersData: any[] = [];
```
krassowski, line 233: We should avoid adding `any` in new code. If there is no interface for user already, let's define a simple one with `name` and whatever else is needed.

packages/collaboration/src/sharedlink.ts
```diff
@@ -45,22 +45,458 @@ export async function showSharedLinkDialog({
...
+    );
+    const sharesData = await sharesResponse.json();
+    this._shares = sharesData.items.map((item: any) => ({
+      name: item.user?.name || item.group?.name,
+      createdAt: item.created_at,
+      type: item.user ? 'user' : 'group'
+    }));
+  }
+
+  private async createShare(sharewith: any, type: any): Promise<void> {
```
krassowski, line 307: It would be great if we could replace `any` with proper typing here.

packages/collaboration/src/sharedlink.ts
```diff
@@ -45,22 +45,458 @@ export async function showSharedLinkDialog({
...
+  let serverName = '';
+  let serverOwner = '';
+
+  // If hubUser is set, we are behind a JupyterHub
+  if (
+    PageConfig.getOption('hubUser') !== ''
+  ) {
+    // Get server name and owner
+    serverOwner = PageConfig.getOption('hubServerUser');
+    serverName = PageConfig.getOption('hubServerName');
```
krassowski, line 62: Is the `hubServerName` guaranteed to not contain `/` or `#` or `?` which would break the URLs created below?

packages/collaboration/src/sharedlink.ts
```diff
@@ -45,22 +45,458 @@ export async function showSharedLinkDialog({
...
+    );
+    this._searchInput.addEventListener('input', () => {
+      this.updateSearchResults();
+    });
+    searchContainer.appendChild(this._searchInput);
+    dialogBody.appendChild(searchContainer);
+
+    // Add search results container
+    this._searchResults = document.createElement('div');
+    this._searchResults.classList.add('search-results');
```
krassowski, line 367: Let's prefix the class names to avoid conflicts with other extensions and user code:

```suggestion
    this._searchResults.classList.add('jp-ManageSharesBody-search-results');
```

packages/collaboration/src/sharedlink.ts
```diff
@@ -45,22 +45,458 @@ export async function showSharedLinkDialog({
...
+    dialogBody.appendChild(this._searchResults);
+
+    // Add selected users container
+    this._sharesContainer = document.createElement('div');
+    this._sharesContainer.classList.add('selected-users');
```
krassowski, line 372:
```suggestion
    this._sharesContainer.classList.add('jp-ManageSharesBody-selected-users');
```

packages/collaboration/src/sharedlink.ts
```diff
@@ -45,22 +45,458 @@ export async function showSharedLinkDialog({
...
+    dialogBody.appendChild(this._sharesContainer);
+
+    dialogBody.insertAdjacentHTML(
+      'beforeend',
+      `<input readonly value="${this._url}" class="url-input">`
```
krassowski, line 377:
```suggestion
      `<input readonly value="${this._url}" class="jp-ManageSharesBody-url-input">`
```

packages/collaboration/src/sharedlink.ts
```diff
@@ -45,22 +45,458 @@ export async function showSharedLinkDialog({
...
+    const query = this._searchInput?.value || '';
+
+    const filteredUsers = this._users.filter(
+      user =>
+        user.name.toLowerCase().includes(query.toLowerCase()) || query === ''
+    );
+
+    filteredUsers.forEach(user => {
+      const userElement = document.createElement('div');
+      userElement.classList.add('user-item');
```
krassowski, line 409:
```suggestion
      userElement.classList.add('jp-ManageSharesBody-user-item');
```

packages/collaboration/src/sharedlink.ts
```diff
@@ -45,22 +45,458 @@ export async function showSharedLinkDialog({
...
+  }
+
+  private async updateSharesList(): Promise<void> {
+    if (!this._sharesContainer) {
+      return;
+    }
+    this._sharesContainer.innerHTML = '';
+
+    const table = document.createElement('table');
+    table.classList.add('shares-table');
```
krassowski, line 434:
```suggestion
    table.classList.add('jp-ManageSharesBody-shares-table');
```

packages/collaboration/src/sharedlink.ts
```diff
@@ -45,22 +45,458 @@ export async function showSharedLinkDialog({
...
+          year: 'numeric',
+          month: '2-digit',
+          day: '2-digit'
+        });
+
+        row.innerHTML = `
+        <td>${
+          share.type === 'group'
+            ? this._trans.__('Group ') + share.name
+            : share.name
```
krassowski, line 468:
```suggestion
            ? this._trans.__('Group %1', share.name)
            : share.name
```

krassowski, line 468: I worry that this might lead to script injection is `share.name` is comporomised. As a rule we should avoid `innerHTML` as much as possible. Can you instead create this cells with `document.createElement`?

packages/collaboration/style/sharedlink.css
krassowski, line 1: The class names in this file will need prefixing as per the comments above.

packages/collaboration/src/sharedlink.ts
```diff
@@ -45,22 +45,458 @@ export async function showSharedLinkDialog({
...
+    );
+  }
+
+  private populateBody(dialogBody: HTMLElement): void {
+    // Add search input
+    const searchContainer = document.createElement('div');
+    searchContainer.classList.add('search-container');
+    this._searchInput = document.createElement('input');
+    this._searchInput.type = 'text';
+    this._searchInput.classList.add('search-input');
```
krassowski, line 355:
```suggestion
    searchContainer.classList.add('jp-ManageSharesBody-search-container');
    this._searchInput = document.createElement('input');
    this._searchInput.type = 'text';
    this._searchInput.classList.add('jp-ManageSharesBody-search-input');
```

packages/collaboration/src/sharedlink.ts
```diff
@@ -45,22 +45,458 @@ export async function showSharedLinkDialog({
...
+        },
+        body: JSON.stringify({
+          [type]: sharewith.name,
+          scopes
+        })
+      }
+    );
+  }
+
+  private async deleteShare(sharewith: any, type: any): Promise<void> {
```
krassowski, line 334: Needs typing

krassowski (commented):
Coming back to this PR as we are nearing 4.1 release and I thought it might be a nice addition in there. I left a few comments on security, internationalization and style but otherwise it looks good to me. Ideally we would also have a UI test for it, we would probably want to mock the `PageConfig` values.

packages/collaboration/src/sharedlink.ts
Hyrla: I tried to create a named server "test#test/test" and the Hub created the named server "test" instead. So it looks like it's already correctly parsed on Hub's side?

krassowski:
I will push typing updates.

It looks like both `/user` and `/servers` return `kind: 'user'` and `kind: 'server'` respectively, is the addition of `type` needed?

https://jupyterhub.readthedocs.io/en/stable/reference/rest-api.html#operation/get-users
https://jupyterhub.readthedocs.io/en/stable/reference/rest-api.html#operation/get-groups

Hyrla:
Thank you very much @krassowski for your feedbacks and help on this PR!

krassowski:
4.1.0 RC0 is available for testing:
- https://github.com/jupyterlab/jupyter-collaboration/releases/tag/v4.1.0rc0
- https://pypi.org/project/jupyter-collaboration/4.1.0rc0/

Hyrla:
I have installed 4.1.0 RC0 on my school Jupyter Hub and everything seems to work fine

Hyrla:
@krassowski how do you think we should handle the documentation of this feature? Especially regarding the permissions needed to make it working. On my Hub, I've simply added this to my Z2JH yaml config:

```
hub:
[...]
  loadRoles:
    share:
      description: Allow users to manage shares on their server
      scopes: [shares!user, read:users:name, list:users, servers!user]
      groups: [d35c2143-xxxx, 6bdd8535-xxxx, c481a9cf-xxxx, c1927c2a-xxxx, ba715d4d-xxxx, 6c7c5e80-xxxx, 9e98f201-xxxx, 4e3c0034-xxxx]
```

Then I've added this extra config to give these permissions to the lab's scope:

```
hub:
[...]
 extraConfig:
  customLabScopesOptions: |
      c.Spawner.oauth_client_allowed_scopes = ["read:users:name", "shares!user", "list:users", "servers!user"]
```

Finally, I've added this extra config to remove the oauth confirmation page (which I found not necessary for my use case):
```
hub:
[...]
 extraConfig:
 [...]
    disableConfirmation: |
      # Désactivation de la page de confirmation relou
      import jupyterhub.apihandlers.auth
      class DisableConfirmationOAuthAuthorizeHandler(
          jupyterhub.apihandlers.auth.OAuthAuthorizeHandler
      ):
          def needs_oauth_confirm(self, *args, **kwargs):
              return False
      for i, (route, handler) in enumerate(jupyterhub.apihandlers.default_handlers):
          if route == "/api/oauth2/authorize":
              jupyterhub.apihandlers.default_handlers[i] = (route, DisableConfirmationOAuthAuthorizeHandler)
              break
```

If needed, here is my full config: https://gitlab.esiea.fr/gauthier.heiss/esiea-jupyter-image/-/blob/main/JupyterHub.yaml

**I am pretty sure this config is not the perfect way to go and could be much simplier and secure.**

krassowski:
I would defer to @manics on that.

Hyrla:
> You could update the docs on https://jupyterhub.readthedocs.io/en/stable/tutorial/collaboration-users.html https://github.com/jupyterhub/jupyterhub/blob/main/docs/source/tutorial/collaboration-users.md
> 
> Disabling oauth isn't recommended unless all users fully trust each other, so I wouldn't include that.

Thank you for your reply. Concerning oauth, is there other any to handle it? Of course disabling to confirmation UI is not recommended at all, but otherwise every use has to accept the particular sharing role permissions when starting a server, even if it's not meant to be collaborative

PR #429, opened 2025-01-16
Pin jupyterlab at `>=4.4.0a2,<5.0.0` in `projects/jupyter-collaboration` package

andrii-i:
Adds `jupyterlab` pin at `>=4.4.0a2,<5.0.0` in `projects/jupyter-collaboration` subpackage. This would introduce a way to specify which versions of jupyterlab current version of jupyter-collaboration is compatible with which would prevent RTC-jupyterlab version mismatch issues for new versions of jupyter-collaboration.

- Fixes #425

Testing:
1. Create new conda environment
2. Install jupyterlab<4.4.0a2 by running `pip install jupyterlab==4.2.0`
3. Clone this PR and install it locally by running `pip install -e .` from `projects/jupyter-collaboration`
4. Run `jupyter --version `, note updated `jupyterlab` version

PR #446, opened 2025-02-19
Backport PR #429 on branch 3.x (Pin jupyterlab in projects/jupyter-collaboration package)

andrii-i:
* Backports #429 adjusting the version range to the correct one


krassowski:
Note: this also requires a change to release version bump script, otherwise this pin will be removed on release.

krassowski:
@andrii-i is this PR still needed? I see there is a conflict, and a question from Nick above.

Issue #432, opened 2025-01-24
Backport jupyterlab pin in projects/jupyter-collaboration to recent major versions of jupyter-collaboration (v2 and v3)


andrii-i:
### Problem

#429 pinned jupyterlab at >=4.4.0a2,<5.0.0 in projects/jupyter-collaboration package introduce a way to specify which versions of jupyterlab current version of jupyter-collaboration is compatible with which would prevent RTC-jupyterlab version mismatch issues for new versions of jupyter-collaboration.

Jupyter-collaboration v2 and v3 also have similar dependency on JupyterLab. Last new 2.x minor [version 2.1](https://github.com/jupyterlab/jupyter-collaboration/releases/tag/v2.1.0) version was released in May 2024 and last new 3.x minor [version 3.1](https://github.com/jupyterlab/jupyter-collaboration/releases/tag/v3.1.0) was released in December 2024 so within last year during JupyterLab 4 cadence. Therefore there are JupyterLab 4 users who have versions of JupyterLab compatible with jupyter-collaboration v2 or jupyter-collaboration v3 only. 

### Proposed Solution

Backport #429 to jupyter-collaboration v2 (as 2.1.6) and jupyter-collaboration v3 (as 3.1.1) adjusting the jupyterlab pin to compatible jupyterlab versions.

andrii-i:
Related:
* #436

PR #447, opened 2025-02-24
Pin jupyterlab in 2.x

andrii-i:
* Pins jupyterlab to ">=4.0.0,<4.3.0". Reasoning:
  * 4.3.0 bumps jupyter-ydoc to 3.0.0a8 and adds Content Provider API 
* Fixes "backport to 2.x" part of issue #432


krassowski:
I think the tests would need to pass if we want to merge and release this.

PR #456, opened 2025-03-31
Handle the error raised when closing a collaborative document

### LINKED ISSUE #455: Closing a collaborative document logs a error in terminal
## Description

When closing a collaborative document, an error is raised in the terminal.

```bash
[E 2025-03-31 13:39:14.631 ServerApp] Failed to write message
      File "/home/brichet/projects/jupyter_collaboration/projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py", line 278, in send
        self.write_message(message, binary=True)
      File "/home/brichet/mambaforge/envs/jupyter-collab/lib/python3.12/site-packages/tornado/websocket.py", line 332, in write_message
        raise WebSocketClosedError()
    tornado.websocket.WebSocketClosedError
```

## Reproduce

- open a notebook with collaboration enabled
- close the notebook

The error message should be logged in the terminal.

## Additional context

Related issues:
- https://github.com/jupyterlab/jupyter-chat/issues/113
- https://github.com/jupyterlab/jupyter-ai/issues/1293

brichet:
Fix #455 

When a user closes a collaborative document, it deletes the related web-socket connection to the document in the `YDocWebSocketHandler`. 
But it still tries to send a message to the web-socket, to update the user status (the fact that the user is disconnected from this document). This raises an error because the web-socket is already closed.

This PR handles this error to avoid displaying it in the terminal.

### DIFF
--- a/projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
+++ b/projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
@@ -19,7 +19,7 @@ from pycrdt import Doc, UndoManager, write_var_uint
 from pycrdt_websocket.websocket_server import YRoom
 from pycrdt_websocket.ystore import BaseYStore
 from tornado import web
-from tornado.websocket import WebSocketHandler
+from tornado.websocket import WebSocketClosedError, WebSocketHandler
 
 from .loaders import FileLoaderMapping
 from .rooms import DocumentRoom, TransientRoom
@@ -276,6 +276,10 @@ class YDocWebSocketHandler(WebSocketHandler, JupyterHandler):
         # needed to be compatible with WebsocketServer (websocket.send)
         try:
             self.write_message(message, binary=True)
+        except WebSocketClosedError:
+            # Handle error when trying to send a message while the websocket is closed.
+            # It occurs when closing a collaborative document.
+            pass
         except Exception as e:
             self.log.error("Failed to write message", exc_info=e)
 


krassowski:
> But it still tries to send a message to the web-socket, to update the user status (the fact that the user is disconnected from this document).

It sounds like such message should be sent before the websocket connection is deleted? Is this something we could change instead of adding a pretty broad exception handler?

brichet:
I'm a bit confused about the cycles of the disconnection messages.
AFAIU, when closing a document, a message is received through the websocket, and broadcasted to all the clients from the `YRoom` [here](https://github.com/jupyter-server/pycrdt-websocket/blob/a556d432d47a6364dcf3f9c25cfb6d3b63c8d595/pycrdt_websocket/yroom.py#L294), including itself.

I don't know if the disconnection messages need to be broadcasted at all.
If the others clients need to receive it, we could at least avoid sending it back to itself, since the connection is supposed to be closed.

cc. @krassowski @davidbrochart

krassowski:
Is it possibly needed for a UI element showing where the collaborators are? Or is it all independently managed via awareness room?

> If the others clients need to receive it, we could at least avoid sending it back to itself, since the connection is supposed to be closed.

Right, absolutely.

brichet:
I opened https://github.com/jupyter-server/pycrdt-websocket/pull/118, to avoid propagating the disconnection message to the disconnected client.

Closing this one since it should be handled before raising an error.

PR #457, opened 2025-04-01
Skip saving non-writable files

Darshan808:
## Description  
The `maybe_save_content` method in the `FileLoader` class attempted to save files without checking their permissions. If a file was **read-only**, this resulted in an error, which was logged in the frontend whenever changes were made to read-only files.
Screenshot from 2025-04-01 18-06-40

### Fix  
- Added a check to verify if the file is **writable** before attempting to save it; if the file is **read-only**, the save operation is skipped to prevent unnecessary errors.

### DIFF
--- a/projects/jupyter-server-ydoc/jupyter_server_ydoc/loaders.py
+++ b/projects/jupyter-server-ydoc/jupyter_server_ydoc/loaders.py
@@ -153,6 +153,9 @@ class FileLoader:
                     path, format=model["format"], type=model["type"], content=False
                 )
             )
+            # Skip saving if file is not writable"
+            if not m["writable"]:
+                return None
 
             if self.last_modified == m["last_modified"]:
                 self._log.info("Saving file: %s", path)
--- a/projects/jupyter-server-ydoc/jupyter_server_ydoc/loaders.py
+++ b/projects/jupyter-server-ydoc/jupyter_server_ydoc/loaders.py
@@ -153,7 +153,7 @@ class FileLoader:
                     path, format=model["format"], type=model["type"], content=False
                 )
             )
-            # Skip saving if file is not writable"
+            # Skip saving if file is not writable
             if not m["writable"]:
                 return None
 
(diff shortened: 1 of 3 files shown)


Darshan808:
I wonder if the failing test is related!

Darshan808:
Yes, It is related! Looking into it!

projects/jupyter-server-ydoc/jupyter_server_ydoc/loaders.py
```diff
@@ -153,6 +153,9 @@ async def maybe_save_content(self, model: dict[str, Any]) -> dict[str, Any] | No
                     path, format=model["format"], type=model["type"], content=False
                 )
             )
+            # Skip saving if file is not writable"
```
davidbrochart, line 156:
```suggestion
            # Skip saving if file is not writable
```

projects/jupyter-server-ydoc/jupyter_server_ydoc/pytest_plugin.py
```diff
@@ -271,13 +271,16 @@ def _inner(
         last_modified: datetime | None = None,
         save_delay: float | None = None,
         store: SQLiteYStore | None = None,
+        writable: bool = False,
```
davidbrochart, line 274: Shouldn't it be writable by default?
```suggestion
        writable: bool = True,
```

Darshan808: I think so. However, `FakeContentsManager` had `writable` set to `False` by default. So, just wanted to preserve it as I though it might be for some reasons.
If `"Writable": False` isn't necessary, do you want me to remove last commit and instead modify `Writable` to be `True` by default?

https://github.com/jupyterlab/jupyter-collaboration/blob/bb0e1fb8bb0c0e51f7118af640d84d197ed1b620/projects/jupyter-server-ydoc/jupyter_server_ydoc/test_utils.py#L24-L36

davidbrochart: Thanks, let's keep it like that then, and have it backwards-compatible.

davidbrochart (approved):
Thanks @Darshan808.

PR #458, opened 2025-04-04
Bump to jupyterlab v4.4.0

davidbrochart:


krassowski (approved):
Thank you!

Should we release RC and final?

PR #461, opened 2025-04-08
Fix version pins in bump script

brichet:
While https://github.com/jupyterlab/jupyter-collaboration/pull/459 fixed the version pins, the bump script [reverted it](https://github.com/jupyterlab/jupyter-collaboration/commit/b4fc51bc462677ced63f1da58384ce2bd9fe1222#diff-2271c3877caf2a061a678f38ecb6d1351c52da6c446d4612a92cd701794c8fdfL33-L36) during release.

This PR fix the bump script to pin the project versions to `>=current_version,<next_major` instead of only `>=current_version`.

### DIFF
--- a/scripts/bump_version.py
+++ b/scripts/bump_version.py
@@ -109,7 +109,8 @@ def bump(force, skip_if_dirty, spec):
     dependencies = tomlkit.array()
     for key in sorted(project_pins):
         if key != metapackage.replace("-", "_"):
-            dependencies.add_line(key + ">=" + project_pins[key])
+            next_major = f"{parse_version(project_pins[key]).major + 1}"
+            dependencies.add_line(key + ">=" + project_pins[key] + ",<" + next_major)
     metapackage_toml.get("project").add("dependencies", dependencies.multiline(True))
     metapackage_toml_path.write_text(tomlkit.dumps(metapackage_toml))
 


krassowski:
Yeah, it also removed the pin on JupyterLab >= 4.4.0, see https://github.com/jupyterlab/jupyter-collaboration/issues/462

brichet:
> Yeah, it also removed the pin on JupyterLab >= 4.4.0, see #462

Yeah, we'll need to add this when backporting this PR to `main` (currently it targets `3.x`).
I think Jupyterlab was not pinned in 3.x branch yet: https://github.com/jupyterlab/jupyter-collaboration/pull/446

krassowski:
@brichet would it make sense to also apply this to `main` branch?

Also, should we make the pin `<X.0.0dev0`?

brichet:
> @brichet would it make sense to also apply this to `main` branch?

Yes, to prevent future similar issues, when releasing collaboration `v5` I guess.

> Also, should we make the pin <X.0.0dev0?

Is this necessary ? I thought that `<X` was enough to avoid also the dev and alpha/beta releases of version `X`.

brichet:
I will probably not be able to follow up on it today, please update the PR if needed @krassowski

krassowski:
> Is this necessary?

I have not verified it, but heard it is.

I opened https://github.com/jupyterlab/jupyter-collaboration/pull/463 for the other issue.

krassowski (approved):
Thank you @brichet!

krassowski:
@meeseeksdev please backport to main

PR #472, opened 2025-04-11
Reset dirty state only when the file is saved

Darshan808:
After https://github.com/jupyterlab/jupyter-collaboration/pull/457 was merged, the current implementation resets the document’s dirty state to False even when the file isn't actually saved (e.g., when the file is read-only).

This PR fixes that behavior by resetting the dirty state only when the file is successfully saved. This prevents confusion in scenarios where changes appear to be saved in the frontend for read-only files, when they actually aren't saved.


davidbrochart (approved):
Thanks @Darshan808.

Issue #477, opened 2025-04-19
Investigate unused (?) code for chat message propagation


krassowski:
> Ok, this one looks like a good catch highlighting another logic issue:
> 
> ```
> projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py:312: error:
> "JupyterWebsocketServer" has no attribute "background_tasks"  [attr-defined]
>                         self._websocket_server.background_tasks.add(task)
>                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py:313: error:
> "JupyterWebsocketServer" has no attribute "background_tasks"  [attr-defined]
>                         task.add_done_callback(self._websocket_server.back...
>                                                ^~~~~~~~~~~~~~~~~~~~~~~~~~~...
> ```
> 
> There is no `background_tasks`, only `_background_tasks`, but the code invokes `background_tasks`:
> 
> https://github.com/jupyterlab/jupyter-collaboration/blob/70d02049aa2f6cacecc91b40b004de2be7224f84/projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py#L307-L313
> 
> This suggests that this might be a dead code. It looks like the chat code was not touched in a while (that line is 2 years old).  I will try changing it to `_background_tasks` and open a new issue.

_Originally posted by @krassowski in https://github.com/jupyterlab/jupyter-collaboration/issues/476#issuecomment-2816669153_

> Ah, not renaming it will not help as the `_background_tasks` is on the handler, not server. `WebsocketServer` used to have `background_tasks` attribute, but it was removed in [y-crdt/pycrdt-websocket@17a2e14](https://github.com/y-crdt/pycrdt-websocket/commit/17a2e147b34dea8b46ced173417df1f386d957ee).
> 
> @davidbrochart are we ok to remove this code altogether? CC @brichet who recently worked on chat in case if that is relevant.

_Originally posted by @krassowski in https://github.com/jupyterlab/jupyter-collaboration/pull/476#issuecomment-2816671013

brichet:
Yes, I thought that we already removed all reference to chat message.
I wonder if that is some remaining code for a first attempt of including a chat in jupyter-collaboration.

brichet:
Probably from https://github.com/jupyterlab/jupyter-collaboration/pull/167

davidbrochart:
[...]

PR #478, opened 2025-04-19
Remove suspected unused code

krassowski:
Checking if any tests fail when removing code highlighted in https://github.com/jupyterlab/jupyter-collaboration/issues/477


davidbrochart (approved):
Thanks!

brichet:
The `CHAT` message type can also be removed, I think
https://github.com/jupyterlab/jupyter-collaboration/blob/a7af554ac4e3865cf397cf57cbd61e4dd9facee3/projects/jupyter-server-ydoc/jupyter_server_ydoc/utils.py#L22

PR #479, opened 2025-04-25
Respect autosave setting in RTC backend

Darshan808:
### Fixes https://github.com/jupyterlab/jupyterlab/issues/14619

Previously, documents were always written to disk on changes, even if autosave was disabled. This PR fixes that by:

- Sending the autosave setting from each client via document awareness.
- Skipping disk writes on doc changes if all connected clients have autosave disabled.
- Writing to disk only when autosave is enabled by at least one client.

The related PR in [`jupyterlab`](https://github.com/jupyterlab/jupyterlab/pull/17508) re-enables manual save in RTC mode. This allows users to save explicitly when autosave is off. On manual save, the frontend sends a `save_to_disc` message via the document's WebSocket provider, triggering a backend save.

**Note**: I didn't find a way to determine which client made the change. So if any connected client has autosave enabled, the document will be saved.

Feedback on the approach is welcome!


projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
```diff
@@ -290,8 +290,15 @@ async def on_message(self, message):
...
-        message_type = message[0]
+        if message == "save_to_disc":
+            try:
+                self.room._saving_document = asyncio.create_task(
+                    self.room._maybe_save_document(self.room._saving_document)
+                )
+            except Exception:
+                self.log.error("Couldn't save content from room: %s", self._room_id)
 
+        message_type = message[0]
```
krassowski, line 301: Picked up a conflict here as the code checking for chat message type was removed

packages/docprovider-extension/src/filebrowser.ts
```diff
@@ -75,12 +76,21 @@ export const rtcContentProvider: JupyterFrontEndPlugin<ICollaborativeContentProv
...
         );
       }
+      const docmanagerSettings = settingRegistry
+        ? await settingRegistry.load('@jupyterlab/docmanager-extension:plugin')
+        : null;
+
+      // Force autosave to be true by default initially
+      if (docmanagerSettings) {
+        void docmanagerSettings.set('autosave', true);
+      }
```
krassowski, line 86: Wouldn't this change user's autosave setting? I think we could just take the value as-is because autosave is the default in lab and notebook.

krassowski:
> Note: I didn't find a way to determine which client made the change. So if any connected client has autosave enabled, the document will be saved.

I think this makes sense. If others agree, I wonder if we should make it clear in the UI or at very least document it somehow.

davidbrochart:
Having autosave enabled when at least one client wants it and disabled when all clients wants it doesn't make sense to me, that's why there was no choice but having autosave enabled in the first place. I have never seen this kind of behavior anywhere. Correct me if I'm wrong but all collaborative applications have autosave enabled (Google Docs...)?

packages/docprovider-extension/src/filebrowser.ts
Darshan808: Oh, if autosave is set to true by default, then we don't need this.
Let me go ahead and remove it.

Darshan808:
I think we can decide how to handle autosave when multiple clients are working on a document. However, what makes the most sense to me is to be able to disable autosave when the extension is installed, but I am working alone on the document.

krassowski:
> Having autosave enabled when at least one client wants it and disabled when all clients wants it doesn't make sense to me, that's why there was no choice but having autosave enabled in the first place. I have never seen this kind of behavior anywhere. 

The difference is you can install docprovider without using RTC, this is just to have:
- server-side execution (and offline execution notifications which follows)
- history timeline
- better completion with jupyter-ai

Forcing non-RTC users to use autosave in that scenario is not good because they may (and in fact have, which is why we opened this PR) ran into IO limitations with large enough notebooks stalling high-performance workloads.

So I think what we want to achieve is to:
- respect autosave preferences (on/off, interval) when only one user is connected (regardless of the number of window open?)
- possibly force autosave (if needs be) when multiple users (not clients) connect; this could be enabled by a plugin in collaboration-ui package which could be disabled if users really don't like it, and which could add a flare in UI to indicate that autosave is active (toolbar? statusbar?) with explanation on hover that it was auto-enabled because of multi-user RTC.

> Correct me if I'm wrong but all collaborative applications have autosave enabled (Google Docs...)?

Quoting my earlier comment from a month ago https://github.com/jupyterlab/jupyterlab/issues/14619#issuecomment-2789035456:

> I think the pattern is to enforce autosave in cloud-synced documents, not in collaborative documents; yes most collaborative documents are cloud-synced nowadays, but in cases where they are not, the user still has to deal with their file system limitations and auto-save can lead to inadvertent side-effects, for example if users have file system watch scripts, such as performing expensive anti-virus scans on each modification (which is sometimes enforced and user has no way to disable it).


Databricks also run in issues with autosaving of large notebooks and automatically disables autosave for notebooks larger than 8 MB, see https://kb.databricks.com/notebooks/notebook-autosave

Here is an interesting pattern from OnlyOffice:

> #### How saving works
> 
> You can decide when you want your changes sent to Document Server. Find the Autosaving option in the File tab -> Advanced settings:
> 
> If autosaving is on, your changes are sent to Document Server (the editors cache) automatically.
> If it’s off you need to click the Save button to save your changes in the editors’ cache.

> #### Saving during co-editing
> The editors have two co-editing modes – Fast and Strict and they do have influence on autosaving.
> 
> In Strict mode, you lock the paragraph you are working on. Others can’t see your changes until you click the Save button, and you can’t see theirs. In this mode, when you click Save your changes are sent to Document Server as usual.
> 
> In Fast mode, you can see everything your co-authors are typing in real-time. In this mode, you don’t need to click Save at all – all the changes are saved automatically the second you stop typing. The Save button remains inactive.


https://www.onlyoffice.com/blog/2020/04/save-and-force-save-in-onlyoffice-never-lose-a-document

Another, one is Collabora Office where auto-save is enabled by default but can be disabled (and interval can be configured).

projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
```diff
@@ -291,6 +292,16 @@ async def on_message(self, message):
         """
         On message receive.
         """
+        if message == "save_to_disc":
```
davidbrochart, line 295: I think that we should use a custom message type (see [here](https://github.com/yjs/y-protocols/blob/master/PROTOCOL.md#combining-protocols)). Maybe `2` followed by `save`?

packages/docprovider/src/ydrive.ts
```diff
@@ -123,7 +126,7 @@ export class RtcContentProvider
       const provider = this._providers.get(key);
 
       if (provider) {
-        // Save is done from the backend
+        provider.wsProvider?.ws?.send('save_to_disc');
```
davidbrochart, line 129: I think that we should use a custom message type (see [here](https://github.com/yjs/y-protocols/blob/master/PROTOCOL.md#combining-protocols)). Maybe `2` followed by `save`?
Also, should we wait for a reply indicating that the file has indeed been saved? Otherwise the following `get` will probably not return the state of the saved file.

mlucool:
FWIW I strongly agree with the idea that you don't always want auto-save with RTC. It seems to me that many things are benefiting from RTC's design of moving state to the backend that are not collaborative environments. In those cases, users would like Jupyter to work as it did before, but also allow for many of the benefits of moving state to the server side.

davidbrochart:
Technically no state is moved to the server when using CRDTs, the state is just distributed among all peers, the server being just one of them. But yes in a future where jupyter-collaboration makes it in Jupyter core, one will use them even when "collaborating with oneself", and I can see an interest in deciding when to save to disk.

packages/docprovider/src/ydrive.ts
```diff
@@ -150,6 +161,19 @@ export class RtcContentProvider
     if (typeof options.format !== 'string') {
       return;
     }
+    // Set initial autosave value, used to determine backend autosave (default: true)
+    const autosave =
+      (this._docmanagerSettings?.composite?.['autosave'] as boolean) ?? true;
+
+    sharedModel.awareness.setLocalStateField('autosave', autosave);
```
krassowski, line 168: Should this also include `autosaveInterval`?

Darshan808: Do you think we should also send the `autosaveInterval` to the backend? 

The reason I ask is that currently, JupyterLab (frontend) automatically calls `save` after every `autosaveInterval`, which in turn triggers `save_to_disc` on the backend.  
However, the current issue is that even with autosave enabled, the backend saves the document to disk **on every document change**, regardless of the configured autosave interval.

If we want to respect the autosave interval properly, I think we could consider removing the `_on_document_change` function on the backend.

https://github.com/jupyterlab/jupyter-collaboration/blob/9059310bb7d6b2e27ab9b5dc5989fb4c332a4678/projects/jupyter-server-ydoc/jupyter_server_ydoc/rooms.py#L235-L237

Here's why:
- When autosave is **off**, only manual saves are possible, so observing document changes for saving might not be necessary.
- When autosave is **on**, the client will already call `save` at the configured interval, which will trigger `save_to_disc`. Thus, observing document changes on the backend may also be redundant in this case.

One potential caveat is that if multiple clients with different `autosaveInterval` values are connected to the same document, `save_to_disc` will still be called for each of them when their individual autosave timers trigger.

Would love to hear your thoughts on this approach and whether you see any concerns or alternatives.

CC: @davidbrochart

krassowski: Maybe we should move the discussion to a new issue and address this in a separate PR?

Darshan808: I think that is reasonable.

krassowski (approved):
This looks good to me!

projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
```diff
@@ -291,6 +293,17 @@ async def on_message(self, message):
         """
         On message receive.
         """
+        message_type = message[0]
+        if message_type == MessageType.UPDATE:
+            msg = message[2:].decode("utf-8")
```
davidbrochart, line 298: Could you remind me why it's not:
```suggestion
            msg = message[1:].decode("utf-8")
```

packages/docprovider/src/ydrive.ts
```diff
@@ -21,10 +21,19 @@ import {
...
 import { Awareness } from 'y-protocols/awareness';
+import { ISettingRegistry } from '@jupyterlab/settingregistry';
+import * as encoding from 'lib0/encoding';
 
 const DISABLE_RTC =
   PageConfig.getOption('disableRTC') === 'true' ? true : false;
 
+const SAVE_MESSAGE = (() => {
+  const encoder = encoding.createEncoder();
+  encoding.writeVarUint(encoder, 2);
```
davidbrochart, line 32: Since we have a named constant on the Python side, maybe we could have something similar on the JS side instead of `2`?

projects/jupyter-server-ydoc/jupyter_server_ydoc/utils.py
```diff
@@ -19,6 +19,7 @@
 class MessageType(IntEnum):
     SYNC = 0
     AWARENESS = 1
+    UPDATE = 2
```
davidbrochart, line 22: I'm not sure `UPDATE` is a good name, since it's similar to a `SYNC_UPDATE` message. Why not just `RAW`, since it's really an unstructured message on top of which we implement a custom protocol?

projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
Darshan808: Because `message[1]` contains the length of the string (written by `writeVarString`), so you must skip it to start decoding the actual string.

davidbrochart: I think that we should use a [Decoder](https://github.com/y-crdt/pycrdt/blob/7f9de2a87c38668b97585d778258912cbf9e4e6b/python/pycrdt/_sync.py#L163) to decode that message.

davidbrochart (approved):
Thanks @Darshan808!

krassowski:
Hi, I'm planing to release this as 4.1 later today, anything else we would like to include?

davidbrochart:
I guess we could include #482?

krassowski:
> I guess we could include https://github.com/jupyterlab/jupyter-collaboration/pull/482?

My reading of that PR was that it waits for tests to be added pending merge and release of https://github.com/jupyter-server/jupyter_server/pull/1518 but reading the contents it looks like it should likely work fine to merge and release and follow-up with tests separately. CC @brichet

Issue #430, opened 2025-01-16
Make the use of `YStore` configurable


dlqqq:
### Problem

By default, `jupyter_server_ydoc` persists every Yjs update made by every client connected to the server. These are stored in `.jupyter_ystore.db` by the `SQLiteYStore` class (defined in `pycrdt-websocket`). This feature allows a document's edits to be "replayed" from its origin, and also allows undo/redo history to be persisted across server restarts.

However, the use of a YStore has led to a number of stability & usability issues in `jupyter_collaboration`. Specifically:

- By default, there is no limit to the size of the YStore in `.jupyter_ystore.db`. Even in single-user scenarios, this file can grow large.

    - In my local Jupyter AI v3 folder, this store grew to 28 MB in just one month of light use 5x a week.

- The rate at which the DB file grows is proportional to the number of active users per server, as that determines the rate of incoming Yjs updates. This means that if a server has 10 users, the DB file will grow 10x as fast relative to the single-user case.

    - We have directly verified that the YStore can grow to ~100MB given ~5-10 users making updates over a few weeks. This figure seems like an exaggeration, but makes sense when you consider the vast number of actions which can trigger Yjs updates (e.g. typing a single character, running a cell in a notebook).

- When RTC is enabled, it can take a long time to open a file for the first time after starting a server. This is because when a file is opened for the first time, `jupyter_server_ydoc` must query a ~10-100MB SQLite file, gather every single Yjs update for that file, then replay all those updates in-memory. This is documented in an upstream issue: https://github.com/jupyter-server/pycrdt-websocket/issues/41

- Finally, the YStore implementation adds [a lot of complexity](https://github.com/jupyter-server/pycrdt-websocket/blob/2b122c0c17e0e530d484840a8a06aeee13250490/pycrdt_websocket/ystore.py#L309), which increases the risk of usability bugs. For example, @andrii-i helped identify that #390 was being caused by the YStore database migration logic not working.

cc @krassowski @afshin @andrii-i @3coins @Zsailer

### Proposed Solution

1. Add a configurable, boolean traitlet which enables/disables the use of `YStore`.

2. (contentious) This configurable should default to `False`, i.e. `YStore` should not be used by default.

I proposed taking these actions at the JupyterLab Frontends call on 01/15. While we seemed to agree about point 1), we were not able to reach a consensus on point 2).

[...]

PR #481, opened 2025-04-30
Add configurables to control ystore.db history and cleanup frequency

Darshan808:
### Description

This PR introduces two new configurables to `SQLiteYStore` class which are introduced in https://github.com/y-crdt/pycrdt-store/pull/2:

- `squash_history_older_than`: Maximum age (in seconds) of document history to retain in the YStore. Older entries are pruned. Defaults to None (no pruning).
- `squash_no_more_often_than`: Minimum interval (in seconds) between automatic cleanup operations. Defaults to 60 seconds.

These options provide better control over the size of `ystore.db`, which can grow significantly over time.


Darshan808:
From discussion in https://github.com/jupyterlab/jupyter-collaboration/issues/430, I think we should set some default value for `history_length` (it's currently `None`, meaning no history trimming).
Open to suggestions on this.

Issue #350, opened 2024-09-11
Changing Username in Jupyter Lab and Setting User-Specific Colors


reyemb:
Hello,

I am looking for a way to customize the username in Jupyter Lab, so that instead of "Anonymous," a more identifiable name is displayed.

I came across the following potential solution: [GitHub Issue #11745](https://github.com/jupyterlab/jupyterlab/issues/11745), but it didn’t seem to work in my case. Additionally, it would be great if I could assign specific colors to users as well.

Managing users via URL parameters is an acceptable approach for us.

For reference, I start Jupyter Lab with the following command:

`jupyter lab --ip=0.0.0.0 --allow-root --no-browser --NotebookApp.token='ultrasecrettoken'`

I am using python version 

Python 3.12.6

My environment has the following packages installed

`anyio                     4.4.0
argon2-cffi               23.1.0
argon2-cffi-bindings      21.2.0
arrow                     1.3.0
asttokens                 2.4.1
async-lru                 2.0.4
attrs                     24.2.0
babel                     2.16.0
beautifulsoup4            4.12.3
bleach                    6.1.0
certifi                   2024.8.30
cffi                      1.17.1
charset-normalizer        3.3.2
comm                      0.2.2
debugpy                   1.8.5
decorator                 5.1.1
defusedxml                0.7.1
executing                 2.1.0
fastjsonschema            2.20.0
fqdn                      1.5.1
h11                       0.14.0
httpcore                  1.0.5
httpx                     0.27.2
idna                      3.8
ipykernel                 6.29.5
ipython                   8.27.0
isoduration               20.11.0
jedi                      0.19.1
Jinja2                    3.1.4
json5                     0.9.25
jsonpointer               3.0.0
jsonschema                4.23.0
jsonschema-specifications 2023.12.1
jupyter_client            8.6.2
jupyter_collaboration     2.1.2
jupyter_core              5.7.2
jupyter-events            0.10.0
jupyter-lsp               2.2.5
jupyter_server            2.14.2
jupyter_server_fileid     0.9.3
jupyter_server_terminals  0.5.3
jupyter-ydoc              2.1.1
jupyterlab                4.2.5
jupyterlab_pygments       0.3.0
jupyterlab_server         2.27.3
MarkupSafe                2.1.5
matplotlib-inline         0.1.7
mistune                   3.0.2
nbclient                  0.10.0
nbconvert                 7.16.4
nbformat                  5.10.4
nest-asyncio              1.6.0
[...]

PR #482, opened 2025-05-06
Add dialog to update user info

brichet:
This PR allows the user to update its attributes name, display_name, initials, color and avatar_url.

The updatable fields are send by the server, starting from https://github.com/jupyter-server/jupyter_server/pull/1518.

It is backward compatible, the fields will not be updatable if the property is not provided by the server.

[record-2025-05-07_07.58.20.webm](https://github.com/user-attachments/assets/87c114aa-bf9e-478b-a35d-242817f89c35)

Fixes:
- https://github.com/jupyterlab/jupyter-collaboration/issues/350
- https://github.com/jupyterlab/jupyterlab/issues/14327

### TODO
- [ ] Add test on updating user info, needs https://github.com/jupyter-server/jupyter_server/pull/1518 to be merged and released

### DIFF
--- a/packages/collaboration/src/components.tsx
+++ b/packages/collaboration/src/components.tsx
@@ -96,7 +96,8 @@ export class UserDetailsBody extends ReactWidget {
   };
 
   render() {
-    if (!this._userManager.identity) {
+    const identity = this._userManager.identity;
+    if (!identity) {
       return <div className="jp-UserInfo-Details">Error loading user info</div>;
     }
     const updatableFields = (this._userManager.permissions?.[
@@ -105,7 +106,7 @@ export class UserDetailsBody extends ReactWidget {
 
     return (
       <div className="jp-UserInfo-Details">
-        {Object.keys(this._userManager.identity).map((field: string) => {
+        {Object.keys(identity).map((field: string) => {
           const id = `jp-UserInfo-Value-${field}`;
           return (
             <div key={field} className="jp-UserInfo-Field">
@@ -117,7 +118,7 @@ export class UserDetailsBody extends ReactWidget {
                 onInput={(event: React.ChangeEvent<HTMLInputElement>) =>
                   this._onChange(event, field)
                 }
-                defaultValue={this._userManager.identity![field] as string}
+                defaultValue={identity[field] as string}
                 disabled={!updatableFields?.includes(field)}
               />
             </div>
--- a/packages/collaboration/src/components.tsx
+++ b/packages/collaboration/src/components.tsx
@@ -82,8 +82,15 @@ export class UserDetailsBody extends ReactWidget {
    */
   private _onChange = (
     event: React.ChangeEvent<HTMLInputElement>,
-    field: keyof User.IIdentity
+    field: string
   ) => {
+    const updatableFields = (this._userManager.permissions?.[
+      'updatable_fields'
+    ] || []) as string[];
+    if (!updatableFields?.includes(field)) {
+      return;
+    }
+
     this._userUpdate[field as keyof Omit<User.IIdentity, 'username'>] =
       event.target.value;
   };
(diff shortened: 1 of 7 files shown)


davidbrochart (approved):

packages/collaboration/src/components.tsx
```diff
@@ -2,31 +2,138 @@
...
+            <div key={field} className="jp-UserInfo-Field">
+              <label htmlFor={id}>{field}</label>
+              <input
+                type={'text'}
+                name={field}
+                id={id}
+                onInput={(event: React.ChangeEvent<HTMLInputElement>) =>
+                  this._onChange(event, field)
+                }
+                defaultValue={this._userManager.identity![field] as string}
```
krassowski, line 120: Saving the attribute on a variable helps typescript to remember that it is a non-null object and should remove the need for the non-null assertion:

```suggestion
    const identity = this._userManager.identity;
    if (!identity) {
      return <div className="jp-UserInfo-Details">Error loading user info</div>;
    }
    const updatableFields = (this._userManager.permissions?.[
      'updatable_fields'
    ] || []) as string[];

    return (
      <div className="jp-UserInfo-Details">
        {Object.keys(identity).map((field: string) => {
          const id = `jp-UserInfo-Value-${field}`;
          return (
            <div key={field} className="jp-UserInfo-Field">
              <label htmlFor={id}>{field}</label>
              <input
                type={'text'}
                name={field}
                id={id}
                onInput={(event: React.ChangeEvent<HTMLInputElement>) =>
                  this._onChange(event, field)
                }
                defaultValue={identity[field] as string}
```

packages/collaboration/style/sidepanel.css
```diff
@@ -142,3 +142,37 @@
...
+
+/************************************************************
+                User Info Details
+*************************************************************/
+.jp-UserInfo-Field {
+  display: flex;
+  justify-content: space-between;
+}
+
+.jp-UserInfo-Field > * {
```
krassowski, line 154: Can the `*` be replaced with a more precise selector? `*` has negative performance implications as every DOM element will be evaluated against the path.

brichet: I updated it with the expected tags.

packages/collaboration/src/components.tsx
brichet: Thanks

packages/collaboration/src/userinfopanel.tsx
```diff
@@ -38,30 +39,87 @@ export class UserInfoPanel extends Panel {
...
 
+  private onClick = () => {
+    if (!this._userManager.identity) {
+      return;
+    }
+    showDialog({
+      body: new UserDetailsBody({
+        userManager: this._userManager
+      }),
+      title: 'User Details'
```
krassowski, line 83: Do we want to translate this title and the error message below?

packages/collaboration/src/userinfopanel.tsx
```diff
@@ -41,27 +58,79 @@ export class UserInfoPanel extends Panel {
...
+  private _translation: IRenderMime.TranslationBundle;
   /**
    * Constructs a new settings widget.
    */
-  constructor(user: User.IIdentity) {
+  constructor(props: UserInfoProps) {
     super();
-    this._user = user;
+    this._userManager = props.userManager;
+    this._translation = props.translation;
```
krassowski, line 73: This needs to be named `_trans`:

```suggestion
    this._trans = props.translation;
```

Because it gets extracted using very simple static analysis, see https://jupyterlab.readthedocs.io/en/latest/extension/internationalization.html#rules

brichet: Do you mean renaming also the property name in the options ?

krassowski: > but I thought the name was required when using it only.

Yes, that's correct. My bad I misread the extend of changes in your previous commit.

brichet: :smile: took me 5 min reading again, to be sure I was not missing something again.

brichet:
Thanks for the review @krassowski and @davidbrochart

PR #490, opened 2025-06-10
Fix confirmation on manual saving

### LINKED ISSUE #485: RTC - Save the file on the disk (Auto-save)
Hi @Darshan808, @krassowski, @davidbrochart!

**Issue**:
During **RTC** in Jupyter Lab started using: `jupyter lab --collaborative`
- When adding, editing, deleting cells in a Notebook, during **RTC**, the JSON of the Notebook in the disk is not changing despite manual save using both `cmd + s` or `File > Save Notebook`.
- Although the UI shows `Saving completed` message, the `dot` remains visible in the Notebook tab indicating that the file is not saved on the disk.
- When opening Jupyter Lab, a pop-up shows up to build `krassowski/jupyterlab-lsp`.

**What is done**:
- Have download the `.zip` of the [Jupyter Collaboration](https://github.com/jupyterlab/jupyter-collaboration) with the latest changes of the [#479](https://github.com/jupyterlab/jupyter-collaboration/pull/479).
-  Have download the `.zip` of the [Jupyterlab](https://github.com/jupyterlab/jupyterlab) with the latest changes of the [#17595](https://github.com/jupyterlab/jupyterlab/pull/17595).
- Install both of them with `pip install .` as the releases were not triggered for both the repositories, i.e., changes are not included in the last releases.

**Goal**:
- Are the savings happen in a latent space?
- Ideal use case: when saving `cmd + s`, the file will be saved in the disk with the latest changes.

P.S.: (And of course, I understand how easy it is to raise issues compared to the effort it takes to maintain and merge PRs—thank you all for your work!)

krassowski:
### References

Fixes #485 and addresses https://github.com/jupyterlab/jupyterlab/pull/17508#pullrequestreview-2863834659.

### Code changes

In response to the raw `save` message on websocket the server now sends a raw message with a JSON string containing:

```ts
{
  'response-to': 'save';
  'status': 'success' | 'skipped' | 'failed';
}
```

The choice of a JSON structure over a simple `status` string is because:
- there are other raw messages being broadcast from the server, we need to distinguish ours somehow
- in the future we may want to include `hash` in the successful response to avoid an extra round trip, which we currently do by consecutively calling `this.get(localPath, fetchOptions)`.

The `RtcContentProvider.save` methods now waits for the response from the server, potentially throwing an error if server returns a failed or skipped status which then indicates that saving failed in the UI.

### User-facing changes

Previously (in `4.1.0rc0`) when saving a document manually (e.g. by pressing <kbd>Ctrl</kbd> + <kbd>S</kbd>) errors were ignored and status bar was incorrectly showing "saving completed" as soon as saving started, even when it was still in progress.

With this PR:
- the errors on backend are propagated to frontend
- the saving status is correctly reflected as in progress/completed depending on server reply

### DIFF
--- a/packages/docprovider/src/ydrive.ts
+++ b/packages/docprovider/src/ydrive.ts
@@ -11,6 +11,7 @@ import {
   ServerConnection,
   User
 } from '@jupyterlab/services';
+import { PromiseDelegate } from '@lumino/coreutils';
 import { ISignal, Signal } from '@lumino/signaling';
 
 import { DocumentChange, ISharedDocument, YDocument } from '@jupyter/ydoc';
@@ -22,6 +23,7 @@ import {
 } from '@jupyter/collaborative-drive';
 import { Awareness } from 'y-protocols/awareness';
 import { ISettingRegistry } from '@jupyterlab/settingregistry';
+import * as decoding from 'lib0/decoding';
 import * as encoding from 'lib0/encoding';
 
 const DISABLE_RTC =
@@ -136,7 +138,47 @@ export class RtcContentProvider
       const provider = this._providers.get(key);
 
       if (provider) {
-        provider.wsProvider?.ws?.send(SAVE_MESSAGE);
+        const ws = provider.wsProvider?.ws;
+        if (ws) {
+          const delegate = new PromiseDelegate<void>();
+          const handler = (event: MessageEvent) => {
+            const data = new Uint8Array(event.data);
+            const decoder = decoding.createDecoder(data);
+            try {
+              const messageType = decoding.readVarUint(decoder);
+              if (messageType !== RAW_MESSAGE_TYPE) {
+                return;
+              }
+            } catch {
+              return;
+            }
+            const rawReply = decoding.readVarString(decoder);
+            let reply: {
+              'response-to': 'save';
+              status: 'success' | 'skipped' | 'failed';
+            } | null = null;
+            try {
+              reply = JSON.parse(rawReply);
+            } catch (e) {
+              console.debug('The raw reply received was not a JSON reply');
+            }
+            if (reply && reply['response-to'] === 'save') {
+              if (reply.status === 'success') {
+                delegate.resolve();
+              } else if (reply.status === 'failed') {
+                delegate.reject('Saving failed');
+              } else if (reply.status === 'skipped') {
+                delegate.reject('Model update was already in progress');
+              } else {
+                delegate.reject('Unrecognised save reply');
+              }
+            }
+          };
+          ws.addEventListener('message', handler);
+          ws.send(SAVE_MESSAGE);
+          await delegate.promise;
+          ws.removeEventListener('message', handler);
+        }
         const fetchOptions: Contents.IFetchOptions = {
           type: options.type,
           format: options.format,
--- a/projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
+++ b/projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
@@ -15,7 +15,7 @@ from jupyter_server.auth import authorized
 from jupyter_server.base.handlers import APIHandler, JupyterHandler
 from jupyter_server.utils import ensure_async
 from jupyter_ydoc import ydocs as YDOCS
-from pycrdt import Doc, UndoManager
+from pycrdt import Doc, Encoder, UndoManager
 from pycrdt_websocket.yroom import YRoom
 from pycrdt_websocket.ystore import BaseYStore
 from tornado import web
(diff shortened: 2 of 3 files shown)


krassowski:
@davidbrochart does it look good to you too? If so I would love to merge, cut another RC and then final.

projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
```diff
@@ -301,14 +301,34 @@ async def on_message(self, message):
             if msg == "save":
                 try:
                     room = cast(DocumentRoom, self.room)
-                    room._save_to_disc()
+                    save_task = room._save_to_disc()
+                    # task may be missing if save was already in progress
```
davidbrochart, line 305: It looks like `_save_to_disc()` returns `None` when updating the document's `dirty` flag, not when saving is in progress?

packages/docprovider/src/ydrive.ts
```diff
@@ -136,7 +138,47 @@ export class RtcContentProvider
...
+            } catch (e) {
+              console.debug('The raw reply received was not a JSON reply');
+            }
+            if (reply && reply['response-to'] === 'save') {
+              if (reply.status === 'success') {
+                delegate.resolve();
+              } else if (reply.status === 'failed') {
+                delegate.reject('Saving failed');
+              } else if (reply.status === 'skipped') {
+                delegate.reject('Model update was already in progress');
```
davidbrochart, line 171:
```suggestion
                delegate.reject('Saving already in progress');
```

davidbrochart (commented):
I'm not sure about where this is going. It looks like manual saving could have been done through HTTP?

projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
krassowski: Yes, I was not sure how we should handle it. The lock is used right after save here:

https://github.com/jupyterlab/jupyter-collaboration/blob/1198452a900ed3f0ce2bdefb496d1ce8dd59c924/projects/jupyter-server-ydoc/jupyter_server_ydoc/rooms.py#L308-L318

It is also used when saving fails due to out of band changes:

https://github.com/jupyterlab/jupyter-collaboration/blob/1198452a900ed3f0ce2bdefb496d1ce8dd59c924/projects/jupyter-server-ydoc/jupyter_server_ydoc/rooms.py#L335-L337

and during initialization:

https://github.com/jupyterlab/jupyter-collaboration/blob/1198452a900ed3f0ce2bdefb496d1ce8dd59c924/projects/jupyter-server-ydoc/jupyter_server_ydoc/rooms.py#L113-L115

so ignoring initialization:

https://github.com/jupyterlab/jupyter-collaboration/blob/1198452a900ed3f0ce2bdefb496d1ce8dd59c924/projects/jupyter-server-ydoc/jupyter_server_ydoc/rooms.py#L273-L278

will only short-circuit during saves, but does guarantee that every concurrent save will short circuit.

projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
```diff
@@ -301,14 +301,34 @@ async def on_message(self, message):
...
+                    await self.send(
+                        self._encode_json_message({"response-to": "save", "status": "failed"})
+                    )
             return
 
         self._message_queue.put_nowait(message)
         self._websocket_server.ypatch_nb += 1
 
+    def _encode_json_message(self, message: dict) -> bytes:
+        RAW_MESSAGE_TYPE = 2
```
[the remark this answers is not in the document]
krassowski: Yup, even better we can use `MessageType.RAW`.

krassowski:
> It looks like manual saving could have been done through HTTP?

Yes, it feels like it http would give us a round trip for free, but I think the convenience of using websocket is that `YDocWebSocketHandler` already has all the context (and locks), otherwise we need to share the locks between tornado handlers, and I always feel uneasy as to how robust that is (easy to make a mistake).

krassowski:
> But then should messages include an ID, so that we know to which message a reply refers to?

That's right, if there are multiple documents saved at once with "save all" the save replies would get mixed up with the current implementation. Thank you, I will add an ID.

krassowski:
No, wait there is one websocket per document, so worst case would be multiple save actions on the same doc getting mixed. I still can add an ID but it seems less critical.

davidbrochart:
I don't want to nitpick if you want to release soon, I was just wondering if the raw message that is introduced is generic enough so that it can be used for other types of message. But it's fine to merge this PR as-is.

krassowski:
What makes more sense for jupyverse? If you were to add manual save there, would you like it to be over ws or an extra http handler? I think that the addition of manual save request `docprovider` forces `jupyverse` to implement follow the same approach.

krassowski:
Would you like to approve it, or are there any changes I can make to improve it? I am happy to change the format of the messages, maybe make it use JSON both ways to make it more future-proof?

krassowski:
@davidbrochart just checking back here - I would love to unblock the release :)

davidbrochart:
Sorry, I'm on a sick leave but I'll try to catch up soon.

PR #499, opened 2025-09-03
Define `ICollaborativeContentProvider` as inheriting from `IContentProvider`

### LINKED ISSUE #498: `ICollaborativeContentProvider` should likely inherit from `IContentProvider`
> That seems to have gotten us past that error, but now:
> 
> ```
> Property 'get' does not exist on type 'ICollaborativeContentProvider'.
> ```
> 
> 🤔

Hmm the implementation actually does have a `get()` method. This is due to the interface of `ICollaborativeContentProvider` not inheriting from `IContentProvider`. I think we may want to expose it if this is desirable.

_Originally posted by @krassowski in https://github.com/geojupyter/jupytergis/issues/848#issuecomment-3246567754_

https://github.com/jupyterlab/jupyter-collaboration/blob/67453e04dad30978d42fdef07040ae94cabe2bf0/packages/collaborative-drive/src/tokens.ts#L25

krassowski:
Fixes #498

### DIFF
--- a/packages/collaborative-drive/src/tokens.ts
+++ b/packages/collaborative-drive/src/tokens.ts
@@ -2,7 +2,11 @@
 // Distributed under the terms of the Modified BSD License.
 
 import { IAwareness } from '@jupyter/ydoc';
-import { Contents, IContentProvider, SharedDocumentFactory } from '@jupyterlab/services';
+import {
+  Contents,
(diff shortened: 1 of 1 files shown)


Darshan808 (approved):
This was needed.
Thanks!

PR #519, opened 2025-11-28
Enable `TempFileYStore` to work with Jupyter Lab

### LINKED ISSUE #483: Error when using a different YStore.
### Description

When trying to use a different YStore implementation via command line configuration, I'm encountering an error.
The [documentation](https://jupyterlab-realtime-collaboration.readthedocs.io/en/latest/configuration.html) shows we can specify a YStore class:
```
jupyter lab --YDocExtension.ystore_class=pycrdt_websocket.ystore.TempFileYStore
```

However, when using this, I get the following error:
```
__init__ got an extra keyword argument config
```

### Problem Details

The `config` argument seems to be automatically passed via Jupyterlab. 
The default `SQLiteYStore` implementation passes the `config` argument to `LoggingConfigurable`, not to the base store (according to its MRO).
When specifying a different `YStore` via command line, the `config` is automatically passed to the class's `__init__` method
Classes from `pycrdt_store` don't expect or handle this `config` parameter

### Potential Solutions

- Add `**kwargs` to all `__init__` methods in `pycrdt_store` classes to accept the extra configuration argument
- Create wrapper implementations for all stores from `pycrdt_store` in `jupyter-server-ydoc` package, customize them (e.g., modifying DB paths and others) and provide `LoggingConfigurable` to all similar to the default `SQLiteYStore` and point to use these classes in documentation rather than using pycrdt-store's classes.

The second approach seems cleaner and would ensure all store implementations have the same behavior and would be inside same package i.e `jupyter-server-ydoc`.

Open to feedback on this.

Darshan808:
## Fixes #483 

This PR updates `TempFileYStore` so it can be used via command-line configuration.

The class now:

- Inherits from `LoggingConfigurable`. (Config argument now gets passed here).
- Uses a combined metaclass just like `SQLiteYStore`

With this change, `TempFileYStore` behaves consistently like `SQLiteYStore` when selected from the CLI.

### DIFF
--- a/projects/jupyter-server-ydoc/jupyter_server_ydoc/stores.py
+++ b/projects/jupyter-server-ydoc/jupyter_server_ydoc/stores.py
@@ -7,7 +7,11 @@ from traitlets import Int, Unicode
 from traitlets.config import LoggingConfigurable
 
 
-class TempFileYStore(_TempFileYStore):
+class TempFileYStoreMetaclass(type(LoggingConfigurable), type(_TempFileYStore)):  # type: ignore
+    pass
+
+
+class TempFileYStore(LoggingConfigurable, _TempFileYStore, metaclass=TempFileYStoreMetaclass):
     prefix_dir = "jupyter_ystore_"
 
 


krassowski (commented):
I think the only sane way to test this is to run `ui-tests` with a separate startup script to pass configuration like we do for timeline

https://github.com/jupyterlab/jupyter-collaboration/blob/main/ui-tests/playwright.config.js
https://github.com/jupyterlab/jupyter-collaboration/blob/main/ui-tests/playwright.timeline.config.js

https://github.com/jupyterlab/jupyter-collaboration/blob/e52b14495b46c2719bfded61c7aee4dc56a76a88/.github/workflows/test.yml#L294-L298

(without the need to skip tests conditionally, but just running one test that would fail before this PR)

davidbrochart (approved):

PR #525, opened 2025-12-04
`RtcContentProvider`: depend on `DefaultDrive` rather than on the `RestContentProvider` for routing `get()` and `save()`

martinRenou:
This was making jupyter-collaboration bound to JupyterLab's implementation and default drive.

This PR gets rid of it so we rely on the current drive's default get implementation instead.

Note: On top of https://github.com/jupyterlab/jupyter-collaboration/pull/524 due to missing drive typings that are not in <4.5


krassowski:
The release went through after setting up trusted publishers: https://github.com/jupyterlab/jupyter-collaboration/releases/tag/v4.2.0

PR #475, opened 2025-04-18
Fix cleanup logic and suppress repetitive error logs

Darshan808:
## Description
This PR addresses two issues in the file-watching and room cleanup logic:

### 1. Fix room cleanup condition
The condition `self.room.clients == [self]` was incorrect because `self.room.clients` is a `set`, not a `list`.  
It has been updated to `self.room.clients == {self}` to ensure proper comparison.  
This change ensures that when a room becomes empty, it is now correctly cleaned up after the default delay (60 seconds).

### 2. Add error log suppression
To avoid spamming error logs during the cleanup delay window, a `consecutive_error_logs` counter and suppression mechanism have been added.  
This ensures that only the first few errors are logged, and further repetitive logs are suppressed with a single warning.

PR #529, opened 2025-12-11
Add a test for file watcher cleanup after delay

krassowski:
Follow-up to https://github.com/jupyterlab/jupyter-collaboration/pull/475, adding a test to make sure this does not break in the future.

The structure of the test is based on `test_room_handler_doc_client_should_cleanup_room_file` below.


davidbrochart:
So `pycrdt-store` v0.1.3 seems to be the culprit?

krassowski:
Yes... `document_ttl` was deprecated in favour of `squash_after_inactivity_of` so maybe the deprecation warning interferes with the test? I will open a separate PR to fix it tomorrow.

https://github.com/jupyterlab/jupyter-collaboration/blob/611b9532977c0aa4739b4147f42be2999bbf8e92/tests/test_app.py#L67-L77

PR #539, opened 2026-01-29
Use `serverSettings` for API requests

### LINKED ISSUE #538: Frontend does not respect auth settings for websockets
## Description

Websockets are tricky to handle wrt auth, especially when running in a cross-origin setup.

JupyterLab allows to configure `ServerSettings` to append a token (default), but the websocket connections here do not follow the JupyterLab best practices and ignore any settings https://jupyterlab.readthedocs.io/en/latest/extension/extension_dev.html#service-manager-plugins

https://github.com/jupyterlab/jupyter-collaboration/blob/64b5f4c6f0c3c5101d773bc5c9f3873b9b143c39/packages/docprovider/src/yprovider.ts#L104-L114

## Expected

We should append the token if one is set in `ISettings.token` and if `ISettings.appendToken` is true as in:

https://github.com/jupyterlab/jupyterlab/blob/4b15fcf7b0aa9128117d83766076c2bfe83ccf40/packages/services/src/event/index.ts#L104-L121

or

https://github.com/jupyterlab/jupyterlab/blob/4b15fcf7b0aa9128117d83766076c2bfe83ccf40/packages/services/src/kernel/default.ts#L1454-L1466

Additionally, the access to `wsUrl` should be deferred until connection, this is doing it here:

https://github.com/jupyterlab/jupyter-collaboration/blob/fdf409bd4df3c5fd41e8e61dbb1de0d586cf0596/packages/docprovider/src/ydrive.ts#L247-L250

is too early and prevents changing server during runtime.

See https://github.com/jupyter/nbdime/pull/793 as an example.

Same for awareness websocket.

Ideally `ISettings.init.credentials` would be respected too.

krassowski:
Respect `IServerSettings` service token

- Closes #538

### DIFF
--- a/packages/docprovider/src/yprovider.ts
+++ b/packages/docprovider/src/yprovider.ts
@@ -130,7 +130,8 @@ export class WebSocketProvider implements IDocumentProvider, IForkProvider {
       {
         disableBc: true,
         params,
-        awareness: this._awareness
+        awareness: this._awareness,
+        WebSocketPolyfill: this._serverSettings.WebSocket
       }
     );
 
@@ -152,7 +153,8 @@ export class WebSocketProvider implements IDocumentProvider, IForkProvider {
       {
         disableBc: true,
         params,
-        awareness: this._awareness
+        awareness: this._awareness,
+        WebSocketPolyfill: this._serverSettings.WebSocket
       }
     );
   }


krassowski:
@jtpio would you potentially be interested in reviewing this one (since you were making similar changes elsewhere)?

PR #543, opened 2026-02-06
Disambiguate log messages

Carreau:
Two log messages are identical; even if technically you can update the logger to display the file and line number, I think it is better to have slightly different log messages so you can follow what is heppening.


krassowski (approved):
Thank you @Carreau!

PR #544, opened 2026-02-06
Use jupyter-ydoc's `aget`/`aset`

davidbrochart:


krassowski:
```
> fatal: couldn't find remote ref refs/pull/544/merge
```

I saw this a couple of times when GitHub API acts up. Let's see if reopening helps.

PR #514, opened 2025-11-13
fix: clean up background tasks on close

agoose77:
This PR attempts to close #503 and fix #161 by cleaning up background tasks upon closure of the websocket. I'm not sure if this fits into the wider design space for resource management, but it does fix "hang on shutdown" on my system.

PR #546, opened 2026-02-12
FIX: prevent hanging on server shutdown

### LINKED ISSUE #161: Jupyter Collaboration prevents JupyterLab from being killed from the terminal
# Issue

Installing `jupyter_collaboration` results in not being able to kill JupyterLab from the terminal.

# Reproduce

1. Install `jupyterlab` and `jupyter_collaboration`
2. Create a Notebook, so that you create at least one room
3. Try to kill jupyterlab with `Ctrl + C` in the terminal

# Attempts to fix the issue

After some investigation, it looks like [this `sys.exit` call](https://github.com/jupyterlab/jupyterlab/blob/main/jupyterlab/__main__.py#L8) does not work as expected even though it properly raises a `SystemExit`.

One reason for `sys.exit` to not work can be a hanging thread or subprocess that has not been terminated.

My investigations so far have been:
- Removed [the `_confirm_exit` thread](https://github.com/jupyter-server/jupyter_server/blob/d48843eea78fe2ee37608a767ceea3d0d2f71763/jupyter_server/serverapp.py#L2199) from jupyter_server to see if that thread did not prevent from finishing
- Refactored the tasks cleanup in ypy-websockets and jupyter_collaboration
  - https://github.com/y-crdt/ypy-websocket/pull/76
  - https://github.com/jupyterlab/jupyter_collaboration/pull/160
- Remove any `AnyIO` usage in ypy-websockets. I tried this because there seems to be an "AnyIO worker thread" thread hanging around upon termination. 

None of these attempts fixed the issue. Though I think the refactor of the tasks cleanup could be good to include anyway.

ianhi:
fixes #161 supersedes #514 

---
🤖 full disclosure this a PR that I used claude for nearly in its entirety. I have reviewed every line to the best of my ability and confirmed that the fix works for me locally. I did multiple local rounds of review, tightening up exceptions, logic and improving comments. But ultimately I learned more about async through this PR than I knew at the start, so very possible that there is a subtlety i missed.

To debug I ended up designing a loop for claude where it used [jupyter-collab-mcp](https://ianhuntisaak.com/jupyterlab-collab-mcp/) to interact with the notebooks (creating rooms crucial for causing hte bug) and then sent sigterms from a shell with `PYTHONFAULTHANDLER=1` and looped through possible fixes before getting it working. I then guiding claude through coming up with a minimal fix.

---

There were three issues preventing shutdown on receiving ctrl-C

## 1 client serve loop never finishing

the server was stopped. but client rooms were never stopped. We needed to sent a stop message to the clients to cause their loops to stop
https://github.com/jupyterlab/jupyter-collaboration/blob/c750ddd1ac2fcebbeaab1aaf261930a2e475ae27/projects/jupyter-server-ydoc/jupyter_server_ydoc/websocketserver.py#L65-L75

## 2 Rooms never being stopped
rooms were persisted after a client disconnect (I think for allowing reconnect). But on shutdown they were never being stopped, so the task was staying alive. now we explcitly stop all rooms.

This option means they weren't being autoclosed on disconnect (which I think is good)
https://github.com/jupyterlab/jupyter-collaboration/blob/c750ddd1ac2fcebbeaab1aaf261930a2e475ae27/projects/jupyter-server-ydoc/jupyter_server_ydoc/app.py#L119-L120

## 3 the root task was never awaited
because we didn't hold on to a reference to the server start task there was no way to await it and ultimately trigger the worker threads were never cleaned up. now we save a reference and await it at the end of `clean`
https://github.com/jupyterlab/jupyter-collaboration/blob/c750ddd1ac2fcebbeaab1aaf261930a2e475ae27/projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py#L82-L84


davidbrochart (commented):
This is why structured concurrency is helpful. Jupyverse's RTC implementation is much simpler and doesn't have this bug.

ianhi:
Thank you for the review! I cleaned up based on your suggestions. Two follow ups:

1. I'm down to use systems that work better! But how would jupyverse have fixed this, per the docs it still would have used this library, and I thought the issues were internal? (This may be showing my async noviceness)

2. Im noticing now that I get clean shutdowns with multiple notebooks, but not when I have a jupyter terminal open. any ideas on that?

krassowski:
@ianhi do you think this is worth merging and releasing or is (2) a blocker?

PR #558, opened 2026-03-26
Make document provider pluggable

davidbrochart:
Closes #557.

~~This makes the document provider configurable through settings, where one can choose the WebSocket provider (the default, as before this PR) or the WebRTC provider. For WebRTC, one can also provide a list of signaling server URLs, for instance https://flyio-signaling-server.fly.dev/. Otherwise, the signaling server will default to the Jupyter server's URL at path `api/signaling`, which can be backed by a new `jupyter-server-signaling` Python package provided in this PR.~~

~~Notable differences with the WebSocket provider:~~
- ~~the WebRTC provider is a serverless solution to real-time collaboration (if we omit the signaling server), which means that Jupyter Collaboration can work with JupyterLite.~~
- ~~documents are not automatically saved after each change, but explicitly by users pressing Ctrl-S (as in non-RTC).~~
- ~~document renames (through Jupyter) or out-of-band changes are not supported yet.~~

~~There are the following bugs:~~
- ~~when a document opened by one peer is "dirty" (not saved) and another peer opens the document, they will not see it as dirty initially.~~
- ~~Creating a new notebook doesn't populate it with one cell (it's empty).~~

This makes the document provider pluggable through provider factory extension points. Jupyter Collaboration provides a WebSocket provider, that could be replaced e.g. with a WebRTC provider.


krassowski (commented):
Initial skimming

packages/y-webrtc/package.json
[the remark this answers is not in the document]
davidbrochart: Maybe we could have a TypeScript version here?

davidbrochart:
Apart from the fact that `@jupyter/y-webrtc` has not been published, there is this error:
```
Error: src/awareness.ts(120,10): error TS2339: Property 'destroy' does not exist on type 'WebRTCAwarenessProvider'.
```

davidbrochart:
I'll remove the WebRTC provider to have this PR just make the provider pluggable. The WebRTC provider can then be provided by another extension.

SylvainCorlay:
Awesome!

tsconfig.json
```diff
@@ -13,6 +13,7 @@
     "noUnusedLocals": true,
     "preserveWatchOutput": true,
     "resolveJsonModule": true,
+    "skipLibCheck": true,
```
krassowski: Can we drop this now?

packages/docprovider-extension/src/filebrowser.ts
```diff
@@ -55,12 +56,13 @@ export const rtcContentProvider: JupyterFrontEndPlugin<ICollaborativeContentProv
...
     requires: [ITranslator],
-    optional: [IGlobalAwareness, ISettingRegistry],
+    optional: [IGlobalAwareness, ISettingRegistry, IDocumentProviderFactory],
     activate: async (
       app: JupyterFrontEnd,
       translator: ITranslator,
       globalAwareness: Awareness | null,
-      settingRegistry: ISettingRegistry | null
+      settingRegistry: ISettingRegistry | null,
+      providerFactory: IDocumentProviderFactory
```
krassowski: Either `| null` or needs to go to `requires` if it is always required (which appears to be the case)

SylvainCorlay:
I would love to see this hook integrated soon!

packages/docprovider-extension/src/filebrowser.ts
davidbrochart: Thanks, it is required: https://github.com/jupyterlab/jupyter-collaboration/pull/558/commits/e21edc685523a19c4b50c249b3efe16199eb486c.

krassowski: That commit did not change argument order, I think tests should fail now. Maybe worth adopting [`@jupyter/eslint-plugin`](https://github.com/jupyterlab/eslint-plugin) that would have caught it earlier?

krassowski: In particular https://eslint-plugin.readthedocs.io/en/latest/rules/plugin-activation-args/

tsconfig.json
davidbrochart: If I do I get an `error TS2315: Type 'Uint8Array' is not generic`. According to [your comment](https://github.com/jupyterlab/jupyter-collaboration/pull/558#discussion_r3016214032), I pinned down `lib0`: https://github.com/jupyterlab/jupyter-collaboration/pull/558/commits/bab2e5126b90a30d7f0a431d2b5c36dc42bbe4ff.

packages/docprovider-extension/src/filebrowser.ts
davidbrochart: I just fixed the argument order for now: https://github.com/jupyterlab/jupyter-collaboration/pull/558/commits/f001f47ec98955902475ffc28d64f6599963a804.

packages/docprovider-extension/src/yprovider.ts
krassowski: I understand the rationale for moving `WebSocketProvider` and `WebSocketAwarenessProvider` out of `docprovider` - with multiple providers being first class it no longer deserves a special treatment. I have three questions:
1. is it backwards-incomptible since users who were importing it from `@jupyter/docprovider` will no longer get it and thus we need to bump version to major?
2. does it belong in `docprovider-extension` or yet another, non-extension package? My thinking is that this is still something that downstreams may want to reuse and remix, so maybe we should have a new `@jupyter/websocketprovider` package; then if we wanted to avoid a major version bump we could re-export these two from `@jupyter/docprovider`, which would be not ideal as it would still pull the code when installing just `@jupyter/docprovider` for everyone, but we could mark it as deprecated and drop the dependency and re-export whenever we feel like making a breaking major version release
3. apart from architectural purity, is there a reason to move these two classes out of `@jupyter/docprovider`

packages/docprovider-extension/src/yprovider.ts
```diff
@@ -349,3 +428,172 @@ export namespace WebSocketProvider {
...
+  create(options: IAwarenessProviderFactory.IOptions) {
+    const url = URLExt.join(
+      this._serverSettings.wsUrl,
+      'api/collaboration/room'
+    );
+    return new WebSocketAwarenessProvider({
+      url,
+      roomID: options.roomID,
+      awareness: options.awareness,
+      user: this._user
```
krassowski: Should it use the `user` from `options`? Currently it is ignored. If yes then we can drop `private _user: User.IManager,` from line 550; if not, then we can drop `user` from `IAwarenessProviderFactory.IOptions` interface.

```suggestion
      user: options.user
```

krassowski:
> I would love to see this hook integrated soon!

Me too!

packages/docprovider-extension/src/yprovider.ts
davidbrochart: > 1. is it backwards-incomptible since users who were importing it from `@jupyter/docprovider` will no longer get it and thus we need to bump version to major?

Yes I think it will be backwards-incompatible and it deserves a major version bump. Actually since working on this PR (when it had the WebRTC provider) led to https://github.com/jupyterlab/jupyterlab/pull/18739, I think it will use `jupyter-ydoc` v4 and it's a good occasion to bump both packages.

For 2. and 3., I was thinking that the extension point is now the doc provider factory, and the WebSocket and WebRTC doc provider factories should be provided by extensions. The former should be provided by `@jupyter/docprovider`, and the WebSocket provider factory should be provided by `@jupyter/docprovider-extension` as a "default" for `jupyter-collaboration`, while still allowing to be swapped by another extension (maybe `@jupyter/webrtcprovider`). But I agree that we could create a new package `@jupyter/websocketprovider` instead of moving the WebSocket provider factory to `@jupyter/docprovider-extension`.

krassowski: The use cases I would like to consider is when downstream wants to:
- reuse `jupyter-server-ydoc` websocket but delegate that to a different jupyter-server than the main one (maybe because the document lives elsewhere, maybe to isolate the ydoc process and thus provide additional resilience)
- reimplement websocket handling using slightly different approach (compatible with but not 1-1, with  `jupyter-server-ydoc`), e.g. adding additional handshake for auth; basically they would want to inherit from `WebSocketProvider` and tweak some behaviour beyond changing the URL.

I saw both use cases in practice and this motivates my question on "how we can import `WebSocketProvider` for reuse" going forward. I see two options:
- downstream adds `@jupyter/docprovider-extension` dependency and imports `WebSocketProvider` from there; in this case we should add an export in `packages/docprovider-extension/src/index.ts` and think about it as a public API going forward; this is a bit awkward because downstream is pulling an extension package from which it does not actually intend to use any extensions
- create a new package (we said `@jupyter/websocketprovider` but maybe `@jupyter/websocket-docprovider` or `@jupyter/docprovider-websocket` would be better names)  which seems like the cleanest approach, although adds a little bit more overhead to this PR and future maintenance

martinRenou: Do we really need to move this class? I guess it's harmless to keep it under `@jupyter/docprovider` and that way this changes is backward compatible and people can continue importing it.

martinRenou: > create a new package (we said @jupyter/websocketprovider but maybe @jupyter/websocket-docprovider or @jupyter/docprovider-websocket would be better names) which seems like the cleanest approach, although adds a little bit more overhead to this PR and future maintenance

If we really should move it, that's the most sensible approach to me

martinRenou: @krassowski should I take your thumb-up as an agreement with keeping it under `@jupyter/docprovider` ? I can push those changes if you do agree. 

@davidbrochart is on vacation this week, we'd love to go ahead with the proposed changes, and I guess making that PR friction-less by making it backward compatible would be good in that sense

krassowski: Yes, I think that if it is possible to keep it under `@jupyter/docprovider` for now, then we should do it and then we probably can release it without major version bump.

martinRenou: Done moving back the interfaces!

packages/collaboration-extension/src/collaboration.ts
```diff
@@ -89,24 +87,25 @@ export const menuBarPlugin: JupyterFrontEndPlugin<void> = {
...
-    const server =
-      app.serviceManager.serverSettings ?? ServerConnection.makeSettings();
-    const url = URLExt.join(server.wsUrl, 'api/collaboration/room');
-
-    new WebSocketAwarenessProvider({
-      url: url,
+    const awarenessOptions = {
       roomID: 'JupyterLab:globalAwareness',
       awareness: awareness,
       user: user
```
krassowski: The server settings need to be passed down to the factory in options, otherwise if the `wsUrl` is different from the default the websocket will not connect; see https://github.com/jupyterlab/jupyter-collaboration/pull/539. Currently:
- `WebSocketAwarenessProviderFactory` takes in `serverSettings` during construction
- `WebSocketDocumentProviderFactory` takes `serverSettings` in options, but they are not passed in here

This should be harmonized, we should either use options pattern, or constructor pattern in both places.

```suggestion
      user: user,
      serverSettings: app.serviceManager.serverSettings
```

krassowski (commented):
Besides the `serverSettings` inconsistency and regression it looks good to me

packages/collaboration-extension/src/collaboration.ts
martinRenou: Makes sense! Thanks. I pushed 68e5af6 and went for the solution to pass server settings through options

packages/docprovider/src/ydrive.ts
```diff
@@ -70,10 +67,11 @@ export class RtcContentProvider implements IContentProvider {
     this._serverSettings = options.serverSettings;
     this._currentDrive = options.currentDrive;
     this.sharedModelFactory = new SharedModelFactory(this._onCreate);
-    this._providers = new Map<string, WebSocketProvider>();
+    this._providers = new Map<string, IDocumentProvider & IForkProvider>();
     this._documentManager = options.documentManager ?? null;
     this._docmanagerSettings = options.docmanagerSettings ?? null;
     this._driveFileChanged = options.fileChanged;
+    this._providerFactory = options.providerFactory;
```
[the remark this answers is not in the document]
krassowski: That's fine I think.

packages/docprovider/src/yprovider.ts
```diff
@@ -104,6 +111,64 @@ export class WebSocketProvider implements IDocumentProvider, IForkProvider {
...
+      encoding.writeVarUint(encoder, saveId);
+      const saveMessage = encoding.toUint8Array(encoder);
+      ws.send(saveMessage);
+      try {
+        await delegate.promise;
+      } finally {
+        ws.removeEventListener('message', handler);
+      }
+    }
+  }
```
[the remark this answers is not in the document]
krassowski: This is not new

packages/docprovider/src/yprovider.ts
```diff
@@ -104,6 +111,64 @@ export class WebSocketProvider implements IDocumentProvider, IForkProvider {
...
+      encoding.writeVarUint(encoder, RAW_MESSAGE_TYPE);
+      encoding.writeVarString(encoder, 'save');
+      encoding.writeVarUint(encoder, saveId);
+      const saveMessage = encoding.toUint8Array(encoder);
+      ws.send(saveMessage);
+      try {
+        await delegate.promise;
+      } finally {
+        ws.removeEventListener('message', handler);
+      }
```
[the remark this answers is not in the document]
krassowski: Pre-existing.

packages/collaboration-extension/src/collaboration.ts
```diff
@@ -89,24 +87,26 @@ export const menuBarPlugin: JupyterFrontEndPlugin<void> = {
...
   description: 'Add global awareness to share working document of users.',
-  requires: [IStateDB],
+  requires: [IStateDB, IAwarenessProviderFactory],
   provides: IGlobalAwareness,
-  activate: (app: JupyterFrontEnd, state: StateDB): IAwareness => {
+  activate: (
+    app: JupyterFrontEnd,
+    state: StateDB,
+    factory: IAwarenessProviderFactory
+  ): IAwareness => {
```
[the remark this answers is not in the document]
krassowski: This is expected

krassowski (approved):
Thank you @davidbrochart and @martinRenou! Good to go once CI is green.

martinRenou:
Thank you for the reviews!

PR #562, opened 2026-04-02
Export editor facet to allow settings cursors from third party extensions

brichet:
This PR export the `editorAwarenessFacet` that is used to display remote cursor and selection.

AFAIU this `Facet` is used to ensure that the local and remote ytext are the same.
Exporting it allow to "simulate" a remote user and display its cursor, for e.g. https://github.com/jupyterlite/ai/issues/298, where there is no remote user but another "local" user.


krassowski:
I'm not sure if the proposed use case is a good justification. Frontend extensions can populate the awareness room with cursors of other local users, I think jupyterlite/ai should not need to acces the facet. That said I have a use case of https://github.com/jupyterlab/jupyter-collaboration/issues/547 where I currently monkey patch the package to get access to the facet. It should not be needed if there was a proper API for https://github.com/jupyterlab/jupyter-collaboration/issues/547 but likely I would still want to implement it differently downstream. So at the end of the day I think it is worth exporting it as it provides downstream an escape hatch.

krassowski (approved):

brichet:
> I'm not sure if the proposed use case is a good justification. Frontend extensions can populate the awareness room with cursors of other local users, I think jupyterlite/ai should not need to acces the facet.

Right, tested again and it seems not required. I had an issue when trying to set the cursors immediately after opening the notebook. The comparison below was not working:
https://github.com/jupyterlab/jupyter-collaboration/blob/1224ee4d89ba4bd771506c63c06600780bb4f6cd/packages/collaboration/src/cursors.ts#L307

davidbrochart (approved):

PR #566, opened 2026-04-09
Bump yjs to 13.6.30 (fixes GC crash in applyUpdate)

xrl:
## Summary

Bumps yjs from 13.6.20 to 13.6.30 and lib0 from 0.2.98 to 0.2.99 in `yarn.lock`. The semver range (`^13.5.40`) already allows this — only the lockfile pin changes.

This fixes a crash during `applyUpdate` → `tryGcDeleteSet` → `findIndexSS` that throws **"Unexpected case"** on every page load when collaboration is enabled.

Three bugs in yjs < 13.6.25 contribute:

- **13.6.25** — `iterateDeletedStructs` iterates past known struct bounds (no clock guard)
- **13.6.26** — null guard for missing client structs in `iterateDeletedStructs`
- **13.6.30** — `sortAndMergeDeleteSet` mutates `DeleteItem.len` in place instead of creating a new object (yjs/yjs#767), corrupting shared references used by GC

## How we found this

We decompiled the minified yjs bundle in JupyterLab's static assets and traced the error through `findIndexSS` → `replaceStruct` → `Item.gc` → `ContentType.gc` (the full GC hierarchy). The old buggy pattern `left.len = math.max(left.len, ...)` was present at offset 30350 in the bundle, matching the pre-13.6.30 source.

## Reproduction

Any JupyterLab instance with `jupyter-collaboration` enabled that has accumulated deleted items in a Yjs document will hit this on page load during the initial document sync. The error is caught by the docprovider (`console.error("Caught error while handling a Yjs update", e)`) so it doesn't crash, but it breaks document sync for the affected document.

## Test plan

- [x] `jlpm test` — 16 JS tests pass
- [x] `pytest tests/` — 52 Python tests pass
- [x] `jlpm build` — all 5 packages build successfully
- [x] Only version changes: yjs 13.6.20→13.6.30, lib0 0.2.98→0.2.99


krassowski (changes requested):
This PR cannot be merged as it includes files which should not have been committed, and extensive changes to `yarn.lock` beyond what is necessary. I am happy to merge a PR bumping yjs version but it shuld be a minimal self-contained change.

krassowski (changes requested):
This commits generated files which should not be committed. It also makes changes to bundling without explanation and containe unnecessary changes in the yarn lock file.

PR #489, opened 2025-06-06
Modify API to optionally create a document room if it does not exist.

Darshan808:
### Description
This PR adds support for creating a document room if it does not already exist.

- Introduces a new `create` argument to the `get_document` method.
- If the room for a given document does not exist and `create=True`, a new room is created.
- The document is then returned with the newly created room.
---

Looking forward to feedbacks!

PR #568, opened 2026-04-22
Add create argument to get_document API

### LINKED ISSUE #328: An API to create a document room if it does not exist?
### Problem

Currently it is possible for other extensions to get the document (or it's copy - by default) if there is a collaborative session for it. However, it is not currently possible to create the document room when it does not exist. This would be useful to allow server-side execution for notebooks which are not currently open (e.g. user clicks from context menu in the file browser on the notebook and chooses to execute it).

### Proposed Solution

Add a new public API for creating the notebook. This could be as simple as a new argument to the existing `get_document`, a boolean `create` which would default to false.

### Additional context

https://github.com/jupyterlab/jupyter-collaboration/blob/bcc36e1a451aaa5dbfbe020db72070149bd934bd/projects/jupyter-server-ydoc/jupyter_server_ydoc/app.py#L136-L145

abbbe:
### Description
Adds a `create` parameter to `get_document()` so that a document room is created if it does not already exist. This enables other extensions to obtain a document for server-side operations (e.g. execution) even when no collaborative session is open.

Based on #489 by @Darshan808, incorporating review feedback from @krassowski and @davidbrochart:
- Room locks are shared between `YDocExtension` and `YDocWebSocketHandler` via `defaultdict(asyncio.Lock)` to prevent race conditions during concurrent room creation.
- Room creation logic is extracted into `_create_room()` helper for clarity.

### Changes
- **`app.py`**: New `_create_room()` helper; `get_document()` gains `create: bool = False` parameter; `_room_locks` shared with handlers.
- **`handlers.py`**: `_room_locks` moved from class-level to instance-level, received via `initialize()` for backward compatibility.
- **`test_app.py`**: New `test_get_document_create_room` test.

### DIFF
--- a/projects/jupyter-server-ydoc/jupyter_server_ydoc/app.py
+++ b/projects/jupyter-server-ydoc/jupyter_server_ydoc/app.py
@@ -3,8 +3,9 @@
 from __future__ import annotations
 
 import asyncio
+from collections import defaultdict
 from functools import partial
-from typing import Literal
+from typing import Literal, cast
 
 from jupyter_server.extension.application import ExtensionApp
 from jupyter_ydoc import ydocs as YDOCS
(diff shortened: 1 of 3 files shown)


abbbe:
This is actually not useful PR, in fact #489 has it all

PR #569, opened 2026-04-22
Add `create: bool` argument to `get_document()` API

### LINKED ISSUE #328: An API to create a document room if it does not exist?

abbbe:
## Fixes #328

Rebase of #489 by @Darshan808 onto current main, with import ordering fix for ruff.

CI on fork: https://github.com/abbbe/jupyter-collaboration/pull/1 — all Python tests pass including Ubuntu/Python 3.13 (which was failing on #489). There are 3 failing checks: check_release, Test Python (macos-latest, 3.14), Test Python (windows-latest, 3.14t).


krassowski (approved):
Well, it looks to be passing now indeed. Thanks for the help with rebasing and reviving this!

abbbe:
Great.

What will happen next -- will you take care of merging things into main and closing all relevant PRs and issues, or any actions needed from me?

Any clue when these changes will be included into a package released on pypi?

krassowski:
Yes, nothing else needed here.

> Any clue when these changes will be included into a package released on pypi?

Probably not today because it is Friday. We have a number of PRs that we would like to release soon so probably next week

projects/jupyter-server-ydoc/jupyter_server_ydoc/app.py
```diff
@@ -207,13 +215,54 @@ async def get_document(
...
-        else:
-            room_id = room_id
 
-        try:
-            room = await self.ywebsocket_server.get_room(room_id)
-        except RoomNotFound:
-            return None
+        if not self.ywebsocket_server.started.is_set():
+            asyncio.create_task(self.ywebsocket_server.start())
+            await self.ywebsocket_server.started.wait()
```
krassowski: @Darshan808 do you remember why these three lines were added here? I wonder if we should push them down to within the `except RoomNotFound` once we confirmed that we want to create the room, otherwise we might be starting the websocket server that is not needed?

projects/jupyter-server-ydoc/jupyter_server_ydoc/app.py
```diff
@@ -207,13 +215,54 @@ async def get_document(
...
+        async with self._room_locks[room_id]:
+            try:
+                room = await self.ywebsocket_server.get_room(room_id)
+            except RoomNotFound:
+                if not create:
+                    return None
```
krassowski:
```suggestion
        async with self._room_locks[room_id]:
            try:
                room = await self.ywebsocket_server.get_room(room_id)
            except RoomNotFound:
                if not create:
                    return None

                if not self.ywebsocket_server.started.is_set():
                    asyncio.create_task(self.ywebsocket_server.start())
                    await self.ywebsocket_server.started.wait()                
```

projects/jupyter-server-ydoc/jupyter_server_ydoc/app.py
```diff
@@ -207,13 +215,55 @@ async def get_document(
...
+
+                if not self.ywebsocket_server.started.is_set():
+                    asyncio.create_task(self.ywebsocket_server.start())
+                    await self.ywebsocket_server.started.wait()
```
krassowski:
```suggestion
                    await self.ywebsocket_server.started.wait()
```

krassowski:
UI failure are unrelated, down to pre-release of JupyterLab which removed simple mode switch.

projects/jupyter-server-ydoc/jupyter_server_ydoc/app.py
Darshan808: > I wonder if we should push them down to within the except RoomNotFound once we confirmed that we want to create the room

Yes this makes sense.

PR #570, opened 2026-04-22
Fix the share button in the top area

jtpio:
Currently the share button shows the full label which takes a lot of valuable space and confuses users.

**Before**

﻿
image

**After**

image

While still making sure the label shows up in the command palette:

image


brichet (approved):
Thanks @jtpio

davidbrochart (approved):

Issue #567, opened 2026-04-09
Investigate proposed fix for content duplication from JSD


krassowski:
See https://github.com/jupyter-ai-contrib/jupyter-server-documents/issues/210. I am not sure how applicable it is here but definitely worth looking into.

Issue #545, opened 2026-02-10
Jupyter Collaboration intermittently causes whole document to duplicate


alasdair-macleod:
## Description

This behaviour has been described before, but I am currently unable to find the issues that relate to it.
Collaboration appears to register the same user as more than one, duplicate the entire document after which validate(nb) updates all of the non unique cell_ids

The following (relatively old) sample logs are from an example with one user using one browser.
I have been trying for some time to find a more detailled explanation - but this appears to start when YDocExtension registers more users than are actually connected.

[I 2025-10-03 11:54:27.190 YDocExtension] Processed 29 Y patches in one minute | jupyter-cwjc51lxuvhnc05le5ww82
[I 2025-10-03 11:54:27.190 YDocExtension] Connected Y users: 2
...
...
validate(nb)	jupyter-cwjc51lxuvhnc05le5ww82
/opt/conda/lib/python3.12/site-packages/nbformat/__init__.py:132: DuplicateCellId: Non-unique cell id 'e99af923' detected. Corrected to 'b12208b6'.	jupyter-cwjc51lxuvhnc05le5ww82
validate(nb)	jupyter-cwjc51lxuvhnc05le5ww82
/opt/conda/lib/python3.12/site-packages/nbformat/__init__.py:132: DuplicateCellId: Non-unique cell id 'a46d6d05' detected. Corrected to '5829dbf8'.	jupyter-cwjc51lxuvhnc05le5ww82
validate(nb)	jupyter-cwjc51lxuvhnc05le5ww82
/opt/conda/lib/python3.12/site-packages/nbformat/__init__.py:132: DuplicateCellId: Non-unique cell id 'be0e838c' detected. Corrected to '1c479935'.	jupyter-cwjc51lxuvhnc05le5ww82
validate(nb)	jupyter-cwjc51lxuvhnc05le5ww82
... (etc. until fininshed)

We run a cloud based service with tens of thousands of users. This issue is intermittent and hard to reproduce, but does happen to a significant number of our users. 

## Reproduce

I have found this issue very hard to reproduce. 
Opening and closing the file repratedly has sometimes made it possible for me to replicate this.
The one common thread is that YDocExtension starts to register more users than are actually connected, despite being connected only once.
I have managed to replicate this running jupyter in a docker container on a single machine (i.e. not in a cloud environment)

## Context

- Operating System and version: appears to be agnostic to operating system, various versions of Linux, MacOS and Windows tested
[...]

PR #571, opened 2026-04-23
Fix content duplication when document room goes away

krassowski:
- Addresses https://github.com/jupyterlab/jupyter-collaboration/issues/545

This builds on JSD investigation (https://github.com/jupyterlab/jupyter-collaboration/issues/567, https://github.com/jupyter-ai-contrib/jupyter-server-documents/pull/215), but uses a different strategy to fix things.

- [x] Add a test case which should fail in first commit
- [x] Fix the issue


krassowski:
@jordanhboxer would you be able to test this patch?

krassowski:
Also CC @dlqqq who may be interested

projects/jupyter-server-ydoc/jupyter_server_ydoc/rooms.py
```diff
@@ -19,6 +20,20 @@
...
 
+async def _apply_deterministic_source_content(ydoc: Doc, file_type: str, content: Any) -> None:
+    """Load source content using a deterministic update.
+
+    Rooms rebuilt from disk must recreate the same Yjs history for identical
+    content, otherwise reconnecting clients can merge duplicate content from a
+    divergent history after server restart or room eviction.
+    """
+
+    source_ydoc: Doc = Doc(client_id=0)
```
davidbrochart: Do we need to specify a `client_id`?
Also, it might be better for this function to be a method of `DocumentRoom`?

krassowski: Yes, we need to specify `client_id`, that's the critical bit. See https://discuss.yjs.dev/t/initial-offline-value-of-a-shared-document/465/26 - this is basically a variation on the workaround that Kevin suggested earlier, here https://discuss.yjs.dev/t/initial-offline-value-of-a-shared-document/465/6 (same topic, earlier post) that he discouraged; however this version does not modify client ID of the final doc (which AFAIK was waht was discouraged), only uses a temporary doc to create the initial update. If you remove it, the test is correctly failing with:

```
E       AssertionError: assert 2 == 1
E        +  where 2 = len([{'cell_type': 'code', 'execution_count': None, 'id': 'cell-1', 'metadata': {}, ...}, {'cell_type': 'code', 'execution_count': None, 'id': 'cell-1', 'metadata': {}, ...}])
```

We could use a different magic number though (but it has to be non-negative so zero seems the obvious choice).

I updated the comment in 5a9f84a78e943f4f0984075d8547cc29a961f024

> Also, it might be better for this function to be a method of `DocumentRoom`?

Done in 5a9f84a78e943f4f0984075d8547cc29a961f024

davidbrochart: Thanks for the detailed explanation Mike.

davidbrochart (approved):
Thanks @krassowski.

PR #572, opened 2026-04-27
Require Python 3.10, update classifiers

jtpio:
Minor maintenance check since Python 3.8 and 3.9 are EOL, and update classifiers.

### DIFF
--- a/projects/jupyter-server-ydoc/pyproject.toml
+++ b/projects/jupyter-server-ydoc/pyproject.toml
@@ -17,11 +17,11 @@ classifiers = [
     "Intended Audience :: Science/Research",
     "License :: OSI Approved :: BSD License",
     "Programming Language :: Python",
-    "Programming Language :: Python :: 3.8",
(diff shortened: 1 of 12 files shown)


projects/jupyter-server-ydoc/pyproject.toml
```diff
@@ -55,7 +55,6 @@ test = [
     "pytest-cov",
     "anyio",
     "httpx-ws >=0.9.0",
-    "importlib_metadata >=4.8.3; python_version<'3.10'",
```
krassowski, line 58: Shall we also simplify:

https://github.com/jupyterlab/jupyter-collaboration/blob/73aa34bc36b62fd3b47a8bab3192ab3f0a000bee/tests/test_documents.py#L7-L10

jtpio: Thanks! Yes we can simplify even more due to the Python 3.10 requirement.

krassowski (approved):
Thank you!

PR #574, opened 2026-04-28
Depend on `IDocumentManager` instead of fetching settings directly

### LINKED ISSUE #573: Plugin activation error when `jupyterlab-collaboration-ui` is not installed
## Description

In theory it should be possible to only use some of the underlying packages without `jupyter-collaboration-ui` to be able to use most of the collaboration features, but without the UI elements.

For example with:

```
pip install jupyterlab jupyter-server-ydoc jupyter-docprovider
```

In practice it does not seem to be possible.

## Reproduce

1. `pip install jupyterlab jupyter-server-ydoc jupyter-docprovider`
2. `jupyter lab`

Image

## Expected behavior

No plugin activation error.

## Context

```
jupyterlab                4.5.6
jupyter-docprovider       2.3.0
jupyter-server-ydoc       2.3.0
```

jtpio:
Fixes https://github.com/jupyterlab/jupyter-collaboration/issues/573

Before the content provider plugin in `jupyter-docprovider` was fetching the settings for the core `@jupyterlab/docmanager-extension:plugin` directly. 

The transformer for that plugin id is registered inside the JupyterLab plugin with that same id, during its own activate: https://github.com/jupyterlab/jupyterlab/blob/8152d62832af21a85c483fae5662566afd9e4d1e/packages/docmanager-extension/src/index.tsx#L360

However when `jupyter-collaboration-ui` is not installed, it seems likely that the plugin loading the settings directly was activating earlier, before `@jupyterlab/docmanager-extension:plugin` had time to define its transform.

Which ends up hitting this code path: https://github.com/jupyterlab/jupyterlab/blob/8152d62832af21a85c483fae5662566afd9e4d1e/packages/settingregistry/src/settingregistry.ts#L682-L690

Also the settings were fetched directly to get some values already exposed via `IDocumentManager`.

### DIFF
--- a/packages/docprovider/src/ydrive.ts
+++ b/packages/docprovider/src/ydrive.ts
@@ -3,6 +3,7 @@
 
 import { IChangedArgs, PageConfig } from '@jupyterlab/coreutils';
 import { IDocumentManager } from '@jupyterlab/docmanager';
+import { ISettingRegistry } from '@jupyterlab/settingregistry';
 import { TranslationBundle } from '@jupyterlab/translation';
 import {
   Contents,
@@ -48,7 +49,12 @@ namespace RtcContentProvider {
     trans: TranslationBundle;
     globalAwareness: Awareness | null;
     serverSettings: ServerConnection.ISettings;
-    documentManager: IDocumentManager | null;
+    documentManager?: IDocumentManager | null;
+    /**
+     * @deprecated Pass `documentManager` instead. Kept for backwards
+     * compatibility; will be removed in a future major release.
+     */
+    docmanagerSettings?: ISettingRegistry.ISettings | null;
     currentDrive: Contents.IDrive;
     fileChanged?: ISignal<Contents.IDrive, Contents.IChangedArgs>;
   }
@@ -63,7 +69,7 @@ export class RtcContentProvider implements IContentProvider {
     this._currentDrive = options.currentDrive;
     this.sharedModelFactory = new SharedModelFactory(this._onCreate);
     this._providers = new Map<string, WebSocketProvider>();
-    this._documentManager = options.documentManager;
+    this._documentManager = options.documentManager ?? null;
     this._driveFileChanged = options.fileChanged;
   }
 
--- a/packages/docprovider/src/ydrive.ts
+++ b/packages/docprovider/src/ydrive.ts
@@ -1,7 +1,8 @@
 // Copyright (c) Jupyter Development Team.
 // Distributed under the terms of the Modified BSD License.
 
-import { PageConfig } from '@jupyterlab/coreutils';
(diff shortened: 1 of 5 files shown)


jtpio:
Trying with the built wheels from the check-release looks fine locally, no error message anymore in the dev tools console:

image

packages/docprovider/src/ydrive.ts
```diff
@@ -48,7 +48,7 @@ namespace RtcContentProvider {
     trans: TranslationBundle;
     globalAwareness: Awareness | null;
     serverSettings: ServerConnection.ISettings;
-    docmanagerSettings: ISettingRegistry.ISettings | null;
+    documentManager: IDocumentManager | null;
```
jtpio: Need to check if we are fine with this change here, or consider it breaking.

jtpio: `ISettingRegistry.ISettings` was pretty broad, not sure we could or should try to continue supporting it.

Also `RtcContentProvider` may not be used a lot downstream, so probably fine to make this change here?

jtpio:
Marking as ready so folks can have a look.

packages/docprovider/src/ydrive.ts
krassowski: I would err on the side of caution and keep both for now, but deprecate `ISettingRegistry.ISettings`


```suggestion
    /**
      * @deprecated - will be removed in next major version, pass documentManager instead
      */
    docmanagerSettings?: ISettingRegistry.ISettings | null;
    documentManager?: IDocumentManager | null;
```

Unless we have a consensus that release is major (but I still see discussion continue on https://github.com/jupyterlab/jupyter-collaboration/pull/558#discussion_r3152237769)

packages/docprovider/src/ydrive.ts
```diff
@@ -234,16 +240,26 @@ export class RtcContentProvider implements IContentProvider {
...
-    sharedModel.awareness.setLocalStateField('autosave', autosave);
-
-    // Watch for changes in settings
-    this._docmanagerSettings?.changed.connect(() => {
-      const newAutosave =
-        (this._docmanagerSettings?.composite?.['autosave'] as boolean) ?? true;
-      sharedModel.awareness.setLocalStateField('autosave', newAutosave);
+    sharedModel.awareness.setLocalStateField(
+      'autosave',
+      this._documentManager?.autosave ?? true
```
krassowski: Should also restore `(this._docmanagerSettings?.composite?.['autosave'] as boolean)` for backward compat?

jtpio: Yeah even though it makes the logic a bit more complicated, we should technically be able to do it.

Will push a new commit.

krassowski (approved):
Thank you @jtpio!

jtpio:
Thanks for the review!

PR #580, opened 2026-04-30
Test using stable versions

krassowski:
Addresses UI test failures seen in #569 and #571 due to release of JupyterLab [`v4.6.0a5`](https://github.com/jupyterlab/jupyterlab/releases/tag/v4.6.0a5) which removed the Simple mode toggle from the statusbar (https://github.com/jupyterlab/jupyterlab/pull/18757).


brichet (approved):
Thanks @krassowski, LGTM.

Quick question, should we update jupyterlab version in the following test too, for consistency with the [requirements](https://github.com/jupyterlab/jupyter-collaboration/blob/73778d684f6332c52ca7406e4fc0abf192b34621/projects/jupyter-collaboration/pyproject.toml#L36) ?
https://github.com/jupyterlab/jupyter-collaboration/blob/73778d684f6332c52ca7406e4fc0abf192b34621/.github/workflows/test.yml#L227

krassowski:
Probably yes, but since it is an older version I imagine it would not hurt.

PR #584, opened 2026-05-15
Do not crash on out of band change when room is disconnected and ystore deleted (or no-op)

### LINKED ISSUE #583: Crash is possible if client reconnects after notebook was modified (with evicted YStore and room)
## Description

The recent fix for content duplication works, but only if file does not get modified on disk in between sessions.

```
Cannot apply update: block parent <0#4> must be deleted or shared ref type. Type: 8
```

## Reproduce

1. Open notebook, make some changes
2. Disconnect browser (freeze tab, hibernate laptop, disconnect wifi if using a remote server)
3. Remove YStore database
4. Wait for the room to be evicted
5. Make out of band change (edit file on disk) to the notebook, changing its metadata or cell structure (i.e. make a change which is more than editing text)
6. Restart server
7. Connect browser
8. See error

## Expected behavior

No crash; not sure what we should do really.

## Context

- Fix from #571 
- Released in `v4.4.0`

krassowski:
- [x] Fixes https://github.com/jupyterlab/jupyter-collaboration/issues/583


krassowski:
@davidbrochart what do you think?

projects/jupyter-server-ydoc/jupyter_server_ydoc/rooms.py
```diff
@@ -229,6 +240,73 @@ async def _broadcast_updates(self):
         except asyncio.CancelledError:
             pass
 
+    async def serve(self, channel):
```
davidbrochart: I'm wondering if we could upstream that change to `pycrdt-websocket`? If it's specific to Jupyter, maybe allow registering an error handler in `pycrdt-websocket` in case we cannot apply a change to the shared document, and have only the handler here?

davidbrochart (approved):
I thought that CRDTs were conflict-free but since it's just an additive change... 😄 
Thanks Mike!

krassowski:
> I thought that CRDTs were conflict-free

They are - the problem is that y.js can only guarantee the resolution of conflicts if we have non-divergent history; since we are now forcing the initial client ID to avoid the duplication of content on out-of-band changes, if we lose the history (someone deletes YStore or uses a no-op YStore) and then have OOB with different structure of the document, we lose the ability to resolve conflicts. I think we are hitting the limitations of y.js for OOB handling either way.

The only alternative I can think of is that we store a copy of the document as it was when it was loaded from file. And then each time we load from disk we store another copy. And each time for OOB without YStore we would seed the room as follows:
1. initial OOB with client id 0
2. diff for second OOB with client id -1
3. diff for third OOB with client id -2
4. etc

This _could_ work without ever breaking the sync. It would be somewhat non-deterministic as there is no guarantee to what state it would resolve on re-connect with an external client (y.js treats all clients as having equal claim on conflicts) and would require storing a lot of extra data. That said, I will explore it before merging this one.

davidbrochart:
Deployments of RTC should probably ensure that a YStore exists and is never deleted.
Also, a lot of complexity has been added just for supporting out-of-band changes, but maybe another route could be to prevent them from happening on shared files, e.g. using file locks?

krassowski:
> Deployments of RTC should probably ensure that a YStore exists and is never deleted.

Yes, but we cannot guarantee that, especially because of the size that YStore can accumulate. This is less of a problem for architectures like y-sweet where any large document only brings down one instance. In essence we should recommend using YStore as much as possible but also have a good failure mode for when YStore gets deleted or corrupted (which is not uncommon for SQLite databases on NFS filesystems).

> The only alternative I can think of is that we store a copy of the document as it was when it was loaded from file. [...] That said, I will explore it before merging this one.

Having put more though into it the alternative solution - I believe would work, but be prone to silent corruptions which may be even worse. If implemented, it would effectively have the same downsides as YStore for storage usage (with worst case scenario being worse) and could lead to silent conflicts which would be worse than surfacing the conflict to the user.

krassowski:
@meeseeksdev please backport to 4.4.x

PR #586, opened 2026-05-20
Configurable `session_store_path`

jtpio:
Follow-up to https://github.com/jupyterlab/jupyter-collaboration/pull/548

This makes it possible to configure where the JSON file is saved with the following:

```bash
jupyter lab --YDocExtension.session_store_path=/path/to/sessions.json
```

To avoid leaving git repos in a dirty state:

image

### DIFF
--- a/docs/source/configuration.md
+++ b/docs/source/configuration.md
@@ -25,6 +25,13 @@ looses connection and goes offline for a while. You should never have to touch i
 fine to just ignore it, including in your version control system (don't commit this file). If
 you happen to delete it, there shouldn't be any serious consequence either.
 
+A second file, `collaboration_sessions.json`, is created inside a `.jupyter` directory under
+that same launch directory. It records recent collaboration session IDs so the backend can
+detect when a client is reconnecting after a server restart or version change, and prompt for
+a reload when the session can no longer be resumed safely. Like the YStore database, it is
+safe to ignore in version control and safe to delete (clients will simply be asked to reload
+on next connect).
+
 There are a number of settings that you can change:
 
 ```bash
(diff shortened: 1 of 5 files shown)


jtpio:
Looks like the existing dirty tests are a bit flaky as CI is now passing after a few restarts.

Darshan808 (approved):
Looks good.
Thanks!

docs/source/configuration.md
```diff
@@ -46,6 +53,13 @@ jupyter lab --YDocExtension.document_cleanup_delay=100
 
 # The YStore class to use for storing Y updates (default: JupyterSQLiteYStore).
 jupyter lab --YDocExtension.ystore_class=pycrdt.store.TempFileYStore
+
+# Relocate the YStore SQLite database (default: '.jupyter_ystore.db' in the launch directory).
+jupyter lab --SQLiteYStore.db_path=/path/to/ystore.db
+
+# Relocate the collaboration session store
+# (default: '<server_root_dir>/.jupyter/collaboration_sessions.json').
```
davidbrochart, line 62: Is `server_root_dir` the same as the "launch directory" on line 57? If so, I suggest using the same words.

projects/jupyter-server-ydoc/jupyter_server_ydoc/app.py
```diff
@@ -93,6 +93,17 @@ class YDocExtension(ExtensionApp):
...
     )
 
+    session_store_path = Unicode(
+        None,
+        allow_none=True,
+        config=True,
+        help="""Path to the JSON file used to record collaboration session IDs for
+        reconnect compatibility checks. When unset, defaults to
+        ``<server_root_dir>/.jupyter/collaboration_sessions.json``. Set this to
+        relocate the file (for example into a per-user data directory) so the
```
davidbrochart: What is a "per-user data directory"? It is ambiguous since this `session_store_path` applies to all users connecting to the Jupyter server.

jtpio: Right the main motivation for this new trait was to let users configure where the files get created, mostly in the single-user use case.

Let's see if we can make the description more explicit.

davidbrochart (approved):

PR #587, opened 2026-05-25
Use jupyter-ydoc v4

davidbrochart:
This works with https://github.com/jupyterlab/jupyterlab/pull/18739.


davidbrochart:
Ready for review now that JupyterLab v4.6.0b1 is out.

davidbrochart:
It seems that I cannot make changes to the GitHub workflows.

krassowski:
I think this needs updating to:

https://github.com/jupyterlab/jupyter-collaboration/blob/b1eee5a090127e985a5ef568c3c32edc0f3618db/projects/jupyter-server-ydoc/pyproject.toml#L32

I can push a commit, just surfacing in comments.

packages/collaboration-extension/package.json
```diff
@@ -56,22 +56,22 @@
...
-    "@jupyter/ydoc": "^2.1.3 || ^3.0.0",
-    "@jupyterlab/application": "^4.5.0",
-    "@jupyterlab/apputils": "^4.5.0",
-    "@jupyterlab/codemirror": "^4.5.0",
-    "@jupyterlab/coreutils": "^6.5.0",
-    "@jupyterlab/services": "^7.5.0",
-    "@jupyterlab/statedb": "^4.5.0",
-    "@jupyterlab/translation": "^4.5.0",
-    "@jupyterlab/ui-components": "^4.5.0",
+    "@jupyter/ydoc": "^4.0.0-a3",
```
krassowski: Should we keep double/triple range here?

krassowski (approved):
Thank you!

davidbrochart:
Thanks a lot @krassowski.

davidbrochart:
Should we release v4.4.1a0?

krassowski:
I think so. Or we could wait for https://github.com/jupyterlab/jupyter-collaboration/pull/584 and then release.

krassowski:
> Should we release v4.4.1a0?

I think this needs to be `v4.5.0` at a minimum since we changed JupyterLab dependency to `4.6.0b1`+. Maybe it should be v5.0.0 to allow backporting things (like https://github.com/jupyterlab/jupyter-collaboration/pull/584) to v4.5.0 targeting JupyterLab `4.5.x`?

davidbrochart:
Technically a minor bump of a dependency doesn't require a major bump, right?

krassowski:
No, it does not. Can we release https://github.com/jupyterlab/jupyter-collaboration/pull/584 as 4.4.1 and this #587 as 4.5.0?

davidbrochart:
I mean, I don't have a strong opinion. Maybe jupyter-collaboration v5.0.0 would be clearer and more in-lined with jupyter-ydoc v4, even if it doesn't require it. But we are not going to release JupyterLab v5 even though it now depends on jupyter-ydoc v4, right?
Either way, I'm fine. So we need to create a `4.4.x` branch and have #584 in and #587 out?

krassowski:
Yes, I can take care of it.

davidbrochart:
Just a note that I'd like https://github.com/y-crdt/pycrdt/pull/393 to be released and used in jupyter-ydoc v4 and jupyter-collaboration v5, since it will be another breaking change.

PR #588, opened 2026-05-27
Reduce noisy monitor logs

### LINKED ISSUE #585: The monitor is too verbose (`Processed %s Y patches in one minute`)
## Description

Every 60 seconds the following two logs are logged, even if there was no change since the last time they were emitted:

https://github.com/jupyterlab/jupyter-collaboration/blob/092b288555a0e9c704c390993e444bda19a9d188/projects/jupyter-server-ydoc/jupyter_server_ydoc/websocketserver.py#L143-L159

## Reproduce

1. Start the server extension
2. Wait two minutes

## Expected behavior

1. Only log it if it changed (there are some users/number of users changed, there were some messages processed)?
2. Switch one or both to debug verbosity?

## Context

Initially introduced in:
- https://github.com/jupyterlab/jupyter-collaboration/pull/55

Initially proposed in:
- https://github.com/jupyterlab/jupyter-collaboration/issues/52

CC @ellisonbg @davidbrochart as authors of the relevant issue/PR. @ellisonbg are you relying on these messages being logged every minute at info level?

Darshan808:
### Description
Fixed noisy monitor logs
  - Changed logs to `debug` level
  - Only logs on activity or state changes

### DIFF
--- a/projects/jupyter-server-ydoc/jupyter_server_ydoc/websocketserver.py
+++ b/projects/jupyter-server-ydoc/jupyter_server_ydoc/websocketserver.py
@@ -148,12 +148,16 @@ class JupyterWebsocketServer(WebsocketServer):
         #### Note:
             This method runs in a coroutine for debugging purposes.
         """
+        prev_clients_nb = 0
         while True:
             try:
                 await asyncio.sleep(60)
             except asyncio.CancelledError:
                 break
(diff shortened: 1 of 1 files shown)


krassowski (approved):
Thanks!

davidbrochart (approved):

PR #589, opened 2026-05-29
Add a "Show diff" option when an unresolvable conflict is detected on out of band change

krassowski:
- Continuation of https://github.com/jupyterlab/jupyter-collaboration/pull/584

Motivation: because the notebook CRDT has a complex structure, when the server reloads the notebook from the disk, in absence of a persistent yjs model storage and if the structure of the document changed significantly we unavoidably either:
- get duplicated content
- run into a conflict (if using same client id for initial load to prevent duplication which is what we do since #571)

In a sense this is a problem of schema migration of yjs, except that our "schema" is notebook structure.

While we do have a YStore as a persistent storage, many users disable it or use an in-memory version due to the size it can accumulate over longer usage periods. This gets surfaced by #584, which adds "revert" and "save as" buttons but does not give the user enough context on what changes diverged. This PR introduces a diff view to help user discover what changes they need to reapply.

To consider before undrafting:
- [x] can we fully upstream a Notebook diff widget that does not depend on a file existing to `nbdime`
   we probably can, tackling https://github.com/jupyter/nbdime/issues/534 but it is fine to merge as-is and later swap to a version contributed upstread, because we are not exposing this as a public API
- [x] what do we do about non-Notebook unresolvable conflicts?
   - [x] these will never happen for text files as they are always resolvable
   - [ ] JupyterCAD or GIS however could run into those - do we just not show the diff button for them?


davidbrochart:
> * can we fully upstream a Notebook diff widget that does not depend on a file existing to `nbdime`

We can always recreate temporary files to pass to `nbdime`, right?

> * JupyterCAD or GIS however could run into those - do we just not show the diff button for them?

IIRC JupyterCAD has a diff viewer @trungleduc?

trungleduc:
> IIRC JupyterCAD has a diff viewer @trungleduc?

yes, that's right

krassowski:
> We can always recreate temporary files to pass to nbdime, right?

Technically yes but:
- we would then need to pull the full `nbdime` as a dependency, including the server extension; what I am trying to do here is pull just the typescript widget for rendering diffs.
- we would need to add a new endpoint on the server side or make assumptions about the disk, neither of which seems attractive to me right now

krassowski:
> IIRC JupyterCAD has a diff viewer @trungleduc?

@trungleduc I cannot easily find it ([this search](https://github.com/search?q=repo%3Ajupytercad%2FJupyterCAD+diff&type=code) does not show anything useful) can you point me in the right direction please?

krassowski:
Marking as ready for review - I think we can tackle the JupyterCAD case in a follow-up PR.

davidbrochart:
@martinRenou I don't think JupyterGIS has a diff viewer, right?

krassowski:
In the meantime, I will also add the buttons to the toolbar of the diff viewer because currently the buttons are only in the dialog which means user has no easy way to save local stat once they opted to view diff (or revert to server state).

trungleduc:
> > IIRC JupyterCAD has a diff viewer @trungleduc?
> 
> @trungleduc I cannot easily find it ([this search](https://github.com/search?q=repo%3Ajupytercad%2FJupyterCAD+diff&type=code) does not show anything useful) can you point me in the right direction please?

Hey @krassowski, the diff view of JupyterCAD is buried in its 3D viewer implementation, and it's not easy to make use of it externally

package.json
```diff
@@ -50,7 +51,7 @@
     "test:cov": "lerna run test:cov",
     "test:debug": "lerna run test:debug",
     "test:debug:watch": "lerna run test:debug:watch",
-    "watch": "lerna run watch"
+    "watch": "lerna run watch --parallel"
```
krassowski, line 54: This is unrelated, just fixes the `jlpm watch` command which was stalling without it

martinRenou:
> @martinRenou I don't think JupyterGIS has a diff viewer, right?

Right, not yet 😁

krassowski:
Thanks! For now, I changed it to only show the Diff button for notebooks.

It would be ideal if we had a diffing service in core where `nbdime`, `jupyterCAD` and `JupyterGIS` in future could register their differs - something to consider for the future.

davidbrochart:
I think that @jtpio is using another diff library than `nbdime`, just wondering if it's worth considering using it here?

krassowski:
@jtpio do you have a link?

jtpio:
David probably refers to https://diffs.com/ (using it with `jupyterlab-git` in https://github.com/jtpio/xtralab), which has a UI for resolving merge conflicts.

krassowski:
> which has a UI for resolving merge conflicts

for files and notebooks or files only?

davidbrochart (approved):
Thanks @krassowski.

PR #593, opened 2026-06-05
Stream notebook loading progressively

MUFFANUJ:
Closes #552  ( contains the frontend handling and absorption of https://github.com/jupyter-server/jupyter_ydoc/pull/418)

This PR enables progressive notebook loading from disk so clients can start rendering before the full notebook content is loaded, with configurable controls for enabling the behavior and delaying large outputs. It wires the new settings through the server app, handlers, and document rooms, updates the frontend provider to resolve readiness on the websocket connection, and blocks client document edits while progressive initialization is still in progress to avoid hiding unsaved changes. Tests and docs were added for the new configuration and loading behavior


krassowski:
1. Can you add a recording to demonstrate how it looks like from the user PoV?
2. Maybe you could pin the jupyter-ydoc dependency temporarily to your branch in `pyproject.toml` (one of them to allow tests to proceed)
3. Maybe it would be worth adding an E2E playwright test to make sure this does not regress?

MUFFANUJ:
> 1. Can you add a recording to demonstrate how it looks like from the user PoV?

https://github.com/user-attachments/assets/90a49709-dfb1-4405-8bbd-76b8aaf0088c

davidbrochart (commented):
I'm wondering if there could be a way to make this behavior more generic. This PR special-cases the notebook, but we could imagine a custom user-defined shared document that implements its own way of being set progressively. How would Jupyter Collaboration automatically take advantage of that?

MUFFANUJ:
> I'm wondering if there could be a way to make this behavior more generic. This PR special-cases the notebook, but we could imagine a custom user-defined shared document that implements its own way of being set progressively. How would Jupyter Collaboration automatically take advantage of that?

Updated the collaboration changes to align with the new generic ydoc API from jupyter-server/jupyter_ydoc#418. The config and room wiring now use document-level names, progressive initialization is used only for shared documents that override `aset_progressively`, and source loading now calls `aset_progressively(...)` directly instead of notebook-specific `aset(...)` options.

davidbrochart:
Thanks a lot @MUFFANUJ, that's great.
I can see an issue related with JupyterLab's frontend that doesn't like to have an empty notebook on load, not even the notebook metadata. In that case it will show a dialog to ask the user to choose a kernel, while it shouldn't because the information about which kernel to use it there. I think we should update https://github.com/jupyter-server/jupyter_ydoc/pull/418 so that `aset_progressively` has a chance to do some initialization (here, set the metadata).

packages/docprovider/src/yprovider.ts
```diff
@@ -431,6 +429,17 @@ export class WebSocketProvider implements IDocumentProvider, IForkProvider {
...
   };
 
+  private _onStatus = ({
+    status
+  }: {
+    status: 'connecting' | 'connected' | 'disconnected';
+  }) => {
+    if (status === 'connected') {
+      this._attachConflictListener();
+      this._ready.resolve();
```
davidbrochart, line 439: We don't want the document to be ready on connection, but only on synchronization.
```suggestion
```

projects/jupyter-server-ydoc/jupyter_server_ydoc/rooms.py
```diff
@@ -235,6 +293,15 @@ async def _broadcast_updates(self):
...
 
+    def _filter_message(self, message: bytes) -> bool:
+        """Skip client document updates while source content is still loading."""
+        return (
+            self._block_client_document_updates
+            and len(message) > 1
+            and message[0] == MessageType.SYNC
+            and message[1] in (YSyncMessageType.SYNC_STEP2, YSyncMessageType.SYNC_UPDATE)
+        )
+
```
davidbrochart, line 304: I don't think we should ignore changes from the frontend while the document is loading. Their view will still show their changes anyway.
```suggestion
```

projects/jupyter-server-ydoc/pyproject.toml
```diff
@@ -29,7 +29,7 @@ authors = [
 ]
 dependencies = [
     "jupyter_server>=2.15.0,<3.0.0",
-    "jupyter_ydoc>=4.0.0b0,<5.0.0",
+    "jupyter_ydoc @ git+https://github.com/MUFFANUJ/jupyter_ydoc.git@streamNotebook",
```
davidbrochart, line 32:
```suggestion
    "jupyter_ydoc >=4.1.0,<5.0.0",
```

projects/jupyter-server-ydoc/pyproject.toml
```diff
@@ -60,6 +60,9 @@ test = [
 [tool.hatch.version]
 path = "jupyter_server_ydoc/_version.py"
 
+[tool.hatch.metadata]
+allow-direct-references = true
+
```
davidbrochart, line 65:
```suggestion
```

projects/jupyter-server-ydoc/jupyter_server_ydoc/rooms.py
```diff
@@ -181,8 +198,39 @@ async def initialize(self) -> None:
...
+                self._update_lock.release()
+
+    def _should_load_document_progressively(self, loaded_from_store: bool) -> bool:
+        progressive_setter = getattr(type(self._document), "aset_progressively", None)
+        return (
+            self._document_load_progressively
+            and not loaded_from_store
+            and progressive_setter is not None
+            and progressive_setter is not YBaseDoc.aset_progressively
+        )
```
davidbrochart, line 212: I think we can simplify this, since the default implementation of `aset_progressively` is `aset`, and we take a different path when we load from store?
```suggestion
    def _should_load_document_progressively(self, loaded_from_store: bool) -> bool:
        return self._document_load_progressively
```

projects/jupyter-server-ydoc/jupyter_server_ydoc/rooms.py
```diff
@@ -170,7 +180,14 @@ async def initialize(self) -> None:
                     self._room_id,
                     self._file.path,
                 )
-                if not loaded_from_store:
+                if self._should_load_document_progressively(loaded_from_store):
+                    self._block_client_document_updates = True
+                    self.ready = True
+                    self._emit(LogLevel.INFO, "initialize", "Room initialized")
+                    self.create_task(self._finish_progressive_initialization(model["content"]))
```
davidbrochart, line 187: We should wait for the document to be "initialized" (for a notebook, to have its metadata populated):
```suggestion
                    initialized = asyncio.Event()
                    self.create_task(self._finish_progressive_initialization(model["content"], initialized))
                    await initialized.wait()
```

krassowski:
> I can see an issue related with JupyterLab's frontend that doesn't like to have an empty notebook on load, not even the notebook metadata. In that case it will show a dialog to ask the user to choose a kernel, while it shouldn't because the information about which kernel to use it there. I think we should update https://github.com/jupyter-server/jupyter_ydoc/pull/418 so that aset_progressively has a chance to do some initialization (here, set the metadata).

https://github.com/jupyterlab/jupyterlab/pull/18987 highlights another scenario where frontend expects cells to be in the model already, even if it does not expect all cells to be rendered initially - in particular it expects to know the total count of cells (https://github.com/jupyterlab/jupyterlab/blob/1a7005fade14ffd0b50c6ea915fa7b8af9cb6f91/packages/ui-components/src/components/windowedlist.ts#L1139-L1153). I wonder if we beside the metadata such as `kernelspec` we should have a way to pass some hints to the frontend about the shape of the notebook (e.g. "it has 100 cells, you can render 100 placeholders"). This is not a blocker for this PR just thinking out loud how this should look long-term.

davidbrochart:
> I wonder if we beside the metadata such as `kernelspec` we should have a way to pass some hints to the frontend about the shape of the notebook (e.g. "it has 100 cells, you can render 100 placeholders").

I think loading a notebook progressively is not different from a user inserting cells manually (it might just be happening faster), so I don't see why the frontend should have a hint about the (number of) cells. Also, the outputs can even be populated after all the cells, so the hint would not be very helpful.

krassowski:
Well, I was mostly thinking about populating UI scaffold/scrollbar. Again, this is hypothetical, future-looking.

docs/source/configuration.md
```diff
@@ -52,6 +52,13 @@ jupyter lab --YDocExtension.file_poll_interval=2
 # If None, the document will be kept in memory forever.
 jupyter lab --YDocExtension.document_cleanup_delay=100
 
+# Whether documents loaded from disk should be streamed progressively (default: True).
+jupyter lab --YDocExtension.document_load_progressively=False
+
+# Delay document outputs larger than this size in MB during progressive loading (default: 100).
+# If None, outputs are loaded with their content.
+jupyter lab --YDocExtension.document_output_delay_threshold_mb=50
```
davidbrochart, line 60:
```suggestion
# Defer notebook outputs larger than this size in MB during progressive loading (default: 100).
# If None, outputs are loaded with the inputs.
jupyter lab --YDocExtension.notebook_output_defer_threshold_mb=50
```

davidbrochart: Ah since jupyter-ydoc already uses "delay" we should keep that word.

krassowski:
Closing as included in #607 - thanks!

PR #596, opened 2026-06-11
Show dialog when document initialization times out

davidbrochart:
This should not happen with a WebSocket provider and `jupyter-server-ydoc`, but it might with a WebRTC provider such as [jupyter-webrtc-provider](https://github.com/davidbrochart/jupyter-webrtc-provider).
I simulated a timeout when loading a document here:

https://github.com/user-attachments/assets/b7001984-7db8-42d1-b70c-3d95f8d32ec2


packages/docprovider/src/yprovider.ts
```diff
@@ -218,8 +227,72 @@ export class WebSocketProvider implements IDocumentProvider, IForkProvider {
...
+      await this.reconnect();
+    } else {
+      this._disconnect();
+      const msg = this._trans.__(
+        'The document failed to load. Please try opening it again.'
+      );
+      this._ready.reject(msg);
+      this._ready.promise.catch(() => undefined);
+      this._sharedModel.dispose();
+    }
```
krassowski, line 302: Could there be a third option, "continue waiting" or "wait more"?

I am concerned that with this timeout it is common for users I work with to spend more than 5 seconds waiting, and it looks like this disconnects with either option - so I worry if this could lead to a regression such that the document never gets loaded? I have not tested, just highlighting my doubt here.

davidbrochart: Yes I was thinking about that too :+1:

krassowski (approved):
I think it makes sense. I am not sure if the 5 seconds default is long enough, but let's try it.

PR #607, opened 2026-07-02
Use jupyter-ydoc's `aset_progressively`

davidbrochart:
Builds on #593.


davidbrochart:
The test about conflict resolution [fails](https://github.com/jupyterlab/jupyter-collaboration/actions/runs/28620799771/job/84877403120?pr=607) because now the updates from the frontend are applied to an empty notebook (the cells are inserted after in the backend) so there is no conflict anymore, compared to before progressive document loading where the notebook is fully populated in the backend at synchronization time.

davidbrochart:
There are still 2 conflict tests that are failing. I don't understand because it should not be using the changes in this PR, and they pass on `main`.
Any idea @krassowski ?

davidbrochart:
CI is green now.
I think we should discuss the fact that there cannot be any conflicts now, with notebook progressive loading.

davidbrochart:
> I think we should discuss the fact that there cannot be any conflicts now, with notebook progressive loading.

Maybe the new feature should be opt-in?

davidbrochart:
I'll merge this, now that this is opt-in.

MUFFANUJ:
I will add a few comments soon!

MUFFANUJ:
I think there are a few blocking edge cases around progressive initialization:

1. `get_document(create=True, copy=True)` can return a partial document. After `await room.initialize()`, `get_document()` immediately snapshots `room.ydoc.get_update()` ([app.py#L291-L313](https://github.com/jupyterlab/jupyter-collaboration/blob/99e3a24eb5f0973887a7705a83dc08766d716e9c/projects/jupyter-server-ydoc/jupyter_server_ydoc/app.py#L291-L313)), but the progressive path returns right after `finish.set()` ([rooms.py#L189-L194](https://github.com/jupyterlab/jupyter-collaboration/blob/99e3a24eb5f0973887a7705a83dc08766d716e9c/projects/jupyter-server-ydoc/jupyter_server_ydoc/rooms.py#L189-L194)) while `_finish_progressive_initialization()` can still be replaying source content. Here, I feel we should track/await the progressive init task before returning copied documents.

2. Progressive initialization can hang on early failures. `initialize()` waits on `initialized.wait()` ([rooms.py#L189](https://github.com/jupyterlab/jupyter-collaboration/blob/99e3a24eb5f0973887a7705a83dc08766d716e9c/projects/jupyter-server-ydoc/jupyter_server_ydoc/rooms.py#L189)), but the background task only logs exceptions and does not wake/propagate to the waiter ([rooms.py#L218-L223](https://github.com/jupyterlab/jupyter-collaboration/blob/99e3a24eb5f0973887a7705a83dc08766d716e9c/projects/jupyter-server-ydoc/jupyter_server_ydoc/rooms.py#L218-L223)). we should probably validate `notebook_output_delay_threshold_mb >= 0` and propagate init failures.

3. Client document updates can be accepted while source replay is still in progress. The room is marked ready before replay finishes ([rooms.py#L190-L192](https://github.com/jupyterlab/jupyter-collaboration/blob/99e3a24eb5f0973887a7705a83dc08766d716e9c/projects/jupyter-server-ydoc/jupyter_server_ydoc/rooms.py#L190-L192)), source updates continue later ([rooms.py#L246-L250](https://github.com/jupyterlab/jupyter-collaboration/blob/99e3a24eb5f0973887a7705a83dc08766d716e9c/projects/jupyter-server-ydoc/jupyter_server_ydoc/rooms.py#L246-L250)), and `_on_document_change()` skips saves while `_update_lock` is held ([rooms.py#L379-L380](https://github.com/jupyterlab/jupyter-collaboration/blob/99e3a24eb5f0973887a7705a83dc08766d716e9c/projects/jupyter-server-ydoc/jupyter_server_ydoc/rooms.py#L379-L380)). Maybe we should block or buffer client `SYNC_STEP2` / `SYNC_UPDATE` messages until progressive replay finishes, and add a test that pauses server-side replay, sends/attempts an edit, then verifies it is not lost and autosave still happens.

davidbrochart:
I'm off for a week or two, I'll look at that when I come back.

davidbrochart:
https://github.com/jupyterlab/jupyter-collaboration/pull/607/commits/61d0b804f96f7990e1de408fe6a1de3d4d179fb5 should take care of point 1.

davidbrochart:
https://github.com/jupyterlab/jupyter-collaboration/pull/607/commits/2339a6b9f927a89debb1832f7c93a2e6fa7c5d41 should take care of point 2, but exceptions can only be raised during the initialization phase of progressive loading, not during the loading of the rest of the document, since this one is done in the background after the room is marked as "ready".

davidbrochart:
I'm not sure about point 3. To me, progressive loading of the document is not very different from a user making changes to a document. I don't think we should prevent users from making changes to a document while it is being progressively loaded, especially if it is very big and it takes a lot of time to load. Also, we don't have a good UX for that, because even if we don't accept the user's Y-updates, they still see their changes locally.
That said, if there are user changes while progressively loading, they should be saved indeed. We could check if the document on disc (that we already read) is different from the content of the shared document that was progressively loaded, and save to disk if it is the case.

projects/jupyter-server-ydoc/jupyter_server_ydoc/rooms.py
```diff
@@ -171,18 +184,70 @@ async def initialize(self) -> None:
...
+                        self.create_task(
+                            self._finish_progressive_initialization(
+                                model["content"], initialized, finish
+                            )
+                        )
+                        await initialized.wait()
+                        if self._document_progressively_loaded.done() and (
+                            exc := self._document_progressively_loaded.exception() is not None
+                        ):
+                            raise cast(BaseException, exc)
```
krassowski: The precedence here appears incorrect, should it be?

```suggestion
                        if self._document_progressively_loaded.done() and (
                            exc := self._document_progressively_loaded.exception()
                        ) is not None:
                            raise exc
```

davidbrochart: Good catch!

krassowski (approved):
Thank you! Obviously, I am not a fan of the conflict detection and progressive load not working at the same time, but I will try to address this to separately in a follow-up.

Issue #594, opened 2026-06-09
Duplication of cells after restart (no after out of band change) with no YStore/YStore deleted


krassowski:
## Description

This is:
- an isolated reproduction of https://github.com/jupyterlab/jupyter-collaboration/issues/545
- a follow-up to #571 which fixed that for cells that are already present on the first load

## Reproduce

1. Start jupyterlab with collaboration
2. Connect a browser client
3. Open an empty notebook
4. Add a few cells
5. Disconnect the browser client = suspend the machine with the browser (do not close the tab)
6. Shut down the server
7. Delete Ydoc
8. Start server again
9. Reconnect
10. See cells duplicated

#571 fixed the scenario in which the notebook was not empty to begin with, and document was reloaded upon room expiry. This is a more general case.

## Expected behavior

No cell duplication

## Context

Had we represented notebook as a dictionary of cell-id → cell content we would not get a silent duplication (but hard conflicts); that ship has shipped though.

martinRenou:
Lately, I've seen multiple times a similar issue, but it was the cell content that was duplicated e.g. the cell would initially contain `print("hello")` before shutting down the server, then when restarting the server the notebook cell contains `print("hello")print("hello")`.

davidbrochart:
> Had we represented notebook as a dictionary of cell-id → cell content we would not get a silent duplication (but hard conflicts); that ship has shipped though.

A `Map` is not ordered though.

PR #608, opened 2026-07-23
Add `disable_ystore` option to make the YStore optional

### LINKED ISSUE #430: Make the use of `YStore` configurable

jtpio:
Fixes https://github.com/jupyterlab/jupyter-collaboration/issues/430

### DIFF
--- a/docs/source/configuration.md
+++ b/docs/source/configuration.md
@@ -32,6 +32,11 @@ relying on the persisted history (such as the document timeline) are not availab
 useful on deployments where the database file would grow too large or where persisting it is
 impractical.
 
+> **Note**: The persisted history is also what allows merging document changes across a server
+> restart. With the YStore disabled, a client reconnecting after a server restart can run into
+> conflicts, which may lead to duplicated content or require reloading the document (see
+> [issue #594](https://github.com/jupyterlab/jupyter-collaboration/issues/594)).
+
 A second file, `collaboration_sessions.json`, is created inside a `.jupyter` directory under
 the server's root directory (`ServerApp.root_dir`, which defaults to the directory where
 JupyterLab was launched). It records recent collaboration session IDs so the backend can
--- a/docs/source/configuration.md
+++ b/docs/source/configuration.md
@@ -25,6 +25,13 @@ looses connection and goes offline for a while. You should never have to touch i
 fine to just ignore it, including in your version control system (don't commit this file). If
 you happen to delete it, there shouldn't be any serious consequence either.
 
+This store of document updates (the "YStore") can also be disabled entirely with
+`--YDocExtension.disable_ystore=True`. In that case, document updates are only kept in memory
+while a document is open on the server, no `.jupyter_ystore.db` file is created, and features
+relying on the persisted history (such as the document timeline) are not available. This can be
+useful on deployments where the database file would grow too large or where persisting it is
+impractical.
+
 A second file, `collaboration_sessions.json`, is created inside a `.jupyter` directory under
 the server's root directory (`ServerApp.root_dir`, which defaults to the directory where
 JupyterLab was launched). It records recent collaboration session IDs so the backend can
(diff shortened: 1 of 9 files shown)


docs/source/configuration.md
```diff
@@ -25,6 +25,13 @@ looses connection and goes offline for a while. You should never have to touch i
 fine to just ignore it, including in your version control system (don't commit this file). If
 you happen to delete it, there shouldn't be any serious consequence either.
 
+This store of document updates (the "YStore") can also be disabled entirely with
+`--YDocExtension.disable_ystore=True`. In that case, document updates are only kept in memory
+while a document is open on the server, no `.jupyter_ystore.db` file is created, and features
+relying on the persisted history (such as the document timeline) are not available. This can be
+useful on deployments where the database file would grow too large or where persisting it is
+impractical.
```
krassowski, line 33: Also, should mention that disabling ystore means conflicts when restarting server which may lead to content duplication and or require reloading the documents. As in https://github.com/jupyterlab/jupyter-collaboration/issues/594

krassowski: The exact wording will depend on what counter measures we decide to merge.

davidbrochart:
That looks like opening a can of worms. If there is no YStore, a user disconnected long enough that the room is cleared will reconnect with their content, which will be added to the content from disk. You can expect lots of complaints about duplication issues again. And if you don't clear rooms when they're empty, the server will quickly run out of memory.

krassowski:
> That looks like opening a can of worms. If there is no YStore, 

FWIW this is what users already do out there - some would achieve this with a no-op `YStore` subclass, some point database location to a ramdisk, some delete the file via cron job. Having this as an explicit option at least allows us to guard certain workarounds by checking if YStore is enabled.

davidbrochart:
I don't think we should encourage users shooting themselves in the foot. If there are problems with the YStore, we should rather try and fix them.

PR #617, opened 2026-08-21
Fix file rename

davidbrochart:
Currently, a file opened by two peers, and renamed by one peer, doesn't update in the other peer document tab (but it does update in their file browser of course).

- [x] Added a test
- [x] Fixed the bug


krassowski (approved):
Thank you!

davidbrochart:
Thanks, I'll cut a patch release.

PR #618, opened 2026-08-24
Fix file rename

davidbrochart:
Follow-up of #617.


djangoliv (approved):

krassowski:
This has a regression compared to `5.0.1`:

```py
async def test_FileLoader_saves_to_new_path_when_renamed_before_save():
    id = "file-4567"
    old_path = "myfile.txt"
    new_path = "renamed.txt"
    file_id_manager = FakeFileIDManager({id: old_path})

    class RenameBeforeSaveContentsManager(FakeContentsManager):
        def __init__(self):
            super().__init__({"last_modified": datetime.now(timezone.utc), "writable": True})
            self.saved_paths: list[str] = []
            self.renamed = False

        def get(self, path, content=True, format=None, type=None, require_hash=None):
            model = super().get(path, content, format, type, require_hash)
            if not content and not self.renamed:
                # The metadata check in maybe_save_content has just run; the
                # file is renamed before the save reaches the contents manager.
                self.renamed = True
                file_id_manager.move(id, new_path)
            return model

        def save(self, model, path):
            self.saved_paths.append(path)
            return self.model

    cm = RenameBeforeSaveContentsManager()
    loader = FileLoader(id, file_id_manager, cm)
    await loader.load_content("text", "file")

    await loader.maybe_save_content({"format": "text", "type": "file", "content": "content"})

    assert cm.saved_paths == [new_path]
```

Raises

```
E       AssertionError: assert ['myfile.txt'] == ['renamed.txt']
```

krassowski:
CC @davidbrochart

krassowski:
Realistically this is just an edge case race condition unlikely to happen too often (it requires a few user interactions within 1-2s window and bad network timing), but I suppose the goal of these PRs was to fixed cases exactly like that.

davidbrochart:
Thanks, I opened #620.

Issue #518, opened 2025-11-26
Race condition when manually saving after renaming files: updating file name takes too long


krassowski:
## Description

If two users (or a single user with a session open in two browsers) have the same new file open and one user renames it, there is 5-10 seconds delay before the file name is updated for the second user.

If the second user saves the file, in that period there will be two files on disk!

Related to:
- https://github.com/jupyterlab/jupyter-collaboration/issues/260

## Reproduce

https://github.com/user-attachments/assets/bf77a792-d0ba-49a7-8471-5e4793bde81d

## Expected behavior

The updates to file names propagate faster?

## Context

- jupyter-collaboration v4.1.2
- JupyterLab version: 4.5.0

PR #620, opened 2026-08-25
Fix file rename

davidbrochart:
See https://github.com/jupyterlab/jupyter-collaboration/pull/618#issuecomment-5407741974.


krassowski (approved):
Thank you!

Of note, there is one more somewhat related bug, but it might addressing elsewhere: https://github.com/jupyterlab/jupyter-collaboration/issues/518

PR #622, opened 2026-09-08
Make `errorReason` required to show exact reason to client

nakul-py:
### Description
Propagate the initialization error reason to the client through `errorReason`, so users can see the exact reason why a document failed to initialize instead of only the generic error message.

### Motivation

image

Currently, the client only receives `initialization_error` and `reloadable: false`. Without `errorReason`, the client cannot know the exact reason why the document failed to initialize.

Reproduced in https://github.com/geojupyter/jupytergis/pull/1836


nakul-py:
@martinRenou @davidbrochart

martinRenou (approved):
Thanks!

projects/jupyter-server-ydoc/jupyter_server_ydoc/handlers.py
```diff
@@ -327,6 +327,7 @@ async def open(self, room_id: str) -> None:  # type:ignore[override]
                             {
                                 "reason": "initialization_error",
                                 "reloadable": False,
+                                "errorReason": str(e),
```
davidbrochart, line 330: Why not reusing `"reason"` on line 328?

krassowski: That one gets parsed by the frontend already and is a deterministic string. If you pass along only the raw stringified error it will be less clear and no longer translated. I think this is a fine as-is.

davidbrochart (approved):

Issue #261, opened 2024-03-22
Show a user's cursor to collaborators when typing


Zsailer:
### Problem

In most collaborative editing applications, when a user is typing, their name appears over the cursor. This is a really helpful (and now expected UX) when collaborating in documents. This allows everyone to know who is making the current changes as they appear.

Jupyter collaboration doesn't do this. It only shows a name when you hover over the cursor. There are some improvements to that behavior we can make too, highlighted in #259.

### Proposed Solution

We should change Jupyter Collaboration to always show the user's name when they are typing.

### Additional context

See this video of Google docs as an example:

https://github.com/jupyterlab/jupyter-collaboration/assets/2791223/14127276-3b81-40e5-9407-77ddfa1585b8

coolsyntax:
Hi, I would like to work on this issue. Please assign it to me.

davidbrochart:
Thanks for offering your help @coolsyntax!

coolsyntax:
Hi, I am working on this issue, but when I try to use the showTooltip function to create a state field and layer for the username tooltips, the usernames do not appear. I've tried creating a custom view plugin which updates the state on keyboard events, I've also tried creating a Widget decoration to display the username directly above the cursor, and looked at the documentation of CodeMirror, but failed to make any progress. Could you please help me solve this issue? I'd appreciate any guidance or feedback you can provide on the best way to display the user's name above their cursor while they are typing in the collaborative Jupyter environment.

krassowski:
@coolsyntax can you open a draft PR with what you tried and think was the best attempt so far?

arjxn-py:
Hi @davidbrochart @Zsailer @krassowski, I'd like to work on this issue. We've recently improved the UI of collaborator cursor on [JupyterGIS](https://github.com/geojupyter/jupytergis) in https://github.com/geojupyter/jupytergis/pull/1848 - we're planning to do something similar for notebooks too.

davidbrochart:
Thanks a lot @arjxn-py, I've assigned you.
Correct me if I'm wrong, but I think JupyterGIS shows the *mouse* cursor, not the textual cursor as mentioned in this issue.

arjxn-py:
Yes, JupyterGIS draws the mouse pointer on the map, not a text cursor. But I'll take the UI inspiration from there - not the mechanism.

arjxn-py:
This can be closed now!

PR #623, opened 2026-09-16
Improve Collaborative Cursor UI

arjxn-py:
#261

https://github.com/user-attachments/assets/c578a4d1-9b58-4f31-bb31-c65f416cbbcc


SylvainCorlay:
I love it. It is much more readable than the current version.

packages/collaboration/src/cursors.ts
```diff
@@ -116,15 +124,215 @@ const remoteSelectionTheme = EditorView.baseTheme({
...
+        authors.push(clientID);
+      }
+    });
+    if (authors.length > 0) {
+      markActive(awareness, authors);
+    }
+  });
+}
+
+function collaboratorPill(user: User.IIdentity | undefined): HTMLDivElement {
```
martinRenou: Should we/could we export this? That way we can simply reuse it in JupyterGIS and JupyterCAD?

martinRenou (dismissed):

packages/collaboration/src/cursors.ts
arjxn-py: Done in [283a90c](https://github.com/jupyterlab/jupyter-collaboration/pull/623/commits/283a90cf044ecbb15f803f9557e29610b94b977f)

krassowski:
Bidner does not load up for me, do you have a screenshot how it looks in File Editor and in the Console?

arjxn-py:
https://github.com/user-attachments/assets/800b1830-7fd8-4c4b-8de7-4b0993b3bb28


@krassowski other than terminal, it is there on both file editor and notebook
I'm not sure if we would want the same for terminal

krassowski:
No terminal is not backed by CRDT. I meant the Console view not Terminal. Anyways looks great for files too!

davidbrochart:
> No terminal is not backed by CRDT.

But it [could be](https://github.com/jupyterlab/jupyterlab/issues/13339) :)

arjxn-py:
> No terminal is not backed by CRDT. I meant the Console view not Terminal. Anyways looks great for files too!

I wonder how do i open same console in 2 separate windows - I tried but didn't manage to

davidbrochart:
Yes consoles are not backed by a file so I don't think you can collaborate on them.

krassowski (changes requested):
I think this is an amazing improvement! No strong opinion on the style

Testing this locally a couple of issues:

1. Deleting text at the end where the collaborator's cursor is leads to a crash with `Error: No tile at position` in the console - this bug needs fixing before merge IMO

https://github.com/user-attachments/assets/3b788b8d-b472-4e68-baf1-80fd9a16b5bf

2. Hovering over the collaborator's cursor puts the pill one line too high (see 0:11 below) - this should be fixed before merge

https://github.com/user-attachments/assets/336a1d83-42c0-4e40-9cb4-67b4c9bf4b9e

3. User identity is frozen after first cursor creation:

https://github.com/user-attachments/assets/830ce9c8-948c-40b2-bd8d-84d58389a8c5

4. Nice to have: selection should show the pill too - this is what Google Docs does; currently it does not. This could be a new issue

https://github.com/user-attachments/assets/ce495bba-a333-44a9-bb55-a4923ce9f5ba

arjxn-py:
@krassowski I've addressed all the issues you mentioned including the selection one. Thanks for comprehensive testing and review :)

krassowski:
I think that the pill rendered on selection should disappear after some time, otherwise it might be stuck forever

https://github.com/user-attachments/assets/1bffc8d0-08b5-43e7-98ad-ce969dd89424

arjxn-py:
Thanks @krassowski, I've addressed this too.

krassowski (approved):
Thanks, looks great!

---

Suggestion for a future enhancement: currently the pill takes up a lot of space at all times; in environments with multiple collaborators this could occlude the content. 

Maybe we could collapse pills to a round collaborator icon only when there are many cursors in a neighborhood and only expand when hovered over.

arjxn-py:
> Maybe we could collapse pills to a round collaborator icon only when there are many cursors in a neighborhood and only expand when hovered over.

Sure, i'll be happy to work on it as a follow up - I'll open an issue for now referencing this comment. Thanks a lot @krassowski.

PR #624, opened 2026-09-18
Fix binder build 

krassowski:
It's working now:

image


binder/environment.yml
```diff
@@ -4,17 +4,13 @@ channels:
   - nodefaults
 dependencies:
   - jupyterlab >=4.6.0
-  - nodejs >=18,<19
-  - python >=3.11,<3.12
+  - nodejs >=24,<25
```
krassowski, line 7: Newer Node is required for jupyter-builder (and v18 is EOL since Mar 27, 2025 anyways)

davidbrochart (approved):
Thanks @krassowski.

PR #628, opened 2026-09-21
Ship the cursor pill styles as plain CSS

### LINKED ISSUE #626: Collaborator pill can't be styled outside CodeMirror
#623 exports `collaboratorPill()`, but its styles sit in `remoteSelectionTheme`, a CodeMirror base theme. CodeMirror prefixes those selectors with the generated editor class, so they only apply inside an editor. Call `collaboratorPill()` anywhere else and you get an unstyled div.

The package already ships plain CSS. Could the `.jp-remote-userFlag` rules move to a `style/cursors.css` imported from `style/base.css`, leaving the cursor and selection rules in the base theme?

I hit this in [JupyterGIS](https://github.com/geojupyter/jupytergis/pull/1880), where we draw the same badge on remote map cursors. Happy to open a PR.

arjxn-py:
Shall close #626


arjxn-py:
cc @krassowski @davidbrochart this is ready for a review.

davidbrochart (approved):

krassowski:
The border got lost. 

Before:
image

After:
image

Already before this PR I was thinking whether this deserves a UI test but now since we are now also making this a contract for other extensions to depend on.

Maybe we should add a screenshot of the pill in editor just to catch any unexpected regressions when styles get edited (whether here or upstream).

arjxn-py:
Ahh, i did not catch this difference - I agree it'll be wise to add a UI test for this. I'll work on writing one.

arjxn-py:
Can we do update snapshots on this one please?

krassowski:
bot please update snapshots

krassowski (approved):
Thanks!

